PR #35, opened 2015-05-15
Limiting the in memory queue sizes for slower handlers

AlanMullett:
Using a BlockingCollection within the AzureServiceBusDequeueStrategy allows for the consumption of messages from the Azure Service Bus to match the rates through the handlers.  The size of the queue is determined by the BatchSize configuration setting.  This is also used to restrict the Azure client libraries own prefetch mechanism.


SeanFeldman:
@AlanMullett we can't commit to master, it needs to be the `develop` branch... I can pick the changes and do it or you can re-submit PR to target `develop`. Thank you :)

Issue #34, opened 2015-05-15
Slow handlers with large queues result in many Dead Letters


AlanMullett:
With a slow handler that has a large queue the Azure Service Bus prefetch cache fetches an inappropriately large number of messages.  This results in the peeklock timing out before the message is handled.  If this continues eventually the message ends up as a dead letter due to the number of fetch attempts.

The NServiceBus code should not hard code the prefetch size to 100 but instead use the BatchSize (or similar setting) to limit the cache size and use a BlockingCollection in the *AzureServiceBusDequeueStrategy* rather than a simple Queue.  This will result in the call to EnqueueMessage blocking and so limit the number of messages held in memory to an appropriate number.

Example of a possible solution: https://github.com/AlanMullett/NServiceBus.AzureServiceBus/tree/LimitingTheInMemoryQueueSizesForSlowerHandlers

SimonCropp:
@AlanMullett that looks like a good addition. can you submit a PR for the configuration change.

I notice you are also using BlockingCollection to limit the number of messages. That looks like a more sensible way. Would you have time to also submit a different PR for that?  

I think it is better to have them as separate PRs since the config change is obvious thing to do while the BlockingCollection  will probably need more discussion.

AlanMullett:
I don't think that separte PRs are appropriate, just the config change on its own achieves nothing as the _queue in the AzureServiceBusDequeueStrategy class will just fill up at a different rate but it will still end up with every available message on the Azure Service Bus queue on it PeekLock'ed.  The notifier will just get them at different rates as it has no way to be paused.  By forcing it to wait in the BlockingCollection.Add method it gets throttled down to the speed of consumption.  In addition we reduce the Azure client libraries own buffer as that was defaulted to much too large a number for us.

Without the blocking collection I was seeing over a thousand messages locked to an instance that would take hours to get through that list with the result that after the peek lock timeouts and refetches most messages ended up dead.  With these modifications we are now seeing no message loss whereas without it we were seeing most messages (over 80%) ending up as dead letters for no good reason.

SeanFeldman:
@AlanMullett please sign [CLA](http://www.particular.net/contributors-license-agreement-consent) and go ahead with a single PR
Thank you.

PR #37, opened 2015-05-26
Use BatchSize to limit message consumption to handler performance

AlanMullett:
PR to address issue #34 when messages are being Dead Lettered due to large number of messages on a queue for a slow handler.

This uses a blocking collection which is sized by the BatchSize configuration setting and the number of handler threads.  

Alan.


malkinservicebus:
Hi @AlanMullett, Thank You for PR.
We have a surprise for you!
Please forward your email address to operations@nservicebus.com
Thanks again!
Malki

Issue #40, opened 2015-06-04
Individualization can cause subscription names to exceed 50 character limit


yvesgoeleven:
## Who's affected

Anyone who's using ASB and running outside of cloud services where combination of subscription and machine names can exceed 50 characters limit.

## Symptoms

We check character limit before individualization

https://github.com/Particular/NServiceBus.AzureServiceBus/blob/master/src/Transport/Topology/QueueAndTopicByEndpoint/NamingConventions.cs#L94

that should have been after

_Note_: there might [one more problem](https://github.com/Particular/NServiceBus.AzureServiceBus/issues/40#issuecomment-112531497) associated with this

SeanFeldman:
Before core 5.2.x transport was applying individualization, so the logic to create subscription name was the following:
1. check subscription name (S) for a max length (50). 
2. If max length is exceeded, we turn S into a deterministic GUID (32 characters w/o dashes) 
3. Individualize by appending "-X" where X is either instance number (cloud) or machine name (not-cloud)

This is where the _reported issue_ could be monitored if S + "-X" > 50 characters

After core 5.2.x endpoint can be already individualized by core and S will contain endpoint name and "-X"  when the check for max length is done, turning into GUID combination of S + "-X".

SeanFeldman:
@yvesgoeleven where was it brought up (google user group / question)?

yvesgoeleven:
twitter

PR #46, opened 2015-07-01
fixed bug with individualized subscription name lengths >= 50

johningle:
Added logic to detect individualized subscription name lengths >= 50 and
utilize deterministic guids.  Name lengths < 50 go through the prior
convention logic.


SeanFeldman:
@johningle - thank you for PR.
I have gone through suggested change and think we'll need to come up with a different strategy here. Change has to be backwards compatible to ensure other users are not broken by the change. Consider the following scenario:

Event: `Some.Namespace.Events.EventNameThatExceedsMaximumLength` (55 chars, exceeds max 50)
Machine name: `JOHNI-D`

Today, this will be converted into a deterministic GUID (`8e8faf92-24b0-6296-e76b-f34231bdb758`) and individualized resulting in subscription name `8e8faf92-24b0-6296-e76b-f34231bdb758-JOHNI-D`

If PR is applied, same subscription will result in subscription name `39f0981d-c539-2c08-ea74-141cab944843`.

johningle:
@SeanFeldman - Thank you for the review Sean, I see the issue and have come up with the following scenarios that any fix will need to account for:

1. subscription name length < 50 characters, no guid calculated, should not individualize, so final name length < 50 characters, backward compatible, handled by existing code
2. subscription name length < 50 characters, no guid calculated, should not individualize, so final name length >= 50 characters is impossible since original name length < 50 and not changed, backward compatible, handled by existing code
3. subscription name length < 50 characters, no guid calculated, should individualize, and final name length < 50 characters, backward compatible, handled by existing code
4. subscription name length >= 50 characters, guid calculated, should not individualize, and final name length < 50 characters, backward compatible, handled by existing code
5. subscription name length >= 50 characters, guid calculated, should not individualize, and final name length >= 50 characters is impossible since guid length < 50, backward compatible, handled by existing code
6. subscription name length >= 50 characters, guid calculated, should individualize, and final name length < 50 characters, backward compatible, handled by existing code

7. subscription name length < 50 characters, no guid calculated, should individualize, and final name length >= 50 characters, individualization sent it over, not handled by existing code, no need for backward compatible solution because it causes an exception, need fix
8. subscription name length >= 50 characters, guid calculated, should individualize, and final name length >= 50 characters, individualization sent it over, not handled by existing code, no need for backward compatible solution because it causes an exception, need fix

I have since committed a revision that I believe accounts for all of these scenarios while retaining backward compatibility.
No existing environment should contain object names meeting cases 7 and 8 because the excessive length would cause an exception in Microsoft's Service Bus DLL and prevent creation of the object.

SeanFeldman:
@johningle :+1: for backwards compatibility.
@yvesgoeleven what do you think?

yvesgoeleven:
Thank you for this PR! @johningle

SeanFeldman:
related issue: #40

malkinservicebus:
Hi @johningle - Thank You for PR.
We have a surprise for your. Please forward  your email address to operations@nservicebus.com
Thanks again!
Malki

johningle:
@malkinservicebus I'm afraid that I am only just seeing your response.  Though it has been a few months, can I still get the surprise?

malkinservicebus:
sure @johningle on condition that you send me your address :)

johningle:
@malkinservicebus Sent!

malkinservicebus:
:+1:

PR #67, opened 2015-12-10
Integrating circuit breaker with MessagePump

SeanFeldman:
A few things:
1. had to copy CB into the project since it was internalized
1. test added under seam, not sure what place would be better

@yvesgoeleven please review

### DIFF
--- a/src/Transport/Topology/Topologies/StandardTopology.cs
+++ b/src/Transport/Topology/Topologies/StandardTopology.cs
@@ -90,8 +90,11 @@ namespace NServiceBus.AzureServiceBus
 
         public Func<CriticalError, IPushMessages> GetMessagePumpFactory()
         {
-            // todo, get criticial error integrated
-            return error => container.Resolve<IPushMessages>();
+            return error =>
+            {
+                container.Register<CriticalError>(() => error);
+                return container.Resolve<IPushMessages>();
+            };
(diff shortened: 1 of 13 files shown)


src/Transport/Topology/Topologies/StandardTopology.cs
```diff
@@ -90,8 +90,11 @@ public Func<ICreateQueues> GetQueueCreatorFactory()
 
         public Func<CriticalError, IPushMessages> GetMessagePumpFactory()
         {
-            // todo, get criticial error integrated
-            return error => container.Resolve<IPushMessages>();
+            return error =>
+            {
+                container.Register<CriticalError>(() => error);
```
SeanFeldman: @yvesgoeleven we've discussed this one, but I'm not convinced if changing this would be a good idea. To get that working, we'd need to cast, and that means if another implementation of `IPushMessages` is registered, this code will not work as expected.

```c#
var pump = container.Resolve<IPushMessages>()
(pump as MessagePump)?.SetCriticalError(error); // <-- issue
return pump;```

SeanFeldman: @danielmarbach is there a way to get CriticalError as part of the `IPushMessages.Init` call from the core?

PR #68, opened 2015-12-11
Entities cleanup on atts execution

SeanFeldman:
For each ATT delete messages from queues & subscriptions that belong solely to the test that was executed. Includes DLQ messages.

**Hack warning**: reading connecting string from the environment variables since there's no to get it from settings and it's not enjected.


SeanFeldman:
@yvesgoeleven what do you think? This ensures that we don't leave garbage messages behind that would skew test results. Also, it doesn't try to clean everything, scoping to the current queues/topics only.

SeanFeldman:
Will not work. From 26 mins it went up to 57 mins :(
Closing this.

PR #69, opened 2015-12-14
removing CloneWithMessageId extension methods as they are not needed

SeanFeldman:
Native `BrokeredMessage.Clone` clones ID as well.


SeanFeldman:
@yvesgoeleven please review

PR #70, opened 2015-12-16
[WIP] Idempotent test execution

DavidBoike:
Not mergeable but want to see how the test suite responds to the idempotent tet execution behavior.


DavidBoike:
Forget it, this will never be mergeable.

Issue #77, opened 2016-01-05
Forwarding topology to use SHA1 instead of MD5 for deterministic UUIDs


SeanFeldman:
Currently we're using MD5 for the standard topology and need to keep it that way for the backwards compatibility. With the new topology, forwarding topology, we shouldn't be using MD5 and instead utilize SHA1.

Related to https://github.com/Particular/NServiceBus/issues/1217

PR #79, opened 2016-01-05
[WIP] Replacing MD5 with SHA1 for ForwardingTopology

SeanFeldman:
New topology (`ForwardingTopology`) does not have the need to support backwards compatible subscriptions and can be FIPS complient. Instead of MD5 it will use SHA1.

### Potential issue

MD5 when is used to generate a hash create 16 bytes, which is exactly what is needed to form a GUID.
SHA1 creates 20 bytes, 4 bytes more than a GUID can take.

The current solution is trimming the last 4 bytes to create a valid GUID. This could eventually lead disruption of the uniqueness given the right input data. In order to keep the value unique, ideally we'd keep all 20 bytes.

### Potential issue resolution

@dbelcham and myself have discussed this and would like to propose modification to how we create the deterministic subscription _name_. Since `ForwardingTopology` does not have to be backwards compatible, it is not force to comply with the GUID format. Instead of trying to create a GUID, we create a **UUID** from the full 20 bytes. With some sort of formatting for readability, it will still be under the ASB subscription limit (50 characters). 

#### Example
**Subscription name**: TestEvent

**MD5**:
be3914d0f55bc3f12bdefcf7fb799bab ==> be3914d0-f55b-c3f1-2bde-fcf7fb799bab

**SHA1**:
58a82ca4cdb455fb5fff937e6cbcbfd794daafe2 ==> **58a82-ca4cd-b455f-b5fff-937e6-cbcbf-d794d-aafe2**

@Particular/azure please review the current commit and let know what you think of the suggested improvement/refactoring.


SeanFeldman:
Fixes #77

yvesgoeleven:
All we care about is getting a unique, but predictable, name that fits in the length limitations imposed, so using an 20 character UUID is fine, no need to strip off characters; Can we use different names for the classes though? f.e. MD5DeterministicNameBuilder & SHA1DeterministicNameBuilder.

SimonCropp:
> Can we use different names for the classes though? f.e. MD5DeterministicNameBuilder & SHA1DeterministicNameBuilder.

cant we have a sensible default and then a code api for the override?

SeanFeldman:
@yvesgoeleven +1 for better names. Will apply those.

@SimonCropp could you please elaborate on what do you mean by "have a sensible default and then a code api for the override"? Each generator/builder is specific to a topology it's used in and they are not interchangeable.

SimonCropp:
@SeanFeldman why would one topology use md5 and another use sha?

SeanFeldman:
@SimonCropp for backwards compatibility. The existing customers will require to use MD5 when they upgrade. The new customers that go with the `ForwardingTopology` will only use SHA1.

SeanFeldman:
@Particular/azure all green. Could someone pull the trigger please? Than you :)

SimonCropp:
@SeanFeldman ok fair enough. ignore my previous comment

SimonCropp:
on suggested change... the NameBuilders. They have no state and should be static

SeanFeldman:
Good point @SimonCropp. Done.

SeanFeldman:
If there are no objections, I'd like to merge this.

yvesgoeleven:
go ahead

PR #89, opened 2016-01-15
Send poisonous message to DLQ natively

SeanFeldman:
Sending poisonous message to DLQ natively.

I put together an acceptance test (`When_receiving_a_message_with_unknown_transport_encoding`) that is a hack. Perhaps would be better to test it in a "unit test". Similar to `When_receiving_incomingmessages_from_queues` so that we can control queue name & connection string.

@yvesgoeleven please review.

### DIFF
--- a/src/AcceptanceTests/TransportEncoding/When_receiving_a_message_with_unknown_transport_encoding.cs
+++ b/src/AcceptanceTests/TransportEncoding/When_receiving_a_message_with_unknown_transport_encoding.cs
@@ -1,13 +1,18 @@
 ﻿namespace NServiceBus.Azure.Transports.WindowsAzureServiceBus.AcceptanceTests.TransportEncoding
 {
+    using System;
     using System.Threading.Tasks;
+    using Microsoft.ServiceBus;
+    using Microsoft.ServiceBus.Messaging;
     using NServiceBus.AcceptanceTesting;
+    using NServiceBus.AcceptanceTesting.Support;
     using NServiceBus.AcceptanceTests;
     using NServiceBus.AcceptanceTests.EndpointTemplates;
     using NServiceBus.AzureServiceBus;
+    using NServiceBus.MessageMutator;
     using NUnit.Framework;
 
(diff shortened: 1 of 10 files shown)


src/AcceptanceTests/TransportEncoding/When_receiving_a_message_with_unknown_transport_encoding.cs
```diff
@@ -0,0 +1,103 @@
...
+                        ctx.OriginalMessageId = "MyMessageId";
+                        var sendOptions = new SendOptions();
+                        sendOptions.SetMessageId(ctx.OriginalMessageId);
+
+                        return bus.Send(new MyMessage { Id = ctx.OriginalMessageId }, sendOptions);
+                    }))
+                    .WithEndpoint<Receiver>()
+                    .Done(ctx =>
+                    {
+                        var namespaceManager = new NamespaceManagerAdapter(NamespaceManager.CreateFromConnectionString(Environment.GetEnvironmentVariable("AzureServiceBusTransport.ConnectionString"))); // get connection string from test env
```
SeanFeldman: @yvesgoeleven this is the hacky part

yvesgoeleven: I can live with it

src/Transport/Receiving/MessageReceiverNotifier.cs
```diff
@@ -114,7 +115,17 @@ public void Start()
                 return;
             }
 
-            var incomingMessage = brokeredMessageConverter.Convert(message);
+            IncomingMessageDetails incomingMessage;
+            try
+            {
+                incomingMessage = brokeredMessageConverter.Convert(message);
+            }
+            catch (ConfigurationErrorsException exception)
```
yvesgoeleven: We should use a different exception type here, ConfigurationErrorsException is for mistakes in the app.config

SeanFeldman: Changed to use a custom exception

Issue #100, opened 2016-01-26
Throw when PurgeOnStartup is requested


yvesgoeleven:
PurgeOnStartup is not supported by ASB, we should throw when it's requested

melkio:
Why do we not simulate the same behavior (purging) receiving and discarding all messages? 
Inside a while loop we can check if a new message exists (peek) and receive/discard it (with ReceiveMode.ReceiveAndDelete)

yvesgoeleven:
@melkio upgrades are typically done in a running state, which means that something can still be sending to those queues, you would be purging forever, throwing messages away

melkio:
Makes sense :+1: 
Is there a "delivery timestamp" header or something else inside the message? We could drop only messages deliverered before purging starts.

It's only to give the same "experience" across different transports

melkio:
For deferred messages: how does other transport behave with scheduled messages? If a message isn't inside the queue when purging, this message will be delivered to the queue later, and then it's not deleted. I'm pretty sure that Rabbit works in this way or am I wrong?

yvesgoeleven:
You could use EnqueuedTimeUtc property as a time indicator of when it was put on the queue. But messages can be 'invisible' in the queue, when either locked or defered, or in a session, etc... so you will not be able to see them, let alone purge them.

yvesgoeleven:
I do not know how rabbit does it, maybe ask @bording

yvesgoeleven:
PS: in case of deferrral, I think EnqueuedTimeUtc will change when it becomes visible on the queue

SeanFeldman:
The underlying technology doesn't support purging natively. By saying "we can purge", we are introducing failure points where we won't purge all messages, but customers will make assumption that we do. Therefore I see this is something we shouldn't implement, unless native API provides a way.

SeanFeldman:
Another problematic scenario -- code might be configured _not_ to create queues. In that case we'll be deleting queues w/o ability to re-create?

SeanFeldman:
@melkio please update the general description with what's left and either close it or move back if needed. Thank you.

melkio:
Closed by https://github.com/Particular/NServiceBus.AzureServiceBus/issues/109

PR #109, opened 2016-01-28
PurgeOnStartup should throw because not supported

melkio:
During the Init phase of MessagePump there is a new check to verify PurgeOnStartup setting value. 

Connect to #100

### DIFF
--- a/src/Transport/Seam/MessagePump.cs
+++ b/src/Transport/Seam/MessagePump.cs
@@ -1,7 +1,6 @@
 namespace NServiceBus.AzureServiceBus
 {
     using System;
-    using System.Configuration;
     using System.Threading;
     using System.Threading.Tasks;
     using Extensibility;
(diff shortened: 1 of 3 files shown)


src/Transport/Seam/MessagePump.cs
```diff
@@ -27,6 +28,11 @@ public Task Init(Func<PushContext, Task> pump, CriticalError criticalError, Push
             messagePump = pump;
             circuitBreaker = new RepeatedFailuresOverTimeCircuitBreaker("MessagePump", TimeSpan.FromSeconds(30), ex => criticalError.Raise("Failed to receive message from Azure Service Bus.", ex));
 
+            if (settings.PurgeOnStartup)
+            {
+                throw new ConfigurationErrorsException("Azure Service Bus transport doesn't support PurgeOnStartup behaviour");
```
SeanFeldman: We shouldn't throw `ConfigurationErrorsException` for non config section related things. Maybe `InvalidOperationException`?

SeanFeldman:
@melkio can't find a trace for the question you've asked - why not to drain messages when `PurgeOnStartup` is requested. Problem with that is that we could have messages that are not visible yet (deferred) and we can't receive them. Therefore purge won't actually get rid of all the messages.
@yvesgoeleven anything else that would be an issue with the draining approach?

yvesgoeleven:
@SeanFeldman @melkio upgrades are typically done in a running state, which means that something can still be sending to those queues, you would be purging forever

SeanFeldman:
:+1: for upgrade scenario

Issue #76, opened 2016-01-05
Test standard topology with a long running process


SeanFeldman:
Connects to https://github.com/Particular/EndToEnd/issues/2

As a part of NSB.ASB V7 testing, standard topology needs to be tested in the real azure environment as a long running process.

We should log anything that is not handled (using `appdomain.unhandledexception`) to see if we get anything from ASB library we didn't plan for.

- [x] create endurance test
- [ ] setup to run test in testing environment
   - [x] Setup Raygun application for exception tracking
   - [ ] Convert exception logging to Raygun
   - [ ] Setup octopus deployment

SeanFeldman:
A good candidate for GH location is https://github.com/Particular/ProductionTests

SeanFeldman:
The initial idea is to have a simple console app that sends/receives running for a few weeks. 

@Particular/azure should we also add pub/sub to that (for the `StandardTopology`)?

SimonCropp:
If the tests do not need to span multiple versions then there is no reason to move them out of the main azure repro.

SeanFeldman:
@SimonCropp these are not conventional test, but more of the endurance tests. Also, we don't know how small/big it will be, so it doesn't sound quite right to put it with the transport at this point.

Alternatively, we thought of a separate repo for those kind of tests/projects, but didn't want to increase # of repos that is already substantial.

yvesgoeleven:
Just add a project to the asb solution, it's going to be very asb specific

SimonCropp:
@SeanFeldman it may not seem right but it will result in significantly less fiction for you

yvesgoeleven:
Separate repo is fine for me as well, the smaller repo's are the easier to maintain

SeanFeldman:
@SimonCropp ok, we'll start simple, in the transport solution and if need to split it, we'll do so.
/cc @boblangley

boblangley:
Setting up 2 tests to start:
* Long Running Process, sending and receiving messages at regular intervals
* Throttling test: Flood the bus with messages to introduce throttling

I will add a Solution Folder called Soak Tests, and create a Command Line project for each so we can run them as needed.

@Particular/azure Are these the right tests to setup to start and are there others that should be included as well?

SeanFeldman:
For # 1 - that would be a soak test (@yvesgoeleven recommended 1msg/sec to begin with)
[...]

PR #110, opened 2016-01-28
Endurance test

boblangley:
@melkio Here is my WIP test, just running 1msg/sec command over and over. We still need instrumentation to detect unhandled ServiceBus errors

Connects to: #76


src/EnduranceTests/Program.cs
```diff
@@ -0,0 +1,64 @@
...
+                Console.ReadLine();
+            }
+        }
+
+        private static async Task RunEndpointAsync(string endpointName, CancellationToken ct)
+        {
+            var busConfiguration = new BusConfiguration();
+            busConfiguration.UseTransport<AzureServiceBusTransport>();
+            busConfiguration.EndpointName(endpointName);
+            busConfiguration.PurgeOnStartup(true);
```
SeanFeldman: Moving forward this will throw an exception

boblangley: Switch to a random endpoint name instead?

SeanFeldman: Endpoint name is ok. Removing `PurgeOnStartup` should do it.

mauroservienti:
should we consider this in scope for V6? if yes, could you please ad the expected `Tag: Core V6` label? cheers.

boblangley:
@Particular/azure-maintainers Instrumentation finished. Test log messages are in #testmon_zure_asb

SeanFeldman:
@boblangley is this still a WIP or ready for a review? Thanks.

boblangley:
Please review.

SeanFeldman:
@boblangley asked because it's marked as wip :)

src/EnduranceTests/Handlers/HeartbeatHandler.cs
```diff
@@ -0,0 +1,27 @@
...
+
+    public class HeartbeatHandler : IHandleMessages<Heartbeat>
+    {
+        static readonly ILog log = LogManager.GetLogger<HeartbeatHandler>();
+
+        public async Task Handle(Heartbeat message, IMessageHandlerContext context)
+        {
+            log.InfoFormat("HeartbeatHandler Heartbeat with Wait of {0}", TestSettings.Rate);
+
+            if (TestSettings.TestRunId != message.TestRunId)
```
SeanFeldman: What is this for in this context?

src/EnduranceTests/LayoutRenderer/AggregateExceptionLayoutRenderer.cs
```diff
@@ -0,0 +1,43 @@
...
+            if (exception == null)
+            {
+                return;
+            }
+
+            builder.AppendLine(string.Empty);
+            builder.AppendLine(string.Empty);
+            builder.AppendLine("AggregateException Inner Exceptions:");
+
+            foreach (var innerLogEvent in exception.InnerExceptions.Select(iex => new LogEventInfo
```
SeanFeldman: `exception` at this point is an `AggregateException`. Better to iterate over `exception.Flatten().InnerExceptions` to get the real exceptions and not child `AggregateException`.

src/EnduranceTests/Program.cs
```diff
@@ -0,0 +1,96 @@
...
+    using System.Threading;
+    using System.Threading.Tasks;
+    using NServiceBus.AzureServiceBus.EnduranceTests.Commands;
+    using NServiceBus.AzureServiceBus.EnduranceTests.TestUtils;
+
+    internal class Program
+    {
+        private static void WaitForUserInput()
+        {
+            if (Environment.UserInteractive)
```
SeanFeldman: The code is expecting to be executed in an interactive manner. Is this needed?

src/EnduranceTests/Program.cs
```diff
@@ -0,0 +1,96 @@
...
+        {
+            var connectionString = TestEnvironment.AzureServiceBus;
+
+            var busConfiguration = new BusConfiguration();
+            busConfiguration.UseTransport<AzureServiceBusTransport>()
+                .ConnectionString(connectionString);
+            busConfiguration.EndpointName(endpointName);
+            busConfiguration.UseSerialization<JsonSerializer>();
+            busConfiguration.UsePersistence<InMemoryPersistence>();
+            busConfiguration.SendFailedMessagesTo("error");
```
SeanFeldman: Should we also utilize the audit queue as it's a very common thing along with error queue?

src/EnduranceTests/Targets/AzureBlobErrorOnlyTarget.cs
```diff
@@ -0,0 +1,54 @@
...
+            base.InitializeTarget();
+
+            _client = StorageAccount.CreateCloudBlobClient();
+        }
+
+        protected override void Write(LogEventInfo logEvent)
+        {
+            if (logEvent.Level < LogLevel.Error || logEvent.Exception == null) return;
+
+            if (_client == null) return;
```
SeanFeldman: Should it throw?

src/EnduranceTests/Targets/SlackTarget.cs
```diff
@@ -0,0 +1,77 @@
...
+        private void SendMessageToSlack(SlackMessage message)
+        {
+            using (var client = new WebClient())
+            {
+                var data = new NameValueCollection()
+                {
+                    {"payload", JsonConvert.SerializeObject(message)}
+                };
+
+                client.UploadValuesAsync(_uri, "POST", data);
```
SeanFeldman: Async method call in a void message? 
If can't async all the way, could do `client.UploadValuesAsync(_uri, "POST", data).GetAwaiter().GetResult

src/EnduranceTests/TestUtils/BlobNameBuilder.cs
```diff
@@ -0,0 +1,34 @@
...
+            return $"{logEvent.TimeStamp.Ticks}_{logEvent.Exception.GetType().Name}";
+        }
+
+        public static string BuildContainerName(LogEventInfo logEvent)
+        {
+            var name = logEvent.LoggerName.Replace(".", "-").Replace(" ", "").ToLower();
+            return name.Length > 63 ? name.Substring(0, 63) : name;
+        }
+
+        public static string Build(CloudStorageAccount storageAccount, LogEventInfo logEvent)
```
SeanFeldman: Both `Build` methods look like an overload, but are they building the same thing? Hard to figure from the name itself.

src/EnduranceTests/TestUtils/TestEnvironment.cs
```diff
@@ -0,0 +1,44 @@
...
+    using System;
+    using Microsoft.WindowsAzure.Storage;
+
+    internal static class TestEnvironment
+    {
+        public static string AzureServiceBus
+        {
+            get
+            {
+                var connectionString = Environment.GetEnvironmentVariable("AzureServiceBus.ConnectionString");
```
SeanFeldman: Should it follow the convention other environment variable are using by prefixing with `AzureServiceBus.EnduranceTests.`?

SeanFeldman:
@boblangley there was an update on develop with the recent core. Please rebase.

src/EnduranceTests/Handlers/HeartbeatHandler.cs
boblangley: This removes any lingering messages from earlier test runs. Alternative to trying to clear the queue before starting the test.

src/EnduranceTests/Program.cs
boblangley: I can add that. I think we should configure that queue to have a low time to live and disable Dead Lettering on expiration. Can you think of any reason not to set that as the general queue settings? I am thinking setting it to just a couple of hours, I do not think that value would effect the test.

src/EnduranceTests/TestUtils/TestEnvironment.cs
boblangley: I used the same convention the other tests used for the connection string to make running this in other contexts easier.

src/EnduranceTests/Targets/SlackTarget.cs
boblangley: UploadValuesAsync does not use Tasks, it returns void.

src/EnduranceTests/Handlers/HeartbeatHandler.cs
SeanFeldman: Not sure if this is needed... With ATTs it is crucial because outcome of the test should solely depend on the messages from the same run. This is endurance tests, so messages are compatible. As long as we receive and send, all good. I'd say we remove it.

src/EnduranceTests/Program.cs
SeanFeldman: You can set TTL on the messages (TTBR). Or, TTL on the queue description level. DLQ is not enabled by default I think.

src/EnduranceTests/Targets/SlackTarget.cs
SeanFeldman: Sign, great API msft team... `Xyz.Async()` doesn't return a `Task` :)

src/EnduranceTests/Program.cs
boblangley: I set the `DefaultMessageTimeToLive` to 1 minute.

SeanFeldman: :+1:

src/EnduranceTests/Targets/AzureBlobErrorOnlyTarget.cs
boblangley: I just removed it and refactored some of the other initialization. It will fail much earlier.

boblangley:
Rebased, updated to latest core, and made changes based on comments from @SeanFeldman. Ready for review again.

Issue #94, opened 2016-01-26
Validate consistent usage of ConfigureAwait(false)


yvesgoeleven:
Validate consistent usage of ConfigureAwait(false)

SeanFeldman:
@melkio please update and move where needed. Thanks.

SimonCropp:
done as part of https://github.com/Particular/NServiceBus.AzureServiceBus/pull/119

PR #111, opened 2016-01-29
Added ConfigureAwait(false) for awaitable operation

melkio:
Nothing funny to see here @Particular/azure-maintainers :)

Connect to #94


SeanFeldman:
@melkio nice! 
I wonder what kind of impact those had on tests/ATTs.

Issue #96, opened 2016-01-26
StartupCheckResult


yvesgoeleven:
Figure out why there is a StartupCheckResult callback in the transport seam and implement it if necessary

melkio:
StartupCheckResult is used as a pre-checker before starting Receiving and Sending feature in core.
Do we have any check to satisfy before giving back the control to core?

Msmq transport, for example, use this "extension point" to verify and log if queues permissions are properly configured.

yvesgoeleven:
We should definitly validate the following
- At least one connection string /  namespace is set
- If create queues is on, we should validate that those connectionstrings have manage permissions

melkio:
If create queues is off, do we have to check that queues already exist?

yvesgoeleven:
You would need manage rights to check that, which is not always given when create queues is off. So I would say no?

melkio:
:+1:

SeanFeldman:
@melkio please update the general description with what's left and either close it or move back if needed. Thank you.

SeanFeldman:
Considered done. @melkio please reopen if not.

PR #113, opened 2016-02-03
pre startup check integration

melkio:
@Particular/azure-maintainers could you evaluate this new implementation?

Could be an option to open to users the possibilities to inject "pre startup checks" from configuration API or somewhere else

Connects to #96

### DIFF
--- a/src/Tests/API/APIApprovals.ApproveAzureServiceBusTransport.approved.txt
+++ b/src/Tests/API/APIApprovals.ApproveAzureServiceBusTransport.approved.txt
@@ -254,13 +254,13 @@ namespace NServiceBus.AzureServiceBus
         public ForwardingTopology() { }
         public bool HasNativePubSubSupport { get; }
         public bool HasSupportForCentralizedPubSub { get; }
-        public async System.Threading.Tasks.Task<NServiceBus.Transports.StartupCheckResult> ApplyPreStartupChecks() { }
         public System.Func<NServiceBus.Transports.IDispatchMessages> GetDispatcherFactory() { }
         public System.Func<NServiceBus.Transports.IPushMessages> GetMessagePumpFactory() { }
         public NServiceBus.Transports.OutboundRoutingPolicy GetOutboundRoutingPolicy() { }
         public System.Func<NServiceBus.Transports.ICreateQueues> GetQueueCreatorFactory() { }
         public NServiceBus.Transports.IManageSubscriptions GetSubscriptionManager() { }
         public void Initialize(NServiceBus.Settings.SettingsHolder settings) { }
+        public async System.Threading.Tasks.Task<NServiceBus.Transports.StartupCheckResult> RunPreStartupChecks() { }
     }
     public class ForwardingTopologySectionManager : NServiceBus.AzureServiceBus.ITopologySectionManager
     {
@@ -440,13 +440,13 @@ namespace NServiceBus.AzureServiceBus
     {
         bool HasNativePubSubSupport { get; }
         bool HasSupportForCentralizedPubSub { get; }
-        System.Threading.Tasks.Task<NServiceBus.Transports.StartupCheckResult> ApplyPreStartupChecks();
         System.Func<NServiceBus.Transports.IDispatchMessages> GetDispatcherFactory();
         System.Func<NServiceBus.Transports.IPushMessages> GetMessagePumpFactory();
         NServiceBus.Transports.OutboundRoutingPolicy GetOutboundRoutingPolicy();
         System.Func<NServiceBus.Transports.ICreateQueues> GetQueueCreatorFactory();
         NServiceBus.Transports.IManageSubscriptions GetSubscriptionManager();
         void Initialize(NServiceBus.Settings.SettingsHolder settings);
+        System.Threading.Tasks.Task<NServiceBus.Transports.StartupCheckResult> RunPreStartupChecks();
     }
     public interface ITopologySectionManager
     {
@@ -537,13 +537,13 @@ namespace NServiceBus.AzureServiceBus
         public StandardTopology() { }
         public bool HasNativePubSubSupport { get; }
         public bool HasSupportForCentralizedPubSub { get; }
-        public async System.Threading.Tasks.Task<NServiceBus.Transports.StartupCheckResult> ApplyPreStartupChecks() { }
         public System.Func<NServiceBus.Transports.IDispatchMessages> GetDispatcherFactory() { }
         public System.Func<NServiceBus.Transports.IPushMessages> GetMessagePumpFactory() { }
         public NServiceBus.Transports.OutboundRoutingPolicy GetOutboundRoutingPolicy() { }
         public System.Func<NServiceBus.Transports.ICreateQueues> GetQueueCreatorFactory() { }
         public NServiceBus.Transports.IManageSubscriptions GetSubscriptionManager() { }
         public void Initialize(NServiceBus.Settings.SettingsHolder settings) { }
+        public async System.Threading.Tasks.Task<NServiceBus.Transports.StartupCheckResult> RunPreStartupChecks() { }
     }
     public class StandardTopologySectionManager : NServiceBus.AzureServiceBus.ITopologySectionManager
     {
--- a/src/Tests/API/APIApprovals.ApproveAzureServiceBusTransport.approved.txt
+++ b/src/Tests/API/APIApprovals.ApproveAzureServiceBusTransport.approved.txt
@@ -254,7 +254,7 @@ namespace NServiceBus.AzureServiceBus
         public ForwardingTopology() { }
         public bool HasNativePubSubSupport { get; }
         public bool HasSupportForCentralizedPubSub { get; }
-        public NServiceBus.Transports.StartupCheckResult ApplyPreStartupChecks() { }
+        public async System.Threading.Tasks.Task<NServiceBus.Transports.StartupCheckResult> ApplyPreStartupChecks() { }
         public System.Func<NServiceBus.Transports.IDispatchMessages> GetDispatcherFactory() { }
         public System.Func<NServiceBus.Transports.IPushMessages> GetMessagePumpFactory() { }
         public NServiceBus.Transports.OutboundRoutingPolicy GetOutboundRoutingPolicy() { }
@@ -373,7 +373,6 @@ namespace NServiceBus.AzureServiceBus
     public interface INamespaceManager
     {
         System.Uri Address { get; }
-        bool HasManageRights { get; }
         Microsoft.ServiceBus.NamespaceManagerSettings Settings { get; }
         System.Threading.Tasks.Task CreateQueue(Microsoft.ServiceBus.Messaging.QueueDescription description);
         System.Threading.Tasks.Task<Microsoft.ServiceBus.Messaging.SubscriptionDescription> CreateSubscription(Microsoft.ServiceBus.Messaging.SubscriptionDescription subscriptionDescription, string sqlFilter);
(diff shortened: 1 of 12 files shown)


src/Tests/API/APIApprovals.ApproveAzureServiceBusTransport.approved.txt
```diff
@@ -434,10 +436,15 @@ namespace NServiceBus.AzureServiceBus
         System.Threading.Tasks.Task RouteBatch(NServiceBus.AzureServiceBus.Batch batch, NServiceBus.AzureServiceBus.ReceiveContext receiveContext);
         System.Threading.Tasks.Task RouteBatches(System.Collections.Generic.IEnumerable<NServiceBus.AzureServiceBus.Batch> outgoingBatches, NServiceBus.AzureServiceBus.ReceiveContext receiveContext);
     }
+	public interface IStartupCheck
+    {
+        NServiceBus.Transports.StartupCheckResult Apply();
```
SeanFeldman: Nitpicking here... what about `Execute` or `Run`?

src/Tests/API/APIApprovals.ApproveAzureServiceBusTransport.approved.txt
```diff
@@ -434,10 +436,15 @@ namespace NServiceBus.AzureServiceBus
...
+    }
     public interface ITopology
     {
         bool HasNativePubSubSupport { get; }
         bool HasSupportForCentralizedPubSub { get; }
+		NServiceBus.Transports.StartupCheckResult ApplyPreStartupChecks();
```
SeanFeldman: Same as above

src/Transport/Config/AzureServiceBusTransport.cs
```diff
@@ -20,7 +20,7 @@ protected override TransportReceivingConfigurationResult ConfigureForReceiving(T
             return new TransportReceivingConfigurationResult(
                 Topology.GetMessagePumpFactory(),
                 Topology.GetQueueCreatorFactory(),
-                () => Task.FromResult(StartupCheckResult.Success) //TODO: figure out what this is for
+                () => Task.FromResult(Topology.ApplyPreStartupChecks())
```
SeanFeldman: :+1: like that

src/Transport/Connectivity/NamespaceManagerAdapter.cs
```diff
@@ -19,6 +19,22 @@ public NamespaceManagerAdapter(NamespaceManager manager)
 
         public Uri Address => _manager.Address;
 
+        public bool HasManageRights
+        {
+            get
+            {
+                try
+                {
+                    _manager.GetQueues();
```
SeanFeldman: Should async version used instead of sync?

src/Transport/Topology/ITopology.cs
```diff
@@ -12,9 +16,87 @@ public interface ITopology {
...
         OutboundRoutingPolicy GetOutboundRoutingPolicy();
         bool HasNativePubSubSupport { get; }
         bool HasSupportForCentralizedPubSub { get;}
     }
+
+    // NOTE: where do we move this components?
+    public interface IStartupCheck
+    {
+        // NOTE: to make awaitable?
+        StartupCheckResult Apply();
```
SeanFeldman: If startup check has to do IO (and it will), default should be async.
@yvesgoeleven any objections?

src/Transport/Topology/ITopology.cs
```diff
@@ -12,9 +16,87 @@ public interface ITopology {
...
+
+        public StartupCheckResult Apply()
+        {
+            var namespaces = settings.Get<List<string>>(WellKnownConfigurationKeys.Topology.Addressing.Partitioning.Namespaces);
+            var endpointName = settings.Get<EndpointName>();
+            
+            var configuredNamespaces = namespacePartitioningStrategy.GetNamespaces(endpointName.ToString(), PartitioningIntent.Creating);
+
+            return namespaces.Count != configuredNamespaces.Count() ?
+                StartupCheckResult.Failed("...") : StartupCheckResult.Success;
```
SeanFeldman: "..." you know what this check is for :)

src/Transport/Topology/ITopology.cs
```diff
@@ -12,9 +16,87 @@ public interface ITopology {
...
+        {
+            if (!settings.Get<bool>(WellKnownConfigurationKeys.Core.CreateTopology))
+                return StartupCheckResult.Success;
+
+            var hasNotRights = settings.Get<List<string>>(WellKnownConfigurationKeys.Topology.Addressing.Partitioning.Namespaces)
+                .Select(x => manageNamespaceManagerLifeCycle.Get(x))
+                .Any(x => !x.HasManageRights);
+
+            return hasNotRights ?
+                StartupCheckResult.Failed("...") : StartupCheckResult.Success;
```
SeanFeldman: Same as above

src/Transport/Topology/Topologies/ForwardingTopology.cs
```diff
@@ -109,6 +110,14 @@ public Func<IDispatchMessages> GetDispatcherFactory()
             return () => container.Resolve<IDispatchMessages>();
         }
 
+        public StartupCheckResult ApplyPreStartupChecks()
+        {
+            var checks = container.ResolveAll<IStartupCheck>();
+            var compositeStartupChecker = new CompositeStartupCheck(checks);
```
SeanFeldman: Something here feels odd @melkio. We register all the checks in `ApplyDefaults` to resolve those in `ApplyPreStartupChecks` so that we can construct `CompositeStartupCheck`?

src/Transport/Topology/Topologies/StandardTopology.cs
```diff
@@ -105,6 +108,14 @@ public Func<IDispatchMessages> GetDispatcherFactory()
             return () => container.Resolve<IDispatchMessages>();
         }
 
+        public StartupCheckResult ApplyPreStartupChecks()
+        {
+            var checks = container.ResolveAll<IStartupCheck>();
+            var compositeStartupChecker = new CompositeStartupCheck(checks);
```
SeanFeldman: Same as above

src/Transport/Topology/Topologies/ForwardingTopology.cs
melkio: It was a first attempt and I didn't want to introduce many changes :) 
In my opinion we could give to users the possibilities to add "pre startup checks" extending configuration API and then we can resolve all configured checks as I did, without configuring them in the ApplyDefaults method, removing also duplicated code between StandardTopology and ForwardingTopology.

If you agree, I can elaborate on this option.

src/Transport/Topology/ITopology.cs
melkio: yes, obviously :)

A question: is this check useless? All checks have already applied when we resolve namespace partitioning strategy. Right?

SimonCropp:
do we need to have the IStartupCheck  implemented this way? are the list of startup checks known? cant we just call them and not bother with putting them in the container and resolving them?

src/Transport/Topology/ITopology.cs
SeanFeldman: Not sure I follow @melkio. Could you link to the code?
Resolving namespace partitioning strategies returns the objects, doesn't operate on those. Or did you mean something else?

src/Transport/Topology/Topologies/ForwardingTopology.cs
SeanFeldman: > In my opinion we could give to users the possibility to add "pre startup checks" 

You do that by allowing custom topology to implement a pre-startup check.
Beyond that it feels as if we're exposing infrastructure hooks. Does it belong here?

src/Transport/Topology/ITopology.cs
melkio: What I'm trying to say is that each class that implement INamespacePartitioningStrategy interface ensures validity checks into its constructor. Ex: [ReplicatedNamespacePartitioningStrategy](https://github.com/Particular/NServiceBus.AzureServiceBus/blob/develop/src/Transport/Addressing/Partitioning/Strategies/ReplicatedNamespacePartitioningStrategy.cs#L14)

If I resolve INamespacePartitioningStrategy implementation with the container, validity checks have already applied, and my StartupCheck is useless. Right? Or am I misunderstending something?

melkio:
@SimonCropp if we want to give to the users the possibility to add new custom pre-startup checks, we have to register checks into container through configuration API (and not as I did...but this is a first commit ;)). 
If pre-startup checks are "internal" and they aren't configurable by users, I agree with you.

src/Transport/Topology/Topologies/ForwardingTopology.cs
melkio: > You do that by allowing custom topology to implement a pre-startup check.

Do you mean that to add a single pre-startup check, a user is forced to implement from scratch a whole custom topology? Did I get it right?

I don't know the design concepts behind ASB trasport implementation...yet :). If you think that this behaviour is not indispensable, I'll make code easier without using container and building `CompositeStartupCheck` directly.

yvesgoeleven: Imo this doesn't need to be configurable by the user, we can do this without using a container.

src/Transport/Topology/ITopology.cs
SeanFeldman: Get it now. True, all the partitioning strategies have a built in check.

melkio:
@Particular/azure-maintainers ping!

SeanFeldman:
@melkio pong, the test is red :panda_face:

melkio:
@SeanFeldman failed tests are about ATT with forwarding topology. Do I have to check anything? Otherwise for me PR is completed. 

@Particular/azure-maintainers: final review

src/Transport/Connectivity/NamespaceManagerAdapter.cs
```diff
@@ -19,6 +19,20 @@ public NamespaceManagerAdapter(NamespaceManager manager)
 
         public Uri Address => _manager.Address;
 
+        public async Task<bool> HasManageRights()
```
SeanFeldman: Nitpicking here:  `HasManagementRights()` or `CanManageEntities()`?

src/Transport/Topology/PreStartupChecks/ManageRightsCheck.cs
```diff
@@ -0,0 +1,34 @@
...
+        {
+            if (!settings.Get<bool>(WellKnownConfigurationKeys.Core.CreateTopology))
+                return StartupCheckResult.Success;
+
+            var tasks = settings.Get<List<string>>(WellKnownConfigurationKeys.Topology.Addressing.Partitioning.Namespaces)
+                .Select(x => manageNamespaceManagerLifeCycle.Get(x))
+                .Select(async x => await x.HasManageRights().ConfigureAwait(false));
+
+            return (await Task.WhenAll(tasks).ConfigureAwait(false)).Any(x => !x) ?
+                StartupCheckResult.Failed($"Manage rights on namespace is required if {WellKnownConfigurationKeys.Core.CreateTopology} setting is true") :
```
SeanFeldman: Would it help the user if the error would specify what namespace(s) don't have the sufficient rights?

src/Transport/Topology/Topologies/ForwardingTopology.cs
```diff
@@ -109,6 +108,13 @@ public Func<IDispatchMessages> GetDispatcherFactory()
             return () => container.Resolve<IDispatchMessages>();
         }
 
+        public async Task<StartupCheckResult> RunPreStartupChecks()
+        {
+            var check = new ManageRightsCheck(this.container);
+
+            return await check.Run().ConfigureAwait(false);
```
SeanFeldman: Can return `check.Run()` w/o async/await

src/Transport/Topology/Topologies/StandardTopology.cs
```diff
@@ -105,6 +106,13 @@ public Func<IDispatchMessages> GetDispatcherFactory()
             return () => container.Resolve<IDispatchMessages>();
         }
 
+        public async Task<StartupCheckResult> RunPreStartupChecks()
+        {
+            var check = new ManageRightsCheck(this.container);
+
+            return await check.Run().ConfigureAwait(false);
```
SeanFeldman: Same as above

src/Transport/Topology/PreStartupChecks/ManageRightsCheck.cs
melkio: Good point. Next commit ;)

yvesgoeleven:
@melkio looks good to me, but needs a rebase before merging, feel free to do that and merge it in

Issue #13, opened 2015-01-27
New azure servicebus topology (ForwardingTopology)


particularbot:
Things to do:
- [x] Implementation
- [x] Acceptance tests
- [x] ~~Endurance testing~~ replaced with #106 
- [ ] Polymorphic events (in context of ForwardingTopology) (update samples) #53 
- [ ] Documentation
 - [ ] Forwarding topology (differences from `StandardTopology`)
 - [ ] MD5 vs SHA1 subscriptions (FIPS compliance) ==> upgrade guide(s) if needed
 - [ ] Topology limitation 
    - Single namespace - up to 2K endpoints with 2K events in the system
    - Topics cannot be partitioned with multiple rules per subscription (subject to future service change)

---

**Original message**:

<a href="https://github.com/yvesgoeleven"><img src="https://avatars.githubusercontent.com/u/471354?v=3" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [yvesgoeleven](https://github.com/yvesgoeleven)**
_12/15/2014 10:01:07 AM +00:00_

_Originally opened as https://github.com/Particular/NServiceBus.Azure/issues/226_

----

Implement an improved azure servicebus topology

- Endpoint reads from it's own set of queues (multiple namespaces)
- Topics with subscription per endpoint & filter per interested type internally with forwarding rule to endpoint queue

Relates to: 

https://github.com/Particular/NServiceBus.Azure/issues/198
https://github.com/Particular/NServiceBus.Azure/issues/65
https://github.com/Particular/NServiceBus.Azure/issues/37

yvesgoeleven:
Replaced by https://github.com/Particular/docs.particular.net/issues/1259

Issue #117, opened 2016-02-08
Remove ForwardTo on all entities


SeanFeldman:
Have ran into this while working on https://github.com/Particular/NServiceBus.AzureServiceBus/issues/13. As part of the `ForwardingTopology` we forward all messages from subscriptions to the input queues of the subscribers. At the same time, `AzureServiceBusSubscriptionCreator` uses a default `ForwardTo` that can be set up by users via `AzureServiceBusSubscriptionSettings`.

Questions:

1. ~~Should this only be set for Error and Audit queues?~~
1. ~~What will this API look like for customers (pseudo code)?~~
1. ~~Why would we have forwarding `ForwardTo` for subscriptions by default to error and audit queues?~~
1. We have `ForwardDeadLetteredMessagesTo` in place, should that be enough? Yes

@yvesgoeleven any thoughts?

Connect to #13

yvesgoeleven:
indeed, we should get rid of this setting, and figure out an api that allows the user to set forwarding for audit & error queue only

SeanFeldman:
I'd like this conversation to include @timbussmann 
At this point the discussion is solely about transport forwarding endpoint`s error and audit queues.

timbussmann:
> and figure out an api that allows the user to set forwarding for audit & error queue only

why do you need to do this? We not just respect the values configured in the `EndpointConfiguration`? If the user want's to change the queue, why does it require a transport specific alternative?

yvesgoeleven:
Looks like I misunderstood how audit queues were named by default, I though they were by endpoint, we can remove regular forwarding. ForwardDeadLetteredMessagesTo will do.

SeanFeldman:
Decision is to remove `ForwardTo` for **all** entities.

SeanFeldman:
Closed via https://github.com/Particular/NServiceBus.AzureServiceBus/commit/a08b423798736c8a58ef44d2e8a01d6d345421f7

PR #118, opened 2016-02-09
Forwarding topology pub/sub

SeanFeldman:
Issues:
* [x] Subscriptions should forward
* [x] Description on subscription
* [x] Subscriptions should not be listened to (messages forwarded to queues)
* [x] `When_distributing_an_event` with dynamic routing -- failing (subscribers are scaled out using `InstanceDiscriminator` - MSMQ specific test, disabling it.
* [x] Subscribtion creator for this topology needs to be different 
  * [x] spike multiple rules
  * [x] Implement

Connects to https://github.com/Particular/NServiceBus.AzureServiceBus/issues/13

### DIFF
--- a/src/Transport/Creation/AzureServiceBusSubscriptionCreator.cs
+++ b/src/Transport/Creation/AzureServiceBusSubscriptionCreator.cs
@@ -149,9 +149,7 @@
                    || existingDescription.EnableDeadLetteringOnFilterEvaluationExceptions != newDescription.EnableDeadLetteringOnFilterEvaluationExceptions
                    || existingDescription.MaxDeliveryCount != newDescription.MaxDeliveryCount
                    || existingDescription.EnableBatchedOperations != newDescription.EnableBatchedOperations
-                   || existingDescription.ForwardDeadLetteredMessagesTo != newDescription.ForwardDeadLetteredMessagesTo
-                   || existingDescription.ForwardTo != newDescription.ForwardTo;
-
+                   || existingDescription.ForwardDeadLetteredMessagesTo != newDescription.ForwardDeadLetteredMessagesTo;
         }
     }
 }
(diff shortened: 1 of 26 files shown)


SeanFeldman:
@Particular/azure-maintainers need help with the last item

SeanFeldman:
update: 
1. `When_distributing_an_event` was MSMQ specific, resolved by removing the test.
2. Original idea for forwarding subscription won't work as we can't combine all event filters in a single rule. Alternative is to create a rule per event.

src/Transport/Creation/AzureServiceBusSubscriptionCreator.cs
```diff
@@ -31,15 +31,22 @@ public AzureServiceBusSubscriptionCreator(ReadOnlySettings settings)
                     MaxDeliveryCount = setting.GetOrDefault<int>(WellKnownConfigurationKeys.Topology.Resources.Subscriptions.MaxDeliveryCount),
                     RequiresSession = setting.GetOrDefault<bool>(WellKnownConfigurationKeys.Topology.Resources.Subscriptions.RequiresSession),
                     
+                    // TODO: default forwarding for subscriptions is not clear... need to review this
                     ForwardTo = setting.GetConditional<string>(subscriptionName, WellKnownConfigurationKeys.Topology.Resources.Subscriptions.ForwardTo),
```
yvesgoeleven: I think we need to remove this configuration option, if forwarding is set the StandardTopo would not be able to read from the sub, if set on ForwardTopo it would not behave the way we expect it to, so I suggest to get rid of this setting for subscriptions

SeanFeldman:
I have not squashed the commits to make it easier to review.

src/Transport/Creation/AzureServiceBusSubscriptionCreator.cs
SeanFeldman: Will be addressed by a separate issue opened for this (#117).

yvesgoeleven:
Merged!

Issue #53, opened 2015-10-06
Polymorphic events


SeanFeldman:
Currently polymorphic events are ending up being sent multiple times due to inner workings of Azure Service Bus subscriptions.

## Example

**Polymorphic Messages**
`Shared.Events` (Namespace)
`EventA : IEvent`
`EventB : EventA`

**Handlers**
`HandlerA : IHandle<EventA>`
`HandlerB : IHandle<EventB>`

**Invocation**

`bus.Publish<EventB>(e => ...);`

**Expected behaviour**

Handler A and B to be invoked

**Actual behaviour**

Handler A and B are invoked twice each

## Analysis

What's happening is that endpoint publishing these events has 2 subscriptions: one for `EventA` and one for `eventB`. When message of type `EventB` is published, it will be sent to subscriber twice (once because default rule filter on `EventA` subscription will catch it with `[NServiceBus.EnclosedMessageTypes] LIKE '%Shared.Events.EventA%' OR...` and once because default rule filter on `EventB` subscription will catch it with `[NServiceBus.EnclosedMessageTypes] LIKE '%Shared.Events.EventB%' OR...`). This is because published message `NServiceBus.EnclosedMessageTypes` header is `Shared.Events.EventA, Shared, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null;Shared.Events.EventB, Shared, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null`
Since each event message will be delivered to an endpoint, each will go through polymorphic routing and will be invoked with event handlers that can handle it, i.e. `HandlerA` and `handlerB`. This will result in duplicate invocation of handlers.

## Workaround

1. Disable auto-subscription for subscriber endpoint 
1. Subscribe to base event only manually

[GG issue](https://groups.google.com/forum/#!topic/particularsoftware/g7roalODpLU)

Connect to #13

SeanFeldman:
@Particular/azure what do we do with this?

SeanFeldman:
We should implement polymorphic event dispatch using `SqlRuleAction`s. This will allow proper implementation w/o workarounds.

WestDiscGolf:
@SeanFeldman that doesn't sound like a quick fix, any idea when the issue will be resolved?

yvesgoeleven:
[...]

PR #122, opened 2016-02-19
Polymorphic events with forwarding topology

SeanFeldman:
Events should be published to a single topic in the bundle when `ForwardingTopology` is used.
Randomly selecting topic.

Went with the following assumptions:

1. For sends it doesn't matter, throughput limit is applied on the entire namespace level (check with ASB team)
1. Bundles should default to a single topic for simplicity
1. On the sending side: active topic in a bundle is selected randomly
1. On the receiving side: receive from all topics in the bundle

Addresses issue #53

### DIFF
--- a/packaging/nuget/nservicebus.azureservicebus.nuspec
+++ b/packaging/nuget/nservicebus.azureservicebus.nuspec
@@ -19,6 +19,6 @@
     </dependencies>
   </metadata>
   <files>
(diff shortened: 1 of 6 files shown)


SeanFeldman:

SeanFeldman:
Flaky test fixed.
@Particular/azure-maintainers need a 2nd review

Issue #93, opened 2016-01-26
Map namespace name to connectionstring


yvesgoeleven:
In order to not leak security sensitive information in connectionstrings out, we need a map of logical level namespace name to physical level namespace connectionstring, so that we only need to pass the logical name around.

This will fix the following:

- [x] Ensure ReplyTo does not include shared access keys https://github.com/Particular/NServiceBus.AzureServiceBus/issues/16
- [x] Ensure there are no SharedAccessKey in Logs https://github.com/Particular/NServiceBus.AzureServiceBus/issues/17

SeanFeldman:
@melkio @yvesgoeleven would it help if we flesh out the api (pseudo code) to see how it will work.

Scooletz:
Similar work should be done in ASQ https://github.com/Particular/NServiceBus.AzureStorageQueues/issues/21
It might be obvious, but I'd add that the backward compatibility is required, so we should accept the addresses with the connection string as well, but sending only with logical.

melkio:
I'm missing something for sure but I'd approach the issue in this way:

- changing configuration API to register namespaces. Something like this:

```csharp
var extensions = new TransportExtensions<AzureServiceBusTransport>(settings);
extensions  
    .UseDefaultTopology()
    .Addressing()
    .NamespacePartitioning()
    .AddNamespace("name", "connection string");
```

- @yvesgoeleven: suggested to add a new property to NamespaceInfo to store namespace name and easyly lookup when needed

```csharp
public class NamespaceInfo
{
    public NamespaceInfo(string name, string connectionString, NamespaceMode mode = NamespaceMode.Active)
    {
        Name = name;
        ConnectionString = connectionString;
        Mode = mode;
    }

    public string Name { get; set; }
    public string ConnectionString { get; set; }
    public NamespaceMode Mode { get; set; }

    ....
}
```

- another option could be to stay on the current NamespaceInfo implementation and introduce a new internal component into transport infrastracture, that is able to translate from logical to physical when needed. Something like:

```csharp
public interface ICanTranslateFromLogicalToPhysicalNamespace
{
    string Translate(LogicalNamespace name)
}
```

SeanFeldman:
> It might be obvious, but I'd add that the backward compatibility is required, so we should accept the addresses with the connection string as well, but sending only with logical.

It depends. When sending to older endpoints, those need to receive the connection strings.

SeanFeldman:
[...]

PR #125, opened 2016-02-22
Map namespace name to connection string (v2)

melkio:
For security reasons, namespace connection string mustn't be logged.

What I did in details:
- I introduced the concept of "namespace name" at NamespaceInfo level and used it instead of connection string into transport until the code catches SB SDK objects that need "real" connection string.
- To check if everything works as expected, I configured ATT to log "debug level" to file to verify if full connection string is logged. Running all tests, log doesn't contain connection string.
- I configured transport into ATT to have multiple namespaces. We have to configure the first "connection string" with NSB core extension method, otherwise NSB crashes starting. We don't have control over this connection string, if NSB core decides to log it
- I added configuration option to switch between namespace name and namespace connection string. My idea was to use this switch to setup f.e. "ReplyTo" header, but looking at header's value I noticed it contains queue name and now I'm stuck.
- are there others points in which enforcing security and switching between name and connection string?

@Particular/azure-maintainers need help, maybe I'm missing something

Connects to #93


yvesgoeleven:
> _We have to configure the first "connection string" with NSB core extension method, otherwise NSB crashes starting_

There is a boolean on the transport definition 'RequiresConnectionString', maybe we should set this to false (as we have our own namespace check anyway)

yvesgoeleven:
> _but looking at header's value I noticed it contains queue name and now I'm stuck._

the header can contain queuename or queuename@something... only need to validate that 'something' part is not a connectionstring.

yvesgoeleven:
> _are there others points in which enforcing security and switching between name and connection string?_

Exceptions, headers & logs are the only ones I know off

src/Tests/API/APIApprovals.ApproveAzureServiceBusTransport.approved.txt
```diff
@@ -479,12 +479,22 @@ namespace NServiceBus.AzureServiceBus
         public System.Threading.Tasks.Task Send(Microsoft.ServiceBus.Messaging.BrokeredMessage message) { }
         public System.Threading.Tasks.Task SendBatch(System.Collections.Generic.IEnumerable<Microsoft.ServiceBus.Messaging.BrokeredMessage> messages) { }
     }
-    public class NamespaceInfo
+    public class NamespaceDefinition : System.IEquatable<NServiceBus.AzureServiceBus.NamespaceDefinition>
```
yvesgoeleven: Can you explain the reasoning behind having 2 objects here? Definition & Info?

src/Tests/API/APIApprovals.ApproveAzureServiceBusTransport.approved.txt
```diff
@@ -517,6 +527,16 @@ namespace NServiceBus.AzureServiceBus
         Active = 0,
         Passive = 1,
     }
+    public class NamespacesDefinition : System.Collections.Generic.IEnumerable<NServiceBus.AzureServiceBus.NamespaceDefinition>, System.Collections.IEnumerable
```
yvesgoeleven: Rename to NamespaceDefinitions

src/Transport/Topology/MetaModel/NamespaceInfo.cs
```diff
@@ -1,45 +1,36 @@
 ﻿namespace NServiceBus.AzureServiceBus
 {
-    public class NamespaceInfo
+    using System;
+
+    // NOTE: Can namespaces switch mode?
```
yvesgoeleven: Yes, f.e. sharded will select active one based on sharding rule

src/Transport/Config/DefaultConfigurationValues.cs
```diff
@@ -34,7 +33,8 @@ void ApplyDefaultValuesForSerialization(SettingsHolder settings)
 
         void ApplyDefaultValuesForAddressing(SettingsHolder settings)
         {
-            settings.SetDefault(WellKnownConfigurationKeys.Topology.Addressing.Partitioning.Namespaces, new List<string>());
+            settings.SetDefault(WellKnownConfigurationKeys.Topology.Addressing.Partitioning.Namespaces, new NamespacesDefinition());
+            settings.SetDefault(WellKnownConfigurationKeys.Topology.Addressing.UseLogicalNamespaceName, (Func<NamespaceDefinition, string>)(x => x.ConnectionString));
```
yvesgoeleven: Why is this a func and not just a bool?

src/Transport/Addressing/Partitioning/Strategies/SingleNamespacePartitioningStrategy.cs
```diff
@@ -7,32 +7,20 @@ namespace NServiceBus.AzureServiceBus.Addressing
...
-                connectionstrings = new List<string>();
-            }
-
-            if (connectionstrings.Count == 0)
+            if (!settings.TryGet(WellKnownConfigurationKeys.Topology.Addressing.Partitioning.Namespaces, out _namespaces) || _namespaces.Count != 1)
             {
                 throw new ConfigurationErrorsException("The 'Single' namespace partitioning strategy requires exactly one namespace, please configure the connection string to your azure servicebus namespace");
             }
-
-            if (connectionstrings.Count > 1)
```
yvesgoeleven: Why is this check removed?

src/Tests/API/APIApprovals.ApproveAzureServiceBusTransport.approved.txt
melkio: I introduced two new objects: NamespaceDefinition and Namespace**s**Definition.
The first one contains mapping between namespace name and connectionString, the second one contains all the namespace configured in the config phase and it uses the first "value object" as component to store these infos.

Given that, NamespaceDefinition has behaviours to check equality and to setup mapping (name-connectionString). I used the same component into NamespaceInfo to avoid duplicated code. 

I can't use NamespaceInfo during transport configuration phase

src/Tests/API/APIApprovals.ApproveAzureServiceBusTransport.approved.txt
melkio: Ok, maybe it's my english fault :) but this object is about defining one or more namespaces. It's not about defining the same namespace several times.

Is it better NamespaceDefinition**s**?

src/Transport/Addressing/Partitioning/Strategies/SingleNamespacePartitioningStrategy.cs
melkio: The two checks have been joined into one.
Before they were:
`connectionStrings.Count == 0 or connectionStrings.Count > 1`

Now it's:
`connectionStrings.Count != 1`

src/Transport/Config/DefaultConfigurationValues.cs
melkio: Because when I started working on this issue, I thought we need a way to switch between using name or connectionString. Through config API we setup the resolver func and use it in the transport avoiding `if` to check if feature has been enabled.

f.e. I thought to use this func to calculate the right value to put into ReplyTo header. If `UseLogicalNamespaceName`  feature has been enabled, ReplyTo header contains namespace name instead of connectionString. But now I'm struggling with the goal of the issue, because I noticed that ReplyTo header contains the queue name and it doesn't contain connectionString

src/Transport/Topology/MetaModel/NamespaceInfo.cs
melkio: Ok, but why does project compile anyway?
I converted NamespaceInfo into a plain "value object", making `Mode` property readonly and everything works well (project build and tests). Does someone change namespace mode?

melkio: Ah ok, got it: we don't need to change `Mode` property of the namespace instance and NamespaceInfo can be a "value object". 
f.e. for `ShardedNamespacePartitioningStrategy` we create a new NamespaceInfo instance (with right values) each time

melkio: I'll remove the note!

melkio:
> > _but looking at header's value I noticed it contains queue name and now I'm stuck._
> 
> the header can contain queuename or queuename@something... only need to validate that 'something' part is (or not) a connectionstring.

When  does this happen? How can I simulate this behavior (having queuename@something) into ReplyTo header?

src/Tests/API/APIApprovals.ApproveAzureServiceBusTransport.approved.txt
yvesgoeleven: I see, one is at configuration time, the other at runtime. Maybe we could go for different names here to better reflect that (NamespaceConfiguration and NamespaceInstance?), thoughts @SeanFeldman ?

src/Tests/API/APIApprovals.ApproveAzureServiceBusTransport.approved.txt
yvesgoeleven: yes

src/Transport/Addressing/Partitioning/Strategies/SingleNamespacePartitioningStrategy.cs
yvesgoeleven: ok, missed that...

src/Transport/Config/DefaultConfigurationValues.cs
yvesgoeleven: check how ASQ is doing it, you can apply the same logic imo

- incoming: https://github.com/Particular/NServiceBus.AzureStorageQueues/pull/42/files#diff-2527dbd34187f07376b335d39883a92fR151
- outgoing: https://github.com/Particular/NServiceBus.AzureStorageQueues/pull/42/files#diff-582a0cef3422849f67816681c602843dR114

yvesgoeleven:
you can set it in message endpoint mappings, type="WhateverCommand" endpoint="sales@connectionstring"

melkio:
The desired behavior is:

- if `UseLogicalNamespaceName` isn't enabled, do nothing. It's safe for backward compatibility
- if replyTo header value doesn't contain connectionString, do nothing
- if `UseLogicalNamespaceName` is enabled and ReplyTo header value contains queueName@connectionString the transport should replace connectionString with mapped namespace name for outgoing message or throws if it's not mapped
- if `UseLogicalNamespaceName` is enabled and ReplyTo header value of incoming message contains queueName@namespaceName the transport should replace namespaceName with configured connectionString or throws if it's not mapped
- _if `UseLogicalNamespaceName` is enabled and ReplyTo header value of incoming message contains queueName@connectionString, should the transport throws an exception ?_

@Particular/azure-maintainers: any feedback? Is it right?

SeanFeldman:
@yvesgoeleven @melkio should `UseConnectionString()` w/o name be obsoleted? 
Similar to this: https://github.com/Particular/NServiceBus.SqlServer/pull/167/files

SeanFeldman:
@melkio in regards to the last bullet point with question, this could be an older endpoint talking to the new one. In that case throwing an exception would not be good.

src/Tests/API/APIApprovals.ApproveAzureServiceBusTransport.approved.txt
SeanFeldman: We've agreed to go with the following change:
`NamespaceInfo` to be renamed as `RuntimeNamespaceInfo`
and
`NamespaceDefinition` to be renamed as `NamespaceInfo`

src/Transport/Topology/MetaModel/NamespacesDefinition.cs
```diff
@@ -0,0 +1,69 @@
...
+    using System.Collections;
+    using System.Collections.Generic;
+    using System.Linq;
+
+    public class NamespacesDefinition : IEnumerable<NamespaceDefinition>
+    {
+        private static readonly string DefaultName = "default";
+        private static readonly string Prefix = "namespace-";
+
+        private readonly List<NamespaceDefinition> _inner;
```
SeanFeldman: Should this be a concurrent dictionary?

src/Transport/Topology/MetaModel/NamespacesDefinition.cs
```diff
@@ -0,0 +1,69 @@
...
+        public int Count => _inner.Count;
+
+        public void Add(string name, string connectionString)
+        {
+            var definition = new NamespaceDefinition(name, connectionString);
+            if (_inner.Contains(definition)) return;
+
+            var defaultDefinition = _inner.SingleOrDefault(x => x.Name == DefaultName && x.ConnectionString == connectionString);
+            _inner.Remove(defaultDefinition);
+            _inner.Add(definition);
```
SeanFeldman: why do we remove the default to add it right after?

yvesgoeleven:
@SeanFeldman we don't have a UseConnectionString do we?

SeanFeldman:
[Here](https://github.com/Particular/docs.particular.net/blob/master/samples/azure/azure-service-bus/Version_6/Endpoint1/Program.cs#L20) @yvesgoeleven. Maybe it is coming from the core, still needs to be Obsoleted (separate issue on core?)

yvesgoeleven:
@SeanFeldman that's core, feel free to open a pr, but I guess all other transports rely on it?

SeanFeldman:
I'd say as a part of this PR would be good to investigate it.

melkio:
@SeanFeldman what you said (about my latest question) is not a possible combination.
Users can't configure a mixed environment: new endpoint with `UseLogicalNamespaceName` enabled and older endpoints.

src/Transport/Topology/MetaModel/NamespacesDefinition.cs
melkio: I think no: NamespacesDefinition is configured during endpoint and transport configuration which is a "single thread" operation (I suppose).
Once configured, NamespacesDefinition is "multi thread readable"

src/Transport/Topology/MetaModel/NamespacesDefinition.cs
melkio: Through config API, users can setup namespaces in two different ways: adding namespace explicitly or using ConnectionString core extension method.
In the develop branch version, throughout transport set up, we add "default" connection string to namespaces if it doesn't exist. I put this logic inside NamespacesDefinition class: adding a new namespace definition (name + connectionString), if an identical connection string marked as default already exists, I replace it with a more specific namespace definition

melkio:
We use `ConnectionString` method to setup transport in ATT :) and maybe is the simplest way to setup transport as `SingleNamespacePartitionStrategy`
I think trasnports that don't support "multiple connection string" use this method. No way to mark it obsolete

I'll investigate

src/Transport/Topology/MetaModel/NamespacesDefinition.cs
SeanFeldman: Got it, thank you.
Does that mean something like this would take place? 
```c#
config.UseTransport<ASB>()
   .ConnectionString("defaultconn")
   .NamespacePartitioning().AddNamespace("defaultconn");
```
Should we throw an exception saying that there's a conflict between default and named namespace. Since it's configuration time, owner of the configuration should only register it once. Thoughts?

melkio:
Finally completed...I hope / @Particular/azure-maintainers 

Summary:

- a new config API to enable namespace mapping (`UseNamespaceNamesInsteadOfConnectionStrings`) on `Addressing` level
- each brokered message goes through a mapper (`ICanMapConnectionStringToNamespaceName` implementation) before becoming an incoming message: if reply to address header value exists then the mapper is applied
- each outgoing message goes through a mapper (`ICanMapNamespaceNameToConnectionString` implementation) before becoming a brokered message: if reply to address header value exists and `UseNamespaceNamesInsteadOfConnectionStrings` feature hasn't been enabled then the mapper is applied

NOTE: I introduced a `ConnectionString` class. This class contains logic to validate provided value:

- namespace name lenght between 6 and 50 chars
- namespace name starts with a letter
- namespace name contains letters, numbers and hyphens

Do you agree with this component? For me is useful to evaluate if `ReplyToAddress` header value contains a valid connection string, but we can introduce this component "everywhere" we use a "simple" string as connection string to enforce validity (f.e. `NamespaceInfo` or `RuntimeNamespaceInfo`)

Issue #103, opened 2016-01-26
Implement passive namespace fallback


yvesgoeleven:
Only active namespaces are used now for sending messages, implement fallback to passive if the active ones didn't work

mauroservienti:
In today sync was asked if this is mandatory for V6 or can be release as an update to V6.
Also @Particular/platform-dev-squad this is in progress but is lacking a proper task force.

yvesgoeleven:
@mauroservienti relabled as bug, the work for this had already been done as part of the multi namespace support, but wasn't verified yet, now it is.

yvesgoeleven:
@Particular/azure-maintainers PR https://github.com/Particular/NServiceBus.AzureServiceBus/pull/129 ready for review

yvesgoeleven:
@Particular/azure-maintainers note: for local testing you need an additional environment variable containing a fallback namespace, build server has these already

PR #129, opened 2016-02-29
Passive fallback

yvesgoeleven:
Implementation & tests for passive namespace fallback

Connects to #103

### DIFF
--- a/src/Transport/Sending/DefaultOutgoingBatchRouter.cs
+++ b/src/Transport/Sending/DefaultOutgoingBatchRouter.cs
@@ -121,6 +121,10 @@ namespace NServiceBus.AzureServiceBus
                 var message = "Failed to dispatch a batch with the following message IDs: " + string.Join(", ", messagesToSend.Select(x => x.MessageId));
                 logger.Error(message, exception);
 
+                // no need to try and send too large messages to another namespace, won't work
+                if (exception is MessageTooLargeException)
+                    throw;
+
                 var fallBackSucceeded = false;
                 if (fallbacks.Any())
                 {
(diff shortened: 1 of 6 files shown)


src/Transport/Sending/DefaultOutgoingBatchRouter.cs
```diff
@@ -115,7 +120,32 @@ private string GetViaEntityPathFor(EntityInfo entity)
...
+                        try
+                        {
+                            await RouteBatchWithEnforcedBatchSizeAsync(fallback, clones).ConfigureAwait(false);
+                            logger.Info("Successfully dispatched a batch with the following message IDs: " + string.Join(", ", clones.Select(x => x.MessageId) + " to fallback namespace"));
+                            fallBackSucceeded = true;
+                            break;
+                        }
+                        catch (Exception ex)
+                        {
+                            logger.Error("Failed to dispatch a batch with the following message IDs: " + string.Join(", ", clones.Select(x => x.MessageId) + " to fallback namespace"), ex);
```
SeanFeldman: We've already logged that error [here](https://github.com/Particular/NServiceBus.AzureServiceBus/pull/129/files#diff-d1b2d15f9b111a00dedb967449c6a732R121). Perhaps this error message should state that we've failed to send sing fallback namespace X?

SeanFeldman:
Should we rebase prior to merging to have a more linear history?

src/Transport/Sending/DefaultOutgoingBatchRouter.cs
yvesgoeleven: we don't have the namespace name here, but I'll remove the message id's

Issue #92, opened 2016-01-26
Code only configuration API


yvesgoeleven:
To be consistent with core v6 & the rest of the transports & persistences, we should switch to a code only configuration model on top of SettingsHolder.

Todo: should we provide an overlay for backward compat of xml based config model as well?

TODO
- [x] The new configuration API needs simplification
  - [x] `EnablePartitioning` for topics on `ForwardingTopology` #131 
  - [x] `ConnectivityMode` is missing #130 
  - [x] `MaxDeliveryCount` #137
  - [x] `PrefetchCount` on both `MessaginFactory` and `MessageReceiver` #151

yvesgoeleven:
https://github.com/Particular/NServiceBus.AzureServiceBus/issues/20

yvesgoeleven:
The new configuration API needs simplification & doco

SeanFeldman:
For XML based configuration, we could provide a sample to show how to do it. 
Same is applicable for Cloud Services configuration (.cscfg)

SeanFeldman:
1. `ConnectivityMode` - not implemented
1. `ServerWaitTime` - not implemeted
1. `BackoffTimeInSeconds` - this used to be on the receiving poll. We're now using msgReceiver.OnMessage, so think it's not needed (do you know?)
1. `BatchSize` -- not sure where's this one in the v7
1. `EnablePartitioning` won't work with ForwardingTopo :disappointed:
1. `SupportOrdering` -- i question the value of this one

@yvesgoeleven can you comment of these?

yvesgoeleven:
1. Definitly needed, we need to add this back
2. Not needed anymore
3. Not needed anymore
4. Uses PrefetchCount setting now, so not needed anymore
5. Add topology specific startup check?
6. Why? It's easier to reason about queue content if it doesn't change order...

SeanFeldman:
> It's easier to reason about queue content if it doesn't change order...

It's defined for both queues and topics. We say "order is not promised" and at the same time we're exposing this bare-metal configuration. Confusing, isn't it?

yvesgoeleven:
Hmm, it is, remove it then...

SeanFeldman:
[...]

PR #136, opened 2016-03-02
ConnectivityMode implementation

### LINKED ISSUE #130: ConnectivityMode setting missing
Looks like we accidentally removed support for the ConnectivityMode setting, need to add it back and ensure it works

SeanFeldman:
Fixes #130.

- [Default mode](https://github.com/Particular/NServiceBus.AzureServiceBus/issues/130#issuecomment-190981534) is `AutoDetect` and not `Tcp`

Connects to #92

### DIFF
--- a/src/Transport/Config/AzureServiceBusTransport.cs
+++ b/src/Transport/Config/AzureServiceBusTransport.cs
@@ -1,5 +1,6 @@
 ﻿namespace NServiceBus
 {
+    using Microsoft.ServiceBus;
     using NServiceBus.AzureServiceBus;
     using NServiceBus.Settings;
     using NServiceBus.Transports;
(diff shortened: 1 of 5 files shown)


src/Transport/Config/AzureServiceBusTransport.cs
```diff
@@ -17,10 +18,17 @@ protected override TransportInfrastructure Initialize(SettingsHolder settings, s
 
             RegisterConnectionStringAsNamespace(connectionString, settings);
 
+            SetConnectivityMode(settings);
```
melkio: nitpicking here: why not inline?

SeanFeldman: Good question @melkio! It was initially, but I've realized that one can't see the forest for the trees. Similar to `RegisterConnectionStringAsNamespace` that could be done as 
```c#
var namespaces = settings.Get<NamespaceConfigurations>(WellKnownConfigurationKeys.Topology.Addressing.Partitioning.Namespaces);
              namespaces.AddDefault(connectionString).AddDefault(connectionString)
```
but the readability of that is horrendous. I.e. "tell me what it does, not how" approach :)

Issue #137, opened 2016-03-02
MaxDeliveryCount default value


SeanFeldman:
`MaxDeliveryCount` is hard-coded to 6 attempts. The default for ASB has changed to 10.

Should `MaxDeliveryCount` value be based on endpoint`s FLR/SLR values rather than ASB default?

- [x] Fix the default
- [x] Allow to override the default

SeanFeldman:
@yvesgoeleven I'm raising this as a bug for now. It could be another `Core v6` issue that didn't make it on the general list  #90.

SeanFeldman:
I'm going to leave this as-is and make a comment in the code that once decision on https://github.com/Particular/NServiceBus/issues/3354 is made, we can re-open this issue if needed.

SeanFeldman:
Fixed by https://github.com/Particular/NServiceBus.AzureServiceBus/commit/5042238324c8af0a677ad36c29ed95aac9093ebd

PR #138, opened 2016-03-02
MaxDeliveryCount default value

SeanFeldman:
Addresses issue #137 

Connects to #92

### DIFF
--- a/src/Transport/Config/DefaultConfigurationValues.cs
+++ b/src/Transport/Config/DefaultConfigurationValues.cs
@@ -67,7 +67,7 @@
             settings.SetDefault(WellKnownConfigurationKeys.Topology.Resources.Queues.EnableDeadLetteringOnMessageExpiration, false);
             settings.SetDefault(WellKnownConfigurationKeys.Topology.Resources.Queues.DuplicateDetectionHistoryTimeWindow, TimeSpan.FromMilliseconds(600000));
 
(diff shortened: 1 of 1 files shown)


SeanFeldman:
@yvesgoeleven you've commented out the original code that used FLR/SLR. What was the reason?

yvesgoeleven:
@SeanFeldman initialy I thought/reasoned that max delivery setting should be tied to FLR even when it's off. (aka 1 when no FLR). But in hindsight I think it's not true as FLR would move it to ErrorQ or SLR anyway.

SeanFeldman:
@yvesgoeleven the outcome of the conversation taking place [here](https://github.com/Particular/NServiceBus/issues/3354) will influence this issue. For now I was thinking that `MaxDeliveryCount` should not be set to the default of ASB native service, but to the calculation to resolve number of total retries (influenced by FLR  and SLR) based on the [calculation in our doco](http://docs.particular.net/nservicebus/errors/automatic-retries#total-number-of-possible-attempts). 

I'm not sure if going with the formula `Total Attempts = (FLR:MaxRetries) * (SLR:NumberOfRetries + 1)` is the right option either, since not every delivery of a message will necessarily result in a processing attempt. Any ideas?

yvesgoeleven:
@SeanFeldman should multiply by number of instances as well (and that is elastic), so you can't compute this up front

yvesgoeleven:
leave it as is, we can't fix this unless we get control of the FLR & SLR mechanics

Issue #135, opened 2016-03-01
Namespace registration API


SeanFeldman:
As of now, registration of namespaces is done in the following way:

```c#
var nsi1 = new NamespaceInfo("test", "conn");	
var nsi2 = new NamespaceInfo("Test", "conn");
```

which will result in `nsi1` and `ns2` to be treated as two different namespace information objects based on hashcode therefore not equal. 

```c#
nsi1.GetHashCode(); // 1120192152
nsi2.GetHashCode(); // 1120228920
```

By enforcing lower case we eliminate ambiguity and can prevent issues at the configuration time.

Two namespaces are equal if they have the same names (with case insensitive equality) and connectionstrings (with case insensitive equality on namespace name and shared access policy name)

Configuration API doesn't allow to register the same namespace twice

/cc @yvesgoeleven @melkio

Related issue: https://github.com/Particular/NServiceBus.AzureServiceBus/issues/93

melkio:
Good point @SeanFeldman 

A quick question before updating code: what does equality mean for `NamespaceInfo`? Same `names` or same `names` and `connectionStrings`?
Because connection string is built around three parameters:

- namespace name
- shared access policy name
- shared access policy key

`namespace name` and `shared access policy name` are case insensitive, but `shared access policy key` is case sensitive.

Introducing `NamespaceInfo` equality, considering `name` and `connectionString` is a bit harder

SeanFeldman:
I was focusing on the `name` only. Connection strings are usually copied as-is due to complexity. Names are typed, and that's where the chances are higher to make a mistake.

melkio:
Ok. Another question:
What about adding two namespaces with different names and the same connection string?

SeanFeldman:
This is a configuration API. If they register two namespace infos with different names, but same connection string, it could be intentional (one endpoint got deployed with a name `A`, another was deployed with a name `B`). Both look at the same namespace. A third endpoint needs to be able to talk to all three, `A`, `B`, and `C`. I'm talking hypothetically here though :)

SeanFeldman:
@melkio @yvesgoeleven please update issue with the agreed on behavious as per discussion.

yvesgoeleven:
we should check for duplicate connectionstrings, and not allow them. It would only result in multiple receivers on the same entities, simply a waste of connections

PR #139, opened 2016-03-03
Namespace API registration with case insensitive equality check on Name property

melkio:
Crossposting comment from issue: what about adding two namespaces with different names and the same connection string?

@Particular/azure-maintainers ready to review

Connects to #135

### DIFF
--- a/src/Tests/Topology/MetaModel/When_configuring_namespace_configurations.cs
+++ b/src/Tests/Topology/MetaModel/When_configuring_namespace_configurations.cs
@@ -41,43 +41,39 @@
         public void Should_add_definition_if_it_does_not_exist()
         {
             _namespaces.Add("name", ConnectionStringValue.Sample);
+            var connectionString = _namespaces.GetConnectionString("name");
 
             Assert.AreEqual(1, _namespaces.Count);
-        }
-
-        [Test]
-        public void Should_not_add_definition_if_exists()
-        {
-            _namespaces.Add("name", ConnectionStringValue.Sample);
-            _namespaces.Add("Name", ConnectionStringValue.Sample);
-
-            Assert.AreEqual(1, _namespaces.Count);
+            Assert.AreEqual(ConnectionStringValue.Sample, connectionString);
         }
 
         [Test]
         public void Should_add_default_connection_string_if_it_does_not_exist()
         {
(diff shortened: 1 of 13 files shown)


src/Tests/Topology/MetaModel/When_configuring_namespace_configurations.cs
```diff
@@ -49,7 +49,7 @@ public void Should_add_definition_if_it_does_not_exist()
         public void Should_does_not_add_definition_if_exists()
```
SeanFeldman: `Should_does_not_add_definition_if_exists()` ==> `Should_not_add_definition_if_exists()`

src/Transport/Topology/MetaModel/NamespaceInfo.cs
```diff
@@ -21,7 +21,7 @@ public NamespaceInfo(string name, string connectionString)
         public bool Equals(NamespaceInfo other)
         {
             return other != null
-                   && Name.Equals(other.Name)
+                   && Name.Equals(other.Name, StringComparison.InvariantCultureIgnoreCase)
```
SeanFeldman: `StringComparison.InvariantCultureIgnoreCase` should be replaced with `StringComparison.OrdinalIgnoreCase`

src/Tests/Topology/MetaModel/When_configuring_namespace_configurations.cs
melkio: Ops...sorry :)

SeanFeldman: All good  :)

melkio:
@Particular/azure-maintainers still WIP because this comment https://github.com/Particular/NServiceBus.AzureServiceBus/issues/135#issuecomment-191926743  isn't clear to me.

yvesgoeleven:
we should check for duplicate connectionstrings, and not allow them. It would only result in multiple receivers on the same entities, simply a waste of connections

melkio:
@Particular/azure-maintainers ready for review

SeanFeldman:
@melkio when ready, don't forget to remove `[wip]`

yvesgoeleven:
Great work!

Issue #115, opened 2016-02-08
Configure Publisher Names


yvesgoeleven:
Provide capability to configure publisher names at the transport level (where we used endpoint mappings in the past). Ensure this is mandatory for StandardTopology if there are subscriptions.

SeanFeldman:
i.e. add some sort of endpoint mapping via code for `StandardTopology` to utilize.

SeanFeldman:
Discussed this with @bording `UnicastRoutingTable` contains that information, why can't we make it public @SzymonPobiega?

SzymonPobiega:
Well, we can but I think what you need is the `Publishers` collection rather than `URT`. @SeanFeldman start a PR and let @Particular/nservicebus-maintainers comment on that. @yvesgoeleven and @andreasohlund were discussing that on a call but I don't remember what was the outcome.

SeanFeldman:
@melkio @yvesgoeleven please update issue with the outcome of the discussion

yvesgoeleven:
@SeanFeldman result is in the top description :)

SeanFeldman:
When `StandardTopology` is used, the Publishers for a subscribing endpoint **must** be configured.
When subscription is taking place and a publisher is not found in the list of the registered ones, what should be the expected behaviour? Assuming logging and throwing an exception (endpoint is expected to handle an event and it can't, critical error)?

SzymonPobiega:
I think that failure to subscribe on a native pub/sub transport should result in a critical error.

andreasohlund:
>I think that failure to subscribe on a native pub/sub transport should result in a critical error.

At least on rabbit failing to subscribe will cause an ex to be thrown from `.Subscribe` and that would cause the startup to fail. (we need to check how this behaves together with autosubscribe but I think it would prevent the endpoint from starting)

Is that what you are after @SzymonPobiega ?

SzymonPobiega:
@andreasohlund Initially I thought that if you call `endpoint.Subscribe` while the endpoint is running we should raise a critical error but now I am not so such. If a failure in autosubscribe prevents an endpoint from starting that seems good enough. And when I subscribe in runtime it might be better to not raise a critical but rather use a normal exception and let the user decide if he wants to kill the endpoint or not.

andreasohlund:
:+1: 

Also subscribing at runtime should be rare (smart client?) and in those cases it seems reasonable for the user to decide what to do in case of a failure

[...]

PR #152, opened 2016-03-14
[WIP] Configure publisher names

melkio:
@Particular/azure-maintainers I need feedback about implementation.

PR is "feature complete" but now I have to fix more than 30 acceptance tests because, introducing this feature, when we run ATT against `StandardTopology` we have to configure mapping between event types and publishers for each `scenario`

Connects to #115

### DIFF
--- a/src/AcceptanceTests/ConfigureEndpointAzureServiceBusTransport.cs
+++ b/src/AcceptanceTests/ConfigureEndpointAzureServiceBusTransport.cs
@@ -3,7 +3,13 @@ using System.Collections.Generic;
 using System.Threading.Tasks;
 using NServiceBus;
 using NServiceBus.AcceptanceTesting.Support;
+using NServiceBus.AcceptanceTests.ApiExtension;
+using NServiceBus.AcceptanceTests.BestPractices;
+using NServiceBus.AcceptanceTests.Routing;
+using NServiceBus.AcceptanceTests.Routing.AutomaticSubscriptions;
+using NServiceBus.AcceptanceTests.Sagas;
 using NServiceBus.AcceptanceTests.ScenarioDescriptors;
+using NServiceBus.AcceptanceTests.Versioning;
 using NServiceBus.AzureServiceBus;
 using NServiceBus.AzureServiceBus.AcceptanceTests.Infrastructure;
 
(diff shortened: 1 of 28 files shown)


yvesgoeleven:
Looks good to me!

melkio:
Mapping event types and publisher names is only for `standard topology`, but this feature is enabled for all possible topologies.
Should we introduce a sort of branch into configuration API in order to expose this feature only for `standard topology` (and topologies that need these mappings)?

yvesgoeleven:
Maybe somehow include the topology name in the config method? so that it is clear what it is referring to (ps: we're going to rename that topology)

src/Transport/Topology/PublishersConfiguration.cs
```diff
@@ -0,0 +1,76 @@
...
+        {
+            return new ReadOnlyCollection<string>(_publishers[type]);
+        }
+
+        public bool HasPublishersFor(Type type)
+        {
+            return _publishers.ContainsKey(type);
+        }
+
+        public static PublishersConfiguration ConfigureUsingThese(ReadOnlySettings settings)
```
SeanFeldman: I get what then name does, but it's a little weird. 
`ConfigureWith` or `ConfigureUsing`?

src/Tests/Topology/When_configuring_publishers.cs
```diff
@@ -0,0 +1,118 @@
...
+        public void SetUp()
+        {
+            var conventions = new FakeConventions();
+            _configuration = new PublishersConfiguration(conventions);
+        }
+
+        [Test]
+        public void Should_throw_if_mapping_does_not_exist()
+        {
+            Assert.Throws<KeyNotFoundException>(() => _configuration.GetPublishersFor(typeof(MyType)));
```
SeanFeldman: Should it throw a generic `KeyNotFoundException` or something that will contain more information to help to configure mappings properly?

src/Tests/Topology/When_configuring_publishers.cs
```diff
@@ -0,0 +1,118 @@
...
+        }
+
+        [Test]
+        public void Should_not_map_publisher_and_type_if_it_is_not_a_message()
+        {
+            _configuration.Map("publisher", typeof(MyType));
+
+            Assert.Throws<KeyNotFoundException>(() => _configuration.GetPublishersFor(typeof(MyType)));
```
SeanFeldman: Same as [here](https://github.com/Particular/NServiceBus.AzureServiceBus/pull/152/files#r56058246)

src/Transport/Topology/PublishersConfiguration.cs
SeanFeldman: `PublishersConfiguration` has a public constructor, but being initialized via this static construction method. Perhaps this code should be moved to ctor?

SeanFeldman:
@melkio in regards to [your question](https://github.com/Particular/NServiceBus.AzureServiceBus/pull/152#issuecomment-196334439) - what [Yves said](https://github.com/Particular/NServiceBus.AzureServiceBus/pull/152#issuecomment-196353924). Also, I wonder if we should throw if we see that topology being used/configured is not `EndpointOrientedTopology` (previously called `StandardTopology`)?

src/Transport/Config/ExtensionPoints/AzureServiceBusTopologySettings.cs
```diff
@@ -1,14 +1,48 @@
...
+            return this;
         }
 
+        public AzureServiceBusTopologySettings RegisterPublisherForAssembly(string publisherName, Assembly assembly)
+        {
+            AddScannerForPublisher(publisherName, new AssemblyTypesScanner(assembly));
+            return this;
+        }
 
+        private void AddScannerForPublisher(string publisherName, ITypesScanner scanner)
```
SeanFeldman: Fail if `EndpointOrientedTopology` (previously `StandardTopology`) is not set as the current one.

src/Tests/Topology/When_configuring_publishers.cs
melkio: Agree

src/Transport/Config/ExtensionPoints/AzureServiceBusTopologySettings.cs
melkio: We designed our config API to be fluent. Isn't better to give to users only allowed methods for selected topology?
There is an extension point to allow users to add a custom topology implementation. I thought that adding a property to `ITopology` interface we  can decide if a topology needs to configure `publisher names` vs. `event type` mapping or not.

src/Transport/Topology/PublishersConfiguration.cs
melkio: In my first attempt they were two different components: `PublishersConfiguration` to manage mapping between `publisher names` and `event types` and another component (a sort of factory/builder) to setup properly `PublishersConfiguration`. Then I simplified merging the two things in the same class with a builder method.

Making `PublishersConfiguration` .ctor private could be better?

melkio:
@Particular/azure-maintainers for me it's ready to be merged!

src/AcceptanceTests/ConfigureEndpointAzureServiceBusTransport.cs
```diff
@@ -33,7 +39,39 @@ public Task Configure(string endpointName, EndpointConfiguration config, RunSett
         {
             transportConfig.UseTopology<ForwardingTopology>();
         }
-        //else default
+        else
+        {
+            transportConfig.UseTopology<EndpointOrientedTopology>()
```
SeanFeldman: @melkio while this achieves the goal, I feel that hiding this information from the tests is not ideal... Can I suggest that each ATT tests that does pub/sub contain it's conditional mapping?

src/Tests/Topology/When_configuring_publishers.cs
```diff
@@ -0,0 +1,119 @@
+﻿namespace NServiceBus.Azure.WindowsAzureServiceBus.Tests.Topology
+{
+    using System;
+    using System.Linq;
+    using NServiceBus.AzureServiceBus;
+    using NUnit.Framework;
+
+    [TestFixture]
+    [Category("AzureServiceBus")]
+    public class When_configuring_publishers
```
SeanFeldman: `When_configuring_publishers` ==> `When_retrieving_publishers`

src/Tests/Topology/When_configuring_publishers.cs
```diff
@@ -0,0 +1,119 @@
...
+
+        [Test]
+        public void Should_throw_if_mapping_does_not_exist()
+        {
+            var exception = Assert.Throws<InvalidOperationException>(() => _configuration.GetPublishersFor(typeof(MyType)));
+            Assert.AreEqual($"No publishers configured for `{typeof(MyType).FullName}`", exception.Message);
+        }
+
+        [Test]
+        public void Should_not_map_publisher_and_type_if_it_is_not_a_message()
```
SeanFeldman: Did you mean "Should fail if type doesn't represent a message"?

src/Transport/Topology/IConventions.cs
```diff
@@ -0,0 +1,24 @@
+﻿namespace NServiceBus.AzureServiceBus
+{
+    using System;
+
+    interface IConventions
+    {
+        bool IsMessageType(Type type);
+    }
+
+    class ConventionsAdapter : IConventions
```
SeanFeldman: Navigating to `ConventionsAdapter` in GH won't be possible if it's sharing a file with another type.

src/Transport/Topology/PublishersConfiguration.cs
```diff
@@ -0,0 +1,82 @@
...
+            }
+
+            if (!publisherNames.Contains(publisherName))
+                publisherNames.Add(publisherName);
+        }
+    }
+
+    internal static class PublishersConfigurationExtensionMethods
+    {
+        public static void Map(this PublishersConfiguration configuration, string publisherName, IEnumerable<Type> types)
```
SeanFeldman: @melkio what's the value of having this as an extension method rather than on `PublisherConfiguration` directly?

src/Transport/Topology/Topologies/EndpointOrientedTopology.cs
```diff
@@ -90,6 +90,9 @@ private void InitializeContainer(SettingsHolder settings)
 
             var validationStrategyType = (Type)settings.Get(WellKnownConfigurationKeys.Topology.Addressing.Validation.Strategy);
             container.Register(validationStrategyType);
+
+            var publishersConfiguration = PublishersConfiguration.ConfigureWith(settings);
```
SeanFeldman: Don't like that we have a constructor and at the same time `ConfigureWith()` method. It's not aligned. I could easily do ` var publishersConfiguration = new PublishersConfiguration(settings)`, how would I know that I'm missing an important dependency?

If you want ctor to be the creation point, remove `ConfigureWith`. If `ConfigureWith` is the construction point, then don't expose a constructor and rename `ConfigureWith` to represent what it does: `Create`.

src/Tests/Topology/When_configuring_publishers.cs
melkio: It shouldn't fail. Simply `publishers configuration` doesn't register mapping if supplied type doesn't represent a message

src/Transport/Topology/Topologies/EndpointOrientedTopology.cs
melkio: :+1: 
I wrote the same thing on a previous comment, waiting for your replies https://github.com/Particular/NServiceBus.AzureServiceBus/pull/152#discussion_r56180854

src/Transport/Topology/PublishersConfiguration.cs
melkio: When I design the "public shape" of a class, I prefer to put there only what a user could need: `Map` method with a single type is what a user needs to register mapping. `Map` method with a list of types is only an utility method to avoid to repeat the same operation over and over, but we can live also without it.
That it's why I prefer to move this kind of methods outside the class.

src/Transport/Topology/IConventions.cs
melkio: Ops, sorry. Forgot about it

src/Transport/Topology/PublishersConfiguration.cs
yvesgoeleven: the class is internal, so what does it matter what the constructor is like?

src/Transport/Topology/PublishersConfiguration.cs
yvesgoeleven: both classes are internal? Why are we so concerned about it's public shape? It has none

src/AcceptanceTests/ConfigureEndpointAzureServiceBusTransport.cs
SeanFeldman: Scrape that, forgot that ATTs will reset it every time :(

melkio:
(Maybe) now It's finished

/ @Particular/azure-maintainers

SeanFeldman:
`When_subscribing_to_base_and_derived_polymorphic_events_with_forwarding_topology.Should_handle_each_event_once_only` is failing with "No publishers configured for `NServiceBus.AcceptanceTests.Routing.When_subscribing_to_base_and_derived_polymorphic_events_with_forwarding_topology+BaseEvent`"

yvesgoeleven:
That test seems to be flaky, sometimes it fails, sometimes it doesn't, don't think it's related to this change

Issue #157, opened 2016-03-23
Specify default serializer for the the transport


SeanFeldman:
Initially, core v6 was setting default serializer to `XML` and there was no way to override the default. 
With the change in [this merged PR](https://github.com/Particular/NServiceBus/pull/3617), transport should be able to specify the default serializer, which historically was `JSON`.

Without this change, upgraded azure endpoints would switch from `JSON` to `XML`.

PR #158, opened 2016-03-23
Set transport default serializer

SeanFeldman:
ASB sets default serialization to JSON.

Connects to #157


SeanFeldman:

Issue #156, opened 2016-03-21
Using namespace name instead of connectionString and sanitization don't work well together


melkio:
To reproduce the issue with queue name and namespace name you need 2 endpoints, v6 and v7 (nsb.asb version).
When v6 is sending to v7, the "replyTo" on brokered message looks like "queue@connectionstring". It's getting replaces with "queue@default" when get to v7 endpoint and sanitized, which ends up "queuedefault".

SeanFeldman:
Adding more info:

Have found this when working on Transport Compatibility tests (PR is [here](https://github.com/Particular/ProductionTests/tree/azure-service-bus)). Failing test is `It_is_possible_to_send_request_and_receive_replay(6, 7)` when v6 sends command to v6 and v6 replies back to v7.

When traced through what was happening (in general description), noticed the following

| **Call**  | **Queue value**  | **Note** |
|:-----------|:------------------------|-------------|
| ReceiveNotifier | source@connectionString | |
| BrokeredMessageToIncomingMessageConverte | source@connectionString | uses `ReplyTo` from BrokeredMessage |
| DefaultConnectionStringToNamespaceNameMapper | source@default | |
| Handler(message) | | context.Reply(replyMessage) is called here |
| Dispatcher | | replyMessage getting sent out |
| EndpointOrientedTopologySectionManager.DetermineSendDestination() | | **sourcedefault** is returned as a result on internal call to Sanitization strategy which removed the `@` sign |


Connects to #93

melkio:
Closed by #159

PR #159, opened 2016-03-23
Sanitize "composed" entityPath

melkio:
Connects to #156 

Sanitization strategies didn't work with composed entity path like `queueName@connectionString` or `queueName@namespaceName`.
New implementation splits `queueName` and `suffix`, then it sanitizes only `queueName`

@Particular/azure-maintainers ready to review


SeanFeldman:
@melkio since we don't have a way to validate if this fix is working the way it's expected w/o touching transport compatibility tests, do you want to try locally to verify that it does work?

melkio:
@SeanFeldman before creating the PR I tried. It works

SeanFeldman:
@yvesgoeleven I'm :+1: with the change. Merge?

yvesgoeleven:
Merged!

PR #160, opened 2016-03-24
Remove _if_ check over topology type during configuration

melkio:
Improved design changing _if_ check over type with a more OOP approach 

@Particular/azure-maintainers


yvesgoeleven:
+1

Issue #80, opened 2016-01-06
Callbacks with ASB


SeanFeldman:
Taskforce: @seanfeldman, @melkio 

## General Description

With NSB v6 supporting callbacks in a unified way (using NSB.Callbacks nuget package), it is decided to release NSB.ASB with callbacks support that used the unified implementation and not the native one. 

## Plan of Attack
- [x] verify unified callbacks work
- [x] update transport compatibility tests
- [x] smoke test using [sample code](http://docs.particular.net/samples/callbacks/)

### Native Option (original issue description)

Option of native support for callbacks using SessionReceiver.
![image](https://cloud.githubusercontent.com/assets/1309622/12155467/72e11334-b482-11e5-8808-7632599fe79c.png)

Spike repo: https://github.com/SeanFeldman/ASB.NativeCallback

abombss:
+1, its a nice solution to deal with replies, I also know MT3 just implemented / spiked sagas using sessions and session state.  No 3rd party data store is required and its transactional.

A couple things to consider.

1) Whats the overhead like for lots of request / reply, every request requires a new session to listen to?  Are there quotas, limits, or resource overhead with having hundreds or even thousands of concurrent sessions open on a single queue within a single endpoint?

2) How do you deal with abandoned sessions?  In the spike, the listener for the reply is transient, if that endpoint goes down the reply will never get consumed... probably not a problem that reply is lost, the bigger issue is the reply's will get stuck / hung in the queue because when the endpoint starts back up it won't know it should clear out old sessions, whatever the solution is, it needs to also account for competing consumers (multiple instances of the endpoint running against the same queue).  You might need to think about storing state for the session and being able to do things like enumerate all the sessions QueueClient.GetMessageSessions() on startup to clean-up dead sessions or something.

Overall its a nice feature, it will support competing consumers well so if you are looking for reply to a web farm (multiple nodes) each node doesn't need its own queue, they can share the same queue and just use session id's to filter for their specific response.  It also may reduce latency in some scenarios since there is an active receive pending for a session as opposed to having to drain all the messages in the front of the queue to get to the actual reply.

Can a queue handle both sessions and non-sessions or would it require separate queue with sessions enabled that can't process messages without a session id?

SeanFeldman:
Good questions @abombss.

[...]

PR #165, opened 2016-03-29
Callbacks verification

SeanFeldman:
Added acceptance test to verify unified callbacks in core work with ASB.

Connects to #80 

@Particular/azure-maintainers that are not me, please review :)

### DIFF
--- a/src/AcceptanceTests/Routing/When_scaling_out_senders_that_uses_callbacks.cs
+++ b/src/AcceptanceTests/Routing/When_scaling_out_senders_that_uses_callbacks.cs
@@ -74,13 +74,13 @@
                     .AddMapping<MyRequest>(typeof(ServerThatRespondsToCallbacks));
             }
 
-            class MyResponseHandler : IHandleMessages<MyReposnse>
+            class MyResponseHandler : IHandleMessages<MyResponse>
             {
                 public ReadOnlySettings Settings { get; set; }
 
                 public Context Context { get; set; }
 
-                public Task Handle(MyReposnse message, IMessageHandlerContext context)
+                public Task Handle(MyResponse message, IMessageHandlerContext context)
                 {
                     if (Settings.Get<string>("Client") != message.Client)
                     {
@@ -103,7 +103,7 @@
             {
                 public async Task Handle(MyRequest message, IMessageHandlerContext context)
                 {
-                    await context.Reply(new MyReposnse()
+                    await context.Reply(new MyResponse()
                     {
                         Client = message.Client
                     });
@@ -116,7 +116,7 @@
             public string Client { get; set; }
         }
 
-        class MyReposnse : IMessage
+        class MyResponse : IMessage
         {
             public string Client { get; set; }
         }
@@ -125,10 +125,7 @@
         {
             int repliesReceived;
 
-            public int RepliesReceived
-            {
-                get { return repliesReceived; }
-            }
+            public int RepliesReceived => repliesReceived;
 
             public void ReplyReceived()
             {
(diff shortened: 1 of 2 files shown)


andreasohlund:
>@Particular/azure-maintainers that are not me, please review :)

In the core group we use labels for each maintainer to make sure that

* For outside pulls we have 2 maintainers making sure the pull gets in swiftly
* For inside (from a maintainer) at least one other maintainer reviewing it

See: https://waffle.io/Particular/NServiceBus

Perhaps that could work for you as well?

src/AcceptanceTests/Routing/When_scaling_out_senders_that_uses_callbacks.cs
```diff
@@ -0,0 +1,139 @@
...
+                            var sendOptions = new SendOptions();
+                            sendOptions.RouteReplyToThisInstance();
+                            await bus.Send(new MyRequest()
+                            {
+                                Client = "B"
+                            }, sendOptions);
+                        }
+                    });
+                })
+                .Done(c => c.RepliesReceived >= numberOfMessagesToSend * 2)
```
melkio: Why `>=` and not `==`?

src/AcceptanceTests/Routing/When_scaling_out_senders_that_uses_callbacks.cs
```diff
@@ -0,0 +1,139 @@
...
+                }
+            }
+        }
+
+        class MyRequest : IMessage
+        {
+            public string Client { get; set; }
+        }
+
+        class MyReposnse : IMessage
```
melkio: Typo here: should be `MyResponse`

src/AcceptanceTests/Routing/When_scaling_out_senders_that_uses_callbacks.cs
SeanFeldman: The code is originally taken from RabbitMQ.
I assume if original message is retried, then you could have more than an exact number of replies.

src/AcceptanceTests/Routing/When_scaling_out_senders_that_uses_callbacks.cs
SeanFeldman: Guilty, copy/past :)

src/AcceptanceTests/Routing/When_scaling_out_senders_that_uses_callbacks.cs
melkio: Wait. If message could fail and test counts also retries (and it does) it could fail because `RepliesReceived` is greater than `numberOfMessagesToSend * 2`

src/AcceptanceTests/Routing/When_scaling_out_senders_that_uses_callbacks.cs
SeanFeldman: @melkio you might want to check Rabbit doesn't have the same issue :)

src/AcceptanceTests/Routing/When_scaling_out_senders_that_uses_callbacks.cs
SeanFeldman: >  it could fail because RepliesReceived is greater than numberOfMessagesToSend * 2

You mean it would falsely pass? Yes, it would. It would be an edge case though.

SeanFeldman:
Have smoke tested callbacks and going to merge the PR. 
In case we want to refine how callbacks are tested, it can be done later.

PR #170, opened 2016-04-07
update to core beta 1

SimonCropp:
Connects to https://github.com/Particular/PlatformDevelopment/issues/718


SeanFeldman:
The build is failing @SimonCropp 
@boblangley should we take out Endurance test out of the transport solution?

boblangley:
Yes since it will be moved/adapted to EndToEnd.

SeanFeldman:
:+1: I'll take if off the solution (to avoid compilation nuances), but keep in the repo for now.

SeanFeldman:
@SimonCropp done

PR #172, opened 2016-04-11
Switching to alpha naming scheme

danielmarbach:
Connects to https://github.com/Particular/PlatformDevelopment/issues/706


SeanFeldman:
@danielmarbach so now all of the ASB builds will be postfixed with -alpha rather than unstanble?

danielmarbach:
yes. Which is going to make them sort lower than beta

PR #191, opened 2016-04-19
Minor tweaks

danielmarbach:
Few cleanups like

* Redundant, unnecessary code removed
* Unnecessary awaits removed
* Batch sending concurrently
* Proper await in OnError with CircuitBreaker

### DIFF
--- a/src/Transport/CircuitBreakers/RepeatedFailuresOverTimeCircuitBreaker.cs
+++ b/src/Transport/CircuitBreakers/RepeatedFailuresOverTimeCircuitBreaker.cs
@@ -40,7 +40,7 @@
                 Logger.WarnFormat("The circuit breaker for {0} is now in the armed state", name);
             }
 
-            return Task.Delay(TimeSpan.FromSeconds(1));
+            return TaskEx.Completed;
         }
 
         public void Dispose()
--- a/src/Transport/Seam/MessagePump.cs
+++ b/src/Transport/Seam/MessagePump.cs
@@ -40,7 +40,7 @@ namespace NServiceBus.AzureServiceBus
                 var tokenSource = new CancellationTokenSource();
                 receiveContext.CancellationToken = tokenSource.Token;
 
-                circuitBreaker?.Success();
+                circuitBreaker.Success();
 
                 var context = new ContextBag();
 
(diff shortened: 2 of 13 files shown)


src/Transport/Seam/MessagePump.cs
```diff
@@ -56,10 +56,10 @@ public Task Init(Func<PushContext, Task> pump, CriticalError criticalError, Push
         // For internal testing purposes.
         internal void OnError(Func<Exception, Task> func)
         {
-            topologyOperator.OnError(exception =>
+            topologyOperator.OnError(async exception =>
```
SeanFeldman: The change failed the test and I've found an issue with CB thanks to that. Great :)

src/Transport/Sending/DefaultOutgoingBatchRouter.cs
```diff
@@ -33,21 +33,24 @@ public DefaultOutgoingBatchRouter(IConvertOutgoingMessagesToBrokeredMessages out
...
+                pendingBatches.Add(RouteBatch(batch, context));
             }
+            return Task.WhenAll(pendingBatches);
         }
 
-        public async Task RouteBatch(Batch batch, ReceiveContext context)
+        public Task RouteBatch(Batch batch, ReceiveContext context)
         {
             var outgoingBatches = batch.Operations;
-            if (!outgoingBatches.Any()) return;
```
SeanFeldman: I'm not entirely on board with this one. We remove the check for optimization. But we continue to execute the rest of the code. How much of an optimization is that anyway? Wouldn't exiting early and _not_ performing any other operations be better?

src/Transport/CircuitBreakers/RepeatedFailuresOverTimeCircuitBreaker.cs
```diff
@@ -40,7 +40,7 @@ public Task Failure(Exception exception)
                 Logger.WarnFormat("The circuit breaker for {0} is now in the armed state", name);
             }
 
-            return Task.Delay(TimeSpan.FromSeconds(1));
+            return TaskEx.Completed;
```
danielmarbach: Why don't you make the CB just sync then?

SeanFeldman: Different PR. I've already made a mistake by fixing it here, which insight I shouldn't have done on this PR :)

Issue #184, opened 2016-04-18
Update CONTRIBUTING.md with steps to setup ATTs


melkio:
Something similar to https://github.com/Particular/NServiceBus.AzureStorageQueues/pull/85

PR #193, opened 2016-04-20
Added section with steps to setup ATTs

melkio:
Connects to #184 

Since repo has another "tests" project (`NServiceBus.AzureServiceBus.Tests`) that needs environment variables setup (`AzureServiceBus.ConnectionString` and `AzureServiceBus.ConnectionString.Fallback`), does make sense to create a section to describe this steps too?

@Particular/azure-service-bus-maintainers take a look here


SeanFeldman:
@melkio a section for `AzureServiceBus.ConnectionString` and `AzureServiceBus.ConnectionString.Fallback` as well?

SeanFeldman:
Among the lines of:

## Unit/Integration tests

To execute tests under `NServiceBus.AzureServiceBus.Tests`, two environment variables need to be set up.
1. `AzureServiceBus.ConnectionString`
1. `AzureServiceBus.ConnectionString.Fallback`

Note that those should **not** point to the same namespace.

SeanFeldman:
Added unit testing related section.

Issue #131, opened 2016-02-29
Add startup check for ForwardingTopology to validate EnablePartitioning setting is off


yvesgoeleven:
ASB does not support multiple rules on partitioned entities, so we cannot combine these 2 features together. As forwarding topology relies on multiple rules, we need to validate that partitioning is off. 

Note: Instead of using built-in partitioning, the topology supports bundling itself to overcome potential  throughput limits

TODO
- [x] Implement

SeanFeldman:
Clarification: we can't have multiple rules on a partitioned topic. It is still OK to have partitioned queues.

abombss:
Why not change the topology so its a sub per message type per endpoint?  A topic can have 200 subs.  Then a single rule will suffice.

SeanFeldman:
@abombss it's 2,000 subscriptions per topic

SeanFeldman:
Subscription is much easier when subscriber only requires an event full name. Also, decouples from the publisher.

abombss:
Maybe I am not fully understanding how the topology works and why there is an issue.  If the subscription only requires 1 event name, why would you need more than 1 filter rule?  Or are you saying each subscriber gets a single subscription with multiple rules, 1 for each message type?

We use a hash(subscriber address)_hash(message type name) as the subscription name then use metadata to manage it.  We can easily see related subscriptions since the hash for each component is stable.  Not human readable, but that can easily be mapped with metadata.

SeanFeldman:
Correct @abombss. With the new topology (`ForwardingTopology`), each subscriber gets it's subscription with multiple rules. Each rule represents an event subscriber is interested in. 

The lack of metadata on a rule is unfortunate, and hopefully it will be addressed soon by the ASB team :)
If we'd follow your suggestion, polymorphic events would be delivered multiple times as you'd have multiple subscriptions that would capture the same message. With a rule per message type we solved the [issue](https://github.com/Particular/NServiceBus.AzureServiceBus/issues/53) that always had with `StandardTopology`.

SeanFeldman:
Latest tests with ASB show that a partitioned topic can have multiple rules. 
Should be address it via

- [ ] Remove `TopicPartitioningCheckForForwardingTopology`
- [ ] Update [topologies doco](https://github.com/Particular/docs.particular.net/pull/1411/files#diff-2ce33a85eddac8e0ae33e10c13ff051cR73)

abombss:
[...]

PR #194, opened 2016-04-21
Remove TopicPartitioningCheckForForwardingTopology

SeanFeldman:
Connect to https://github.com/Particular/PlatformDevelopment/issues/718

ASB service does allow now to create partitioned topics with multiple rules.

Addresses reopened #131


SeanFeldman:

Issue #205, opened 2016-04-28
A different approach to Sanitization


SimonCropp:
So currently we have hash things that are too long with an MD5. and in the new version we are adding a new hashing algorithm Sha1. I think this approach is wrong. Hashing names is a pain for debugging since the names are nonsensical in the management UIs. Also once names are hashed and something is deployed to production it is very difficult to move over to a shorter name that does not require hashing.

So my proposal is for the next major to 

 * remove all the hashing. so delete all code related to sha and md5
 * provide a delegate API that support renaming entities.
 * throw an exception if names are two long. ths exception can point to the delegate API and suggest using a hashing algorithm 
 * in the upgrade guide give them the exact code to use the delegate API to implement md5 for backwards compat

I think this approach give just enough friction for new user moving forward to take the preferred option of actually using sensible shortening of names and to avoid the blunt approach of a hash. 

I think it is important to consider this for the current major since it is a breaking change and  putting it off till the next major will mean we would have two hashing algorithms  to handle deprecating. 

### PoA

* [x] Refactor
* [x] Add a sample for SHA1 (part of [docs PR #1470](https://github.com/Particular/docs.particular.net/pull/1470))

SimonCropp:
@SeanFeldman ping

SeanFeldman:
Have discussed this with @SimonCropp and think it's a good point to be taken into consideration for v7.

yvesgoeleven:
@SimonCropp @SeanFeldman we actually have that API, it's our validation + sanitization strategy... so, just move the hashing implementations to snippets/upgrade guide? and make the throwing one one by default?

SimonCropp:
>  so, just move the hashing implementations to snippets/upgrade guide? and make the throwing one one by default?

yes i think so.

SeanFeldman:
:+1: I'll take hashing out and test to see if the error messages are good to explain what needs to happen.

SeanFeldman:
Hashing is a part of Sanitization (removing illegal characters _and_ hashing names that are too long). If default behaviour would be NOT to hash, I would suggest we don't remove illegal characters as well. Upgrade/regular doco can explain how to use alternative strategies that are in place (for backward compatibility using MD5 or w/o backward compatibility with  SHA1). Thoughts?

SimonCropp:
Yes i would say if someone gives us illegal characters we should either validate and throw, or pass it to the underlying azure api and let the exception bubble.

[...]

PR #207, opened 2016-04-29
Remove default sanitization

SeanFeldman:
Address issue #205 
Connect to https://github.com/Particular/PlatformDevelopment/issues/718

TLDR: ASB transport should by default throw an exception for entity names that are invalid rather than trying to sanitize that. Users can specify another strategy based on their needs. Default strategy will be `ThrowOnFailingSanitization`. For backward compatibility ~~`AdjustmentSanitizationV6`~~ `EndpointOrientedTopologySanitization` will be provided. Anything else will be provided as documentation.


SeanFeldman:
@Particular/azure-service-bus-maintainers  please review

yvesgoeleven:
looks good, will lead to doco changes though, can you handle those as well?

SeanFeldman:
@yvesgoeleven correct, doco and sample for SHA1

SeanFeldman:
@yvesgoeleven non of the changes showed up in your PR because they were reverted.

@SimonCropp what was the reason to revert this PR?

PR #223, opened 2016-05-09
[WIP] Background process to complete messages

melkio:
Connects to #212 

@Particular/azure-service-bus-maintainers please review

Issue #212, opened 2016-05-02
Implement Batch Completion


yvesgoeleven:
Looks like Autocomplete doesn't guarantee completion until the next receive operation, we should implement batch completion ourselves. Reverted the code to individual completion for now

SeanFeldman:
When `AutoComplete` is turned on, the completion is offloaded on a separate async thread, so that parallel OnMessage API callbacks will handle messages much faster because time it takes is only {the processing time}.
With AutoComplete, there's no guarantee that completion is successful, until the next callback is invoked. If next callback is not invoked, completion won't take place.

melkio:
> With AutoComplete, there's no guarantee that completion is successful, until the next callback is invoked. 

I'm not sure this is true: I went through the ASB SDK source code and I think it's not about "waiting for next callback" but about "timing"
Into ASB SDK there is a component (`MessageReceivePump`) that adds `message's lockId` to an internal list and tries to "complete" received messages on different threads, running completion in parallel. 
If the endpoint shutdown before all stored `message's lockId` has been addressed, there is no way to complete remaining messages anymore.

SeanFeldman:
@melkio is completion based on the separate thread based on a time or some other trigger?

melkio:
@SeanFeldman what I understood is that `MessageReceivePump` uses an approach similar to `TPL Data Flow`: it creates a sort of async pipeline. Each received message goes through this pipeline. 
One step of this pipeline is to collect messages to complete. After a well defined parameter (`completionBufferWaitTime`) messages will be completed

SeanFeldman:
Fixed via https://github.com/Particular/NServiceBus.AzureServiceBus/commit/60507f7dd13e4a1a2800baeeed84c63c410b3858

PR #227, opened 2016-05-10
Batched completion

SeanFeldman:
Connects to https://github.com/Particular/PlatformDevelopment/issues/718

~~A spike for~~ Solves #212.
(related PRs: #223) 

Using ~~`ConcurrentQueue<Guid>`~~ `ConcurrentStack<Guid>` to store incoming message lock tokens for a batched completion.
Batched completion is taking place on a separate task. The task is backing off (delay) by 0.5 second if there's nothing to complete. Upon notifier stop, operation is blocked until all lock tokens have been completed.

**Things to consider**

The original implementation was using `await message.SafeCompleteAsync().ConfigureAwait(false);`. `SafeCompleteAsync() was logging exceptions _per message_. This is no longer an option with batched completion. Exception will be logged per batch of lock tokens (ie no message IDs.)

@Particular/azure-service-bus-maintainers please review and chime in

### DIFF
--- a/src/Transport/Receiving/MessageReceiverNotifier.cs
+++ b/src/Transport/Receiving/MessageReceiverNotifier.cs
@@ -11,6 +11,7 @@ namespace NServiceBus.AzureServiceBus
     using Logging;
     using Azure.Transports.WindowsAzureServiceBus;
     using Settings;
+    using Utils;
 
     class MessageReceiverNotifier : INotifyIncomingMessages
     {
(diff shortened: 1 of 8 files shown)


SeanFeldman:
PS: it does break :)

src/Transport/Receiving/MessageReceiverNotifier.cs
```diff
@@ -223,6 +256,10 @@ Task AbandonAsyncOnCancellation(BrokeredMessage message)
                 logger.Error("The receiver failed to stop with in the time allowed (30s)");
             }
 
+            // drain the last messages before signaling to stop batched completion task
+            while (!locksIdsToComplete.IsEmpty) {}
```
yvesgoeleven: This is weird, why are you not calling CompleteBatchAsync here?

src/Transport/Receiving/MessageReceiverNotifier.cs
```diff
@@ -223,6 +256,10 @@ Task AbandonAsyncOnCancellation(BrokeredMessage message)
...
+            lockIdCompletionCts.Cancel();
```
yvesgoeleven: I would swap these lines, and just call CompleteBatchAsync with the remaining lockids in the list instead of spinning this thread in a while loop

SeanFeldman:
@melkio spiked turned into a PR.

`StartBatchedCompletionTask` is a task that is running until `MessageReceiverNotifier.Stop()` is invoked. When `MessageReceiverNotifier` is stopped, infinite loop is exited and the remaining messages are drained.

Issue #195, opened 2016-04-21
Exception for no topology selected is cryptic


SeanFeldman:
**Exception thrown**

System.Collections.Generic.KeyNotFoundException: "The given key (NServiceBus.AzureServiceBus.ITopology) was not present in the dictionary."

Exception should provide information on how to select a topology.

PR #232, opened 2016-05-13
Provide meaningful exception when topology is not configured

SeanFeldman:
Fixes issue #195


SeanFeldman:
ops, wrong destination

PR #237, opened 2016-05-21
Dependencies update (nunit 3 & GitVersionTask)

SeanFeldman:


SeanFeldman:
@SimonCropp could you please have a look if I didn't miss anything? If all good, please merge.

SimonCropp:
seems it broke the build :(

SeanFeldman:
One of the tests is not supposed to be executed. Will fix tomorrow.

SeanFeldman:
The failing test `When_saga_id_changed.Should_throw` seems to highlight an issue.
The exception that is expected is the one thrown by the core. Instead, we receive message lock exception which is happening after message is re-appearing 10 times on the queue after transport is actively aborts it.

It appears we know that receive operation is cancelled rather than exception being thrown (in `MessageReceiveNotifier`). @yvesgoeleven could you please have a look at it on Monday? Thank you.

SeanFeldman:
`When_saga_id_changed.Should_throw` will be failing until the race condition is fixed in [ATT PR 3768](https://github.com/Particular/NServiceBus/pull/3768).

PR #239, opened 2016-05-25
Update to ASB v3.2.2

SeanFeldman:
A few [bug fixes](https://www.nuget.org/packages/WindowsAzure.ServiceBus/3.2.2) related to messaging  in the ASB library.


SeanFeldman:
ASB 3.2.1 was unlisted, so 3.2.2 is replacing it + additional fixes.

1 ATT is expected to fail `When_saga_id_changed.Should_throw`.

Issue #201, opened 2016-04-26
RequiresDuplicateDetection can lead to a message loss in v7


SeanFeldman:
ASB transport allows [native duplicates detection](http://docs.particular.net/nservicebus/azure-service-bus/configuration/configuration#controlling-entities-queues). By default it's set to false, but can be enabled. 

SLR is implemented by taking a message failing FLR and, creating a new, deferred one, and dispatching it to the endpoint itself. The deferred message will be dispatched to the same queue with the [same message ID](https://github.com/Particular/NServiceBus.AzureServiceBus/blob/576fc84ca3883a9d8231f15b5bea61ce7e316011/src/Transport/Sending/DefaultBatchedOperationsToBrokeredMessagesConverter.cs#L33) that was used for the originally failing FLR message. 

When `RequiresDuplicateDetection` on ASB queue is enabled, that will trigger de-duplication and SLRed message will be considered as a duplicate message, causing its removal (message loss).

**TODO**

Confirm this behaviour with 
 - [x] Core doesn't require transport message ID to match the NSB (header) message ID*
 - [x] ASB v7
 - ~~ASB v6~~ separate issue #242

In case this is confirmed, potential solutions are:

1. **[Required solution]** Do not re-use NSB message ID for transport message ID. Always use a random unique value. 
 - Cons: ???
 - Pros: consistent for all messages
1. ~~**[Agreed solution]**~~ Detect SLR retries from NSB core [header "NServiceBus.Retries"](https://github.com/Particular/NServiceBus/blob/68f6d76bfaeb218525c2878e751bbd149b2b0b6f/src/NServiceBus.Core/Headers.cs#L96-L96) and if found, replace transport message ID with a random unique value. 
 - Cons: hacky, we need to inspect core header, which could change
 - Pros: unique to each message dispatched
1. Inspect and store information about queue setting for `RequiresDuplicateDetection`. When dispatching a message, check if dedup was enabled. If dedup is enabled, do not re-use NSB message ID, but assign a random unique value.
 - Cons: queue setting is read once, but could be updated and endpoint would run with stale info
 - Pros: ???

\* transport message ID is solely used when no `NServiceBus.MessageId` header is found on an incoming message from a native sender

**Update**:
There are 2 scenarios when a message can be lost because of the `RequiresDuplicateDetection` enabled:

1. Enabled SLR (transport message Id is re-used by utilizing `NServiceBus.MessageId` header value)
[...]

PR #240, opened 2016-05-31
Generate unique transport message id

SeanFeldman:
Connects to #201 


SeanFeldman:

Issue #230, opened 2016-05-12
IValidationStrategy should return validation result and not just a bool


SeanFeldman:
`IValidationStrategy` performs check for two things:

1. entity path/name length
1. invalid characters

and return a `boolean` result. `ThrowOnFailingSanitization` sanitization strategy propagates that information to the user, but in the way that is hard to understand ([item # 3 in the list](https://github.com/Particular/NServiceBus.AzureServiceBus/issues/228)).

Change `IValidationStrategy` to return a validation result that would indicate if path/name is valid or not and a specific reason in case it is not valid.

SeanFeldman:
Closed via https://github.com/Particular/NServiceBus.AzureServiceBus/commit/89ace2b4334e094acd8abba6e762161a4fa5a8ab

Issue #228, opened 2016-05-12
Different error message for invalid queue/msg names


dvdstelt:
Got this error today

>System.Exception: Invalid Rule name `GatedPublishRunner+Event` that cannot be used with Azure Service Bus. Rule name exceeds maximum allowed length or contains invalid characters. Check for invalid characters, shorten the name, or use `Sanitization().UseStrategy<ISanitizationStrategy>()` configuration extension.

Perhaps we can do something like this

>System.InvalidOperationException: Validation rule `[rulename]` triggered on `GatedPublishRunner+Event` which cannot be used with Azure Service Bus. Name contains invalid characters. Check for invalid characters or use `Sanitization().UseStrategy<ISanitizationStrategy>()` in configuration API.

1. `InvalidOperationException` instead of the root `Exception`
2. I was confused by "Invalid Rule name", @SeanFeldman suggested something like `entity`?
3. Two validations are performed and user now has to figure out what is actually going on.
4. "configuration extension" is mentioned, but I don't think end user should have to know it's using extension methods. Rather mention configuration API.

SeanFeldman:
Thank you @dvdstelt. This is good info.

:+1: 

#2 Rule name is invalid, not something within the name as @dvdstelt was trying to figure out. It would probably help to be specific that the rule entity name is problematic, to avoid confusion between rule name and rule itself.
#3 Spot on, will be changed in RC
#4 ""configuration extension" will be replace with "configuration API" to be also consistent with other exception messages the transport throws.

:-1: 

#1 `InvalidOperationException` instead of  `Exception` brings no value. There was nothing done incorrectly. There's a configuration that is needed. No need to introduce `InvalidOperationException` (or a custom exception as @ramonsmits suggested in slack). Users can't act upon this error in the code and continue execution.

yvesgoeleven:
Closing, duplicate of #230

SeanFeldman:
Closed via https://github.com/Particular/NServiceBus.AzureServiceBus/commit/89ace2b4334e094acd8abba6e762161a4fa5a8ab

PR #247, opened 2016-06-06
IValidationStrategy to distinguish between errors

SeanFeldman:
Connect to #230
Connect to #228 

`IValidationStrategy` to return `ValidationResult` which in turn can provide if an entity path/name is valid or not, and list reasons why it was considered invalid. Previously, reason was not clear (issue #228).


SeanFeldman:

yvesgoeleven:
looks good, merged

Issue #229, opened 2016-05-12
Namespace naming Vs Multiple Namespaces


SeanFeldman:
'Multi namespace' is used for 2 scenario's

* Targetted sending (from an endpoint in a namespace to one in another)
* partitioning (2 endpoints communicating over a set of namespaces)

Combining namespace naming with the multi namespace seems to lead to a few issues in this context:
* endpoints in different namespaces cannot talk to each other anymore because the namespace name 'default' (coming from the connectionstring extension method) has a different mapping for both.
* namespaces registered are considered for partitioning only, so direct sending from an endpoint in namespace1 to an endpoint in namespace2 is no longer possible.

Changes required (still under analysis by @yvesgoeleven):

**Connectionstring to addnamespace mapping:**

- [x] default namespace name should be settable.
- [x] connectionstring should not be required so the addnamespace api can be used for all config.

**AddNamespace API:**

- [x] some kind of usage indicator should be introduced to indicate whether a registered namespaces is for targetted sending only, or can be used in the implicit model (with or without @destination in the address) for send or receive.

**Dispatcher changes (probably in impl of ITopologySectionManager):**

- *Option 1*: When a targetted send (or publish) is requested, it should bypass the partitioning logic so that the message is sent to the targetted namespace.
- *Option 2*:  Change the PartitioningIntent to include info about the targetted namespace, so that partitioning logic can use it in it's logic (to choose it in our impl's, but user could than decide for custom strategies).

**Partitioning Strategy Changes:**

- [x] The validation logic of partitioning strategies should be changed to take into account / ignore namespaces registered for targetted sending.

---
Original issue by @SeanFeldman 

Going through [configuration API for secured connection strings](http://docs.particular.net/nservicebus/azure-service-bus/securing-connection-strings), as a reader I could not find any information how 

`.ConnectionString("...");` from the [transport getting started doco](http://docs.particular.net/nservicebus/azure-service-bus/#setting-the-connection-string) relates to 

```c#
var namespacePartitioning = transport.NamespacePartitioning();
namespacePartitioning.UseStrategy<SingleNamespacePartitioning>();
namespacePartitioning.AddNamespace("namespaceName", "connString");
```

Questions I had:

1. What's the relationship between `.ConnectionString();` and `.AddNamespace()`? If `.AddNamespace()` is used, do I still need to provide `.ConnectionString();`? If `default` namespace derived from a single registered namespace with a name?
[...]

Issue #163, opened 2016-03-25
Configure properly RequiresConnectionString property on transport definition


melkio:
AzureServiceBus transport supports multiple namespaces and has an extension method to configure them.
From a user perspective using `ConnectionString` method from `TransportExtensions<T>` to setup the first namespace and then using `NamespacePartitioning` to add the others is weird.

Configuring `RequiresConnectionString` to `false` will allow users to register namespaces with a single method

SeanFeldman:
`RequiresConnectionString` is a core flag? If so, then yes. :+1:

PR #249, opened 2016-06-07
Issue 229 multinamespace naming

yvesgoeleven:
Connect to #229 

Connect to #163

### DIFF
--- a/src/AcceptanceTests/ConfigureEndpointAzureServiceBusTransport.cs
+++ b/src/AcceptanceTests/ConfigureEndpointAzureServiceBusTransport.cs
@@ -27,7 +27,7 @@ public class ConfigureEndpointAzureServiceBusTransport : IConfigureEndpointTestE
         var transportConfig = config.UseTransport<AzureServiceBusTransport>();
 
         AzureServiceBusTransportConfigContext azureServiceBusTransportConfigContext;
-        if(settings.TryGet("AzureServiceBus.TransportConfigContext", out azureServiceBusTransportConfigContext))
+        if(settings.TryGet("AzureServiceBus.AcceptanceTests.TransportConfigContext", out azureServiceBusTransportConfigContext))
         {
             azureServiceBusTransportConfigContext.Callback?.Invoke(endpointName, transportConfig);
         }
--- a/src/AcceptanceTests/ConfigureEndpointAzureServiceBusTransport.cs
+++ b/src/AcceptanceTests/ConfigureEndpointAzureServiceBusTransport.cs
@@ -24,8 +24,18 @@ public class ConfigureEndpointAzureServiceBusTransport : IConfigureEndpointTestE
         var topology = Environment.GetEnvironmentVariable("AzureServiceBusTransport.Topology", EnvironmentVariableTarget.User);
         topology = topology ?? Environment.GetEnvironmentVariable("AzureServiceBusTransport.Topology");
 
-        var transportConfig = config.UseTransport<AzureServiceBusTransport>().ConnectionString(connectionString);
-        
+        var transportConfig = config.UseTransport<AzureServiceBusTransport>();
+
+        AzureServiceBusTransportConfigContext azureServiceBusTransportConfigContext;
+        if(settings.TryGet("AzureServiceBus.TransportConfigContext", out azureServiceBusTransportConfigContext))
+        {
+            azureServiceBusTransportConfigContext.Callback?.Invoke(endpointName, transportConfig);
+        }
+        else
+        {
+            transportConfig.ConnectionString(connectionString);
+        }
+
         if (topology == "ForwardingTopology")
         {
             transportConfig.UseTopology<ForwardingTopology>();
--- a/src/Tests/API/APIApprovals.ApproveAzureServiceBusTransport.approved.txt
+++ b/src/Tests/API/APIApprovals.ApproveAzureServiceBusTransport.approved.txt
@@ -571,7 +571,6 @@ namespace NServiceBus
     public class AzureServiceBusNamespacePartitioningSettings : NServiceBus.Configuration.AdvanceExtensibility.ExposeSettings
     {
         public AzureServiceBusNamespacePartitioningSettings(NServiceBus.Settings.SettingsHolder settings) { }
-        public NServiceBus.AzureServiceBusNamespacePartitioningSettings AddNamespace(string name, string connectionString, NServiceBus.AzureServiceBus.NamespacePurpose purpose = 1) { }
         public NServiceBus.AzureServiceBusNamespacePartitioningSettings UseStrategy<T>()
             where T : NServiceBus.AzureServiceBus.Addressing.INamespacePartitioningStrategy { }
     }
@@ -663,10 +662,11 @@ namespace NServiceBus
     }
     public class static AzureServiceBusTransportExtensions
     {
+        public static void AddNamespace(this NServiceBus.TransportExtensions<NServiceBus.AzureServiceBusTransport> transportExtensions, string name, string connectionString, NServiceBus.AzureServiceBus.NamespacePurpose purpose = 1) { }
         public static void BrokeredMessageBodyType(this NServiceBus.TransportExtensions<NServiceBus.AzureServiceBusTransport> transportExtensions, NServiceBus.AzureServiceBus.SupportedBrokeredMessageBodyTypes type) { }
         public static NServiceBus.AzureServiceBusCompositionSettings Composition(this NServiceBus.TransportExtensions<NServiceBus.AzureServiceBusTransport> transportExtensions) { }
         public static void ConnectivityMode(this NServiceBus.TransportExtensions<NServiceBus.AzureServiceBusTransport> transportExtensions, Microsoft.ServiceBus.ConnectivityMode connectivityMode) { }
-        public static void DefaultNamespaceName(this NServiceBus.TransportExtensions<NServiceBus.AzureServiceBusTransport> transportExtensions, System.Func<string> func) { }
+        public static void DefaultNamespaceName(this NServiceBus.TransportExtensions<NServiceBus.AzureServiceBusTransport> transportExtensions, string name) { }
         public static NServiceBus.AzureServiceBusIndividualizationSettings Individualization(this NServiceBus.TransportExtensions<NServiceBus.AzureServiceBusTransport> transportExtensions) { }
         public static NServiceBus.AzureServiceBusMessageReceiverSettings MessageReceivers(this NServiceBus.TransportExtensions<NServiceBus.AzureServiceBusTransport> transportExtensions) { }
         public static NServiceBus.AzureServiceBusMessageSenderSettings MessageSenders(this NServiceBus.TransportExtensions<NServiceBus.AzureServiceBusTransport> transportExtensions) { }
--- a/src/Tests/API/APIApprovals.ApproveAzureServiceBusTransport.approved.txt
+++ b/src/Tests/API/APIApprovals.ApproveAzureServiceBusTransport.approved.txt
@@ -371,7 +371,7 @@ namespace NServiceBus.AzureServiceBus
     {
         public NamespaceConfigurations() { }
         public int Count { get; }
-        public void Add(string name, string connectionString, NServiceBus.AzureServiceBus.NamespacePurpose purpose = 1) { }
+        public void Add(string name, string connectionString, NServiceBus.AzureServiceBus.NamespacePurpose purpose) { }
         public string GetConnectionString(string name) { }
         public System.Collections.Generic.IEnumerator<NServiceBus.AzureServiceBus.NamespaceInfo> GetEnumerator() { }
     }
--- a/src/Tests/API/APIApprovals.ApproveAzureServiceBusTransport.approved.txt
+++ b/src/Tests/API/APIApprovals.ApproveAzureServiceBusTransport.approved.txt
@@ -416,7 +416,7 @@ namespace NServiceBus.AzureServiceBus
     }
     public enum NamespacePurpose
     {
-        DestinationOnly = 0,
+        Routing = 0,
         Partitioning = 1,
     }
     public abstract class ReceiveContext
@@ -571,8 +571,14 @@ namespace NServiceBus
     public class AzureServiceBusNamespacePartitioningSettings : NServiceBus.Configuration.AdvanceExtensibility.ExposeSettings
     {
         public AzureServiceBusNamespacePartitioningSettings(NServiceBus.Settings.SettingsHolder settings) { }
+        public void AddNamespace(string name, string connectionString) { }
         public NServiceBus.AzureServiceBusNamespacePartitioningSettings UseStrategy<T>()
             where T : NServiceBus.AzureServiceBus.Addressing.INamespacePartitioningStrategy { }
+    }
+    public class AzureServiceBusNamespaceRoutingSettings : NServiceBus.Configuration.AdvanceExtensibility.ExposeSettings
+    {
+        public AzureServiceBusNamespaceRoutingSettings(NServiceBus.Settings.SettingsHolder settings) { }
+        public void AddNamespace(string name, string connectionString) { }
     }
     public class AzureServiceBusQueueSettings : NServiceBus.Configuration.AdvanceExtensibility.ExposeSettings
     {
@@ -662,8 +668,6 @@ namespace NServiceBus
     }
     public class static AzureServiceBusTransportExtensions
     {
-        public static void AddDestinationNamespace(this NServiceBus.TransportExtensions<NServiceBus.AzureServiceBusTransport> transportExtensions, string name, string connectionString) { }
-        public static void AddPartitioningNamespace(this NServiceBus.TransportExtensions<NServiceBus.AzureServiceBusTransport> transportExtensions, string name, string connectionString) { }
         public static void BrokeredMessageBodyType(this NServiceBus.TransportExtensions<NServiceBus.AzureServiceBusTransport> transportExtensions, NServiceBus.AzureServiceBus.SupportedBrokeredMessageBodyTypes type) { }
         public static NServiceBus.AzureServiceBusCompositionSettings Composition(this NServiceBus.TransportExtensions<NServiceBus.AzureServiceBusTransport> transportExtensions) { }
         public static void ConnectivityMode(this NServiceBus.TransportExtensions<NServiceBus.AzureServiceBusTransport> transportExtensions, Microsoft.ServiceBus.ConnectivityMode connectivityMode) { }
@@ -673,6 +677,7 @@ namespace NServiceBus
         public static NServiceBus.AzureServiceBusMessageSenderSettings MessageSenders(this NServiceBus.TransportExtensions<NServiceBus.AzureServiceBusTransport> transportExtensions) { }
         public static NServiceBus.AzureServiceBusMessagingFactoriesSettings MessagingFactories(this NServiceBus.TransportExtensions<NServiceBus.AzureServiceBusTransport> transportExtensions) { }
         public static NServiceBus.AzureServiceBusNamespacePartitioningSettings NamespacePartitioning(this NServiceBus.TransportExtensions<NServiceBus.AzureServiceBusTransport> transportExtensions) { }
+        public static NServiceBus.AzureServiceBusNamespaceRoutingSettings NamespaceRouting(this NServiceBus.TransportExtensions<NServiceBus.AzureServiceBusTransport> transportExtensions) { }
         public static void NumberOfClientsPerEntity(this NServiceBus.TransportExtensions<NServiceBus.AzureServiceBusTransport> transportExtensions, int number) { }
         public static NServiceBus.AzureServiceBusQueueSettings Queues(this NServiceBus.TransportExtensions<NServiceBus.AzureServiceBusTransport> transportExtensions) { }
         public static NServiceBus.AzureServiceBusSanitizationSettings Sanitization(this NServiceBus.TransportExtensions<NServiceBus.AzureServiceBusTransport> transportExtensions) { }
--- a/src/Tests/API/APIApprovals.ApproveAzureServiceBusTransport.approved.txt
+++ b/src/Tests/API/APIApprovals.ApproveAzureServiceBusTransport.approved.txt
@@ -371,15 +371,16 @@ namespace NServiceBus.AzureServiceBus
     {
         public NamespaceConfigurations() { }
         public int Count { get; }
-        public void Add(string name, string connectionString) { }
+        public void Add(string name, string connectionString, NServiceBus.AzureServiceBus.NamespacePurpose purpose = 1) { }
         public string GetConnectionString(string name) { }
         public System.Collections.Generic.IEnumerator<NServiceBus.AzureServiceBus.NamespaceInfo> GetEnumerator() { }
     }
     public class NamespaceInfo : System.IEquatable<NServiceBus.AzureServiceBus.NamespaceInfo>
     {
-        public NamespaceInfo(string name, string connectionString) { }
+        public NamespaceInfo(string name, string connectionString, NServiceBus.AzureServiceBus.NamespacePurpose purpose = 1) { }
         public NServiceBus.AzureServiceBus.Topology.MetaModel.ConnectionString ConnectionString { get; }
         public string Name { get; }
+        public NServiceBus.AzureServiceBus.NamespacePurpose Purpose { get; }
         public bool Equals(NServiceBus.AzureServiceBus.NamespaceInfo other) { }
         public override bool Equals(object obj) { }
         public override int GetHashCode() { }
@@ -413,6 +414,11 @@ namespace NServiceBus.AzureServiceBus
         Active = 0,
         Passive = 1,
     }
+    public enum NamespacePurpose
+    {
+        DestinationOnly = 0,
+        Partitioning = 1,
+    }
     public abstract class ReceiveContext
     {
         protected ReceiveContext() { }
@@ -429,7 +435,7 @@ namespace NServiceBus.AzureServiceBus
     }
     public class RuntimeNamespaceInfo : System.IEquatable<NServiceBus.AzureServiceBus.RuntimeNamespaceInfo>
     {
-        public RuntimeNamespaceInfo(string name, string connectionString, NServiceBus.AzureServiceBus.NamespaceMode mode = 0) { }
+        public RuntimeNamespaceInfo(string name, string connectionString, NServiceBus.AzureServiceBus.NamespacePurpose purpose = 1, NServiceBus.AzureServiceBus.NamespaceMode mode = 0) { }
         public string ConnectionString { get; }
         public NServiceBus.AzureServiceBus.NamespaceMode Mode { get; }
         public string Name { get; }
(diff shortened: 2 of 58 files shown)


src/Tests/API/APIApprovals.ApproveAzureServiceBusTransport.approved.txt
```diff
@@ -653,6 +652,7 @@ namespace NServiceBus
         public static void BrokeredMessageBodyType(this NServiceBus.TransportExtensions<NServiceBus.AzureServiceBusTransport> transportExtensions, NServiceBus.AzureServiceBus.SupportedBrokeredMessageBodyTypes type) { }
         public static NServiceBus.AzureServiceBusCompositionSettings Composition(this NServiceBus.TransportExtensions<NServiceBus.AzureServiceBusTransport> transportExtensions) { }
         public static void ConnectivityMode(this NServiceBus.TransportExtensions<NServiceBus.AzureServiceBusTransport> transportExtensions, Microsoft.ServiceBus.ConnectivityMode connectivityMode) { }
+        public static void DefaultNamespaceName(this NServiceBus.TransportExtensions<NServiceBus.AzureServiceBusTransport> transportExtensions, System.Func<string> func) { }
```
melkio: Why a `Func<string>` and not a simple `string`?

yvesgoeleven: good question, the func is not needed indeed

src/Transport/Config/ExtensionPoints/AzureServiceBusNamespacePartitioningSettings.cs
```diff
@@ -29,7 +29,7 @@ public AzureServiceBusNamespacePartitioningSettings(SettingsHolder settings)
             return this;
         }
 
-        public AzureServiceBusNamespacePartitioningSettings AddNamespace(string name, string connectionString)
+        public AzureServiceBusNamespacePartitioningSettings AddNamespace(string name, string connectionString, NamespacePurpose purpose = NamespacePurpose.Partitioning)
```
melkio: The config API is a bit weird here, IMO: we are into "partitioning" config section but, as a user, I can register a `namepsace` for `DestinationOnly` purpose.
On the other side, to register a namespace for `DestinationOnly` purpose, I have to go through "partitioning" configuration to find the API

src/Transport/Sending/DefaultBatchedOperationsToBrokeredMessagesConverter.cs
```diff
@@ -66,7 +66,27 @@ void SetReplyToAddress(OutgoingMessage outgoingMessage, BrokeredMessage brokered
...
+                        selected = namespaces.FirstOrDefault(ns => ns.Purpose == NamespacePurpose.Partitioning);
+                    }
+
+                    if (selected != null)
+                    {
+                        replyTo = mapper.Map(new EntityAddress(replyTo.Name, selected.Name));
+                    }
+                }
+
+                outgoingMessage.Headers[Headers.ReplyToAddress] = replyTo;
```
melkio: Why does transport change `outgoingMessage`? Isn't the purpose of this class to create and to setup the `brokered message`?

src/Transport/Topology/MetaModel/DefaultNamespaceNameToConnectionStringMapper.cs
```diff
@@ -15,7 +15,7 @@ public DefaultNamespaceNameToConnectionStringMapper(ReadOnlySettings settings)
 
         public EntityAddress Map(EntityAddress value)
         {
-            if (!value.HasSuffix) return value;
+            if (!value.HasSuffix || value.HasConnectionString) return value;
```
melkio: It isn't clear to me how `value` can have a connection string without having a suffix. Am I missing something?

yvesgoeleven: suffix can be either be a namespace name, or a true connectionstring (depends on what user supplies in sendoption.SetDestination()) if it already is a connectionstring it should not try to map it

src/Transport/Sending/DefaultBatchedOperationsToBrokeredMessagesConverter.cs
yvesgoeleven: yes it is, but the headers are also copied over into brokered message properties later

src/Transport/Config/ExtensionPoints/AzureServiceBusNamespacePartitioningSettings.cs
yvesgoeleven: where would you put it?

src/AcceptanceTests/ConfigureEndpointAzureServiceBusTransport.cs
```diff
@@ -21,8 +21,18 @@ public Task Configure(string endpointName, EndpointConfiguration config, RunSett
         var topology = Environment.GetEnvironmentVariable("AzureServiceBusTransport.Topology", EnvironmentVariableTarget.User);
         topology = topology ?? Environment.GetEnvironmentVariable("AzureServiceBusTransport.Topology");
 
-        var transportConfig = config.UseTransport<AzureServiceBusTransport>().ConnectionString(connectionString);
-        
+        var transportConfig = config.UseTransport<AzureServiceBusTransport>();
+
+        AzureServiceBusTransportConfigContext azureServiceBusTransportConfigContext;
+        if(settings.TryGet("AzureServiceBus.TransportConfigContext", out azureServiceBusTransportConfigContext))
```
SeanFeldman: Looks like `AzureServiceBus.TransportConfigContext` is a test specific setting. I'd rather to see a name of a setting key that reflects that. Something like `AzureServiceBus.AcceptanceTest.TransportConfigContext`?

src/Tests/API/APIApprovals.ApproveAzureServiceBusTransport.approved.txt
SeanFeldman: From the user perspective, I can see why `Func<string>` is better - less to write.

```c#
ext.DefaultNamespaceName(() => Environment.GetVariable("default.ns.name"));
```

vs

```c#
var defaultNsName = Environment.GetVariable("default.ns.name")
ext.DefaultNamespaceName(defaultNsName);
```

SimonCropp: u mean 

    ext.DefaultNamespaceName(() => Environment.GetVariable("default.ns.name"));

vs

    ext.DefaultNamespaceName(Environment.GetVariable("default.ns.name"));

SeanFeldman: Time to wrap up work for today. Scrap my comment :smile:

src/Transport/Addressing/Partitioning/Strategies/RoundRobinNamespacePartitioning.cs
```diff
@@ -13,7 +13,14 @@ public class RoundRobinNamespacePartitioning : INamespacePartitioningStrategy
...
+            if (!settings.TryGet(WellKnownConfigurationKeys.Topology.Addressing.Partitioning.Namespaces, out namespaces))
+            {
+                throw new ConfigurationErrorsException("The 'RoundRobin' namespace partitioning strategy requires more than one namespace, please configure multiple azure servicebus namespaces");
+            }
+
+            namespaces = new NamespaceConfigurations(namespaces.Where(n => n.Purpose == NamespacePurpose.Partitioning).ToList());
+
+            if (namespaces.Count <= 1)
             {
                 throw new ConfigurationErrorsException("The 'RoundRobin' namespace partitioning strategy requires more than one namespace, please configure multiple azure servicebus namespaces");
```
SeanFeldman: Namespaces are registered, but not for the right purpose. Should the 2nd exception indicate so to guide the users?

src/Transport/Addressing/Partitioning/Strategies/SingleNamespacePartitioning.cs
```diff
@@ -11,7 +11,14 @@ public class SingleNamespacePartitioning : INamespacePartitioningStrategy
...
+            if (!settings.TryGet(WellKnownConfigurationKeys.Topology.Addressing.Partitioning.Namespaces, out namespaces))
+            {
+                throw new ConfigurationErrorsException("The 'Single' namespace partitioning strategy requires exactly one namespace, please configure the connection string to your azure servicebus namespace");
+            }
+
+            namespaces = new NamespaceConfigurations(namespaces.Where(n => n.Purpose == NamespacePurpose.Partitioning).ToList());
+
+            if (namespaces.Count != 1)
             {
                 throw new ConfigurationErrorsException("The 'Single' namespace partitioning strategy requires exactly one namespace, please configure the connection string to your azure servicebus namespace");
```
SeanFeldman: Same as above

src/Transport/Addressing/Partitioning/Strategies/FailOverNamespacePartitioning.cs
```diff
@@ -11,7 +11,14 @@ public class FailOverNamespacePartitioning : INamespacePartitioningStrategy
...
         {
-            if (!settings.TryGet(WellKnownConfigurationKeys.Topology.Addressing.Partitioning.Namespaces, out namespaces) || namespaces.Count != 2)
+            if (!settings.TryGet(WellKnownConfigurationKeys.Topology.Addressing.Partitioning.Namespaces, out namespaces))
+            {
+                throw new ConfigurationErrorsException("The 'FailOver' namespace partitioning strategy requires exactly two namespaces to be configured");
+            }
+
+            namespaces = new NamespaceConfigurations(namespaces.Where(n => n.Purpose == NamespacePurpose.Partitioning).ToList());
+
+            if (namespaces.Count != 2)
```
SeanFeldman: Namespaces are registered, but not for the right purpose. Should the 2nd exception indicate so to guide the users?

SeanFeldman:
Looks good @yvesgoeleven. A few minor comments.

src/Transport/Config/ExtensionPoints/AzureServiceBusNamespacePartitioningSettings.cs
melkio: `NamespaceName` is a useful info only if a user enables `UseNamespaceNameInsteadOfConnectionString`. Why not put the `namespace name` configuration inside that config api? 
Here the same concept I suggested for ASQ: https://github.com/Particular/NServiceBus.AzureStorageQueues/blob/release-7.0.0/src/Transport/Config/Extensions/AzureStorageTransportAddressingExtensions.cs#L14

We could add `purpose` there, and use partitioning config API to setup only partitioning related info

melkio:
Configuring `DefaultNamespaceName` is useful only if users use `ConnectionString` extension method, isn't it? I'd add a check to ensure that everything is configured properly

yvesgoeleven:
@melkio DefaultNamespaceName also has a default value, so it is configured properly by default

src/Transport/Config/ExtensionPoints/AzureServiceBusNamespacePartitioningSettings.cs
yvesgoeleven: I don't think that is a good idea, UseNamespaceNameInsteadOfConnectionString is only temporary, in future releases it will be default. then we'd have to change the API again

yvesgoeleven: I agree though that AddNamespace is not in the right place, it probably belongs in the root of the transportconfig (next to connectionstring)

SeanFeldman:
@yvesgoeleven I can review one more time if ready

yvesgoeleven:
@SeanFeldman yes please do

src/Transport/Config/ExtensionPoints/AzureServiceBusNamespacePartitioningSettings.cs
melkio: > I don't think that is a good idea, UseNamespaceNameInsteadOfConnectionString is only temporary, in future releases it will be default. then we'd have to change the API again

Agree, but we don't know when `securing connection string` will become the default. For sure not in the next version, I think. 
With this context in mind, we have to provide the best configuration API for the current version (IMO). When we'll upgrade to a newer version we'll think about how to change them.

src/Transport/Addressing/Partitioning/Strategies/FailOverNamespacePartitioning.cs
```diff
@@ -11,10 +11,17 @@ public class FailOverNamespacePartitioning : INamespacePartitioningStrategy
 
         public FailOverNamespacePartitioning(ReadOnlySettings settings)
         {
-            if (!settings.TryGet(WellKnownConfigurationKeys.Topology.Addressing.Partitioning.Namespaces, out namespaces) || namespaces.Count != 2)
+            if (!settings.TryGet(WellKnownConfigurationKeys.Topology.Addressing.Partitioning.Namespaces, out namespaces))
             {
                 throw new ConfigurationErrorsException("The 'FailOver' namespace partitioning strategy requires exactly two namespaces to be configured");
```
melkio: Should report the same exception message as below?

src/Transport/Config/ExtensionMethods/AzureServiceBusTransportExtensions.cs
```diff
@@ -104,6 +104,31 @@ public static void UseNamespaceNamesInsteadOfConnectionStrings(this TransportExt
...
+        public static void DefaultNamespaceName(this TransportExtensions<AzureServiceBusTransport> transportExtensions, string name )
+        {
+            var settings = transportExtensions.GetSettings();
+            settings.Set(WellKnownConfigurationKeys.Topology.Addressing.DefaultNamespaceName, name);
+        }
+
+        /// <summary>
+        /// Adds a namespace for a specific purpose.
+        /// </summary>
+        public static void AddNamespace(this TransportExtensions<AzureServiceBusTransport> transportExtensions, string name, string connectionString, NamespacePurpose purpose = NamespacePurpose.Partitioning)
```
melkio: I'd remove the `default` value for `purpose` parameter, to make the choice more explicit.
Or better, I'd create two different methods: one to add namespace for `partitioning` and one to add for `destination only`

melkio:
> DefaultNamespaceName also has a default value, so it is configured properly by default

`DefaultNamespaceName` makes sense only if `UseNamespaceNameInsteadOfConnectionString` feature is enabled. I don't like a default value here. I'd rather a user has to specify the value if they want to enable `securing connection string`

yvesgoeleven:
@melkio I don't understand why you want to make it explicit. 99% of our users have their endpoints in the same namespace, who cares what it's name is? Specifying the default namespace name explicitly is something that will be done by even a subset of the remaining 1%, those that do cross namespace targetted sends, otherwise again, who cares what the name is. It's such an edge case that I don't understand why you would always want to specify it? btw you can, just don't use that connectionstring method, if you want to be explicit about it, use addnamespace instead

src/Transport/Config/ExtensionMethods/AzureServiceBusTransportExtensions.cs
yvesgoeleven: same as below, targetted namespace sends is an edge case, I see no reason why we should make it explicit

src/Transport/Addressing/Partitioning/Strategies/FailOverNamespacePartitioning.cs
yvesgoeleven: I don't think so, user completely forgot to set any here

melkio:
Maybe I'm missing something, but our conversation about "default namespace name" started because a fixed name could bring into some issue. If we provide a default configured value, users can start working with an endpoint without providing a right configuration. Isn't it?

What I'm suggesting about `AddNamespace` method is a simple rule: a method with a boolean value (or an enum with two possible values) as parameter, are simply two methods. Splitting method in two different methods could provide a better description about what each method does.
But I understand your point. I didn't know how many customers will use this configuration.

yvesgoeleven:
@melkio Could is the correct word, but it's rare, only for users that have endpoints in different 'target' namespaces, that use the connectionstring method to configure (instead of addnamespace), and that do request reply. So all we need to provide is a way to make it work if someone runs into the problem and for some reason cannot switch over to the addnamespace api, but definitily not something I expect to see a lot.

About point 2, do I understand correctly that you would do the following?
 * AddNamespace
 * AddDestinationNamespace

melkio:
@yvesgoeleven yes, what I was suggesting is to remove `purpose` parameter from `AddNamespace` method and to expose two different methods to explain what the transport's doing. Something like: `AddNamespaceForPartitioning` and `AddNamespaceForDirectSending`.

But what about leaving `AddNamespace` method on the `Partitioning` configuration API section and providing an `AddNamespaceForDirectSending` (or something similar) to the root level?
In this way the `purpose` should be implicit.

src/Transport/Addressing/Partitioning/Strategies/FailOverNamespacePartitioning.cs
SeanFeldman: The approach we took is to use name of the strategy as-is. Should we do the same?
```c#
throw new ConfigurationErrorsException($"The '{nameof(FailOverNamespacePartitioning)}' namespace partitioning strategy requires exactly two namespaces to be configured");
```

src/Transport/Addressing/Partitioning/Strategies/FailOverNamespacePartitioning.cs
```diff
@@ -11,10 +11,17 @@ public class FailOverNamespacePartitioning : INamespacePartitioningStrategy
...
             }
+
+            namespaces = new NamespaceConfigurations(namespaces.Where(n => n.Purpose == NamespacePurpose.Partitioning).ToList());
+
+            if (namespaces.Count != 2)
+            {
+                throw new ConfigurationErrorsException($"The 'FailOver' namespace partitioning strategy requires exactly two namespaces for the purpose of partitioning, found {namespaces.Count}");
```
SeanFeldman: Same as above in regards to the strategy name

src/Transport/Addressing/Partitioning/Strategies/RoundRobinNamespacePartitioning.cs
```diff
@@ -13,11 +13,18 @@ public class RoundRobinNamespacePartitioning : INamespacePartitioningStrategy
         public RoundRobinNamespacePartitioning(ReadOnlySettings settings)
         {
             NamespaceConfigurations namespaces;
-            if (!settings.TryGet(WellKnownConfigurationKeys.Topology.Addressing.Partitioning.Namespaces, out namespaces) || namespaces.Count <= 1)
+            if (!settings.TryGet(WellKnownConfigurationKeys.Topology.Addressing.Partitioning.Namespaces, out namespaces))
             {
                 throw new ConfigurationErrorsException("The 'RoundRobin' namespace partitioning strategy requires more than one namespace, please configure multiple azure servicebus namespaces");
```
SeanFeldman: Same as above in regards to the strategy name

src/Transport/Addressing/Partitioning/Strategies/RoundRobinNamespacePartitioning.cs
```diff
@@ -13,11 +13,18 @@ public class RoundRobinNamespacePartitioning : INamespacePartitioningStrategy
...
             }
 
+            namespaces = new NamespaceConfigurations(namespaces.Where(n => n.Purpose == NamespacePurpose.Partitioning).ToList());
+
+            if (namespaces.Count <= 1)
+            {
+                throw new ConfigurationErrorsException($"The 'RoundRobin' namespace partitioning strategy requires more than one namespace for the purpose of partitioning, found {namespaces.Count}. Please configure multiple azure servicebus namespaces");
```
SeanFeldman: Same as above in regards to the strategy name

src/Transport/Addressing/Partitioning/Strategies/SingleNamespacePartitioning.cs
```diff
@@ -11,16 +11,23 @@ public class SingleNamespacePartitioning : INamespacePartitioningStrategy
 
         public SingleNamespacePartitioning(ReadOnlySettings settings)
         {
-            if (!settings.TryGet(WellKnownConfigurationKeys.Topology.Addressing.Partitioning.Namespaces, out namespaces) || namespaces.Count != 1)
+            if (!settings.TryGet(WellKnownConfigurationKeys.Topology.Addressing.Partitioning.Namespaces, out namespaces))
```
SeanFeldman: Same as above in regards to the strategy name

src/Transport/Addressing/Partitioning/Strategies/SingleNamespacePartitioning.cs
```diff
@@ -11,16 +11,23 @@ public class SingleNamespacePartitioning : INamespacePartitioningStrategy
...
             {
                 throw new ConfigurationErrorsException("The 'Single' namespace partitioning strategy requires exactly one namespace, please configure the connection string to your azure servicebus namespace");
             }
+
+            namespaces = new NamespaceConfigurations(namespaces.Where(n => n.Purpose == NamespacePurpose.Partitioning).ToList());
+
+            if (namespaces.Count != 1)
+            {
+                throw new ConfigurationErrorsException("The 'Single' namespace partitioning strategy requires exactly one namespace for the purpose of partitioning, found {namespaces.Count}. Please configure the connection string to your azure servicebus namespace");
```
SeanFeldman: Same as above in regards to the strategy name

src/Transport/Config/ExtensionMethods/AzureServiceBusTransportExtensions.cs
SeanFeldman: Agree with @yvesgoeleven

SeanFeldman:
@yvesgoeleven @melkio added a comment (in several places) for consistency with what we do with other error messages. Once that's addressed, :+1: to merge :)

yvesgoeleven:
@melkio wouldn't it be overkill to have 2 entry points in the api?

src/Transport/Addressing/Partitioning/Strategies/FailOverNamespacePartitioning.cs
yvesgoeleven: ok, I see, will change

melkio:
Yes, but they should be two different APIs
From the "usability" point of view, if I'm choosing and configuring `PartitioningStrategy`, I'd rather to have the namespaces configuration in the same place, as we defined in the previous version of our config API.
Configuring `DirectSend` is another scenario and as you said before, only a small percentage of our customers will use it.

I prefer something like this:

``` csharp
var extensions = new TransportExtensions<AzureServiceBusTransport>(...);
extensions.NamespacePartitioning()
    .UseStrategy<AnyStrategy>()
        .AddNamespace("namespace1", "connectionString1")
        .AddNamespace("namespace2", "connectionString2");

extensions.AddDestinationNamespace("directNamespace", "directConnectionString");
```

Two different places to cover two different scenarios

yvesgoeleven:
@melkio I understand you see it as 2 scenario's, which is true, but from a useability this api is bad. People don't read the doco, they just follow intellisense. If the first thing they see is 'AddDestinationNamespace', they'll use that to set their namespace, which won't work. So AddPartitioningNamespace must be in the root as well

yvesgoeleven:
The build check below can be ignored, there is something wrong with TC, it also executes code inspections on the included APP_Packages folder

melkio:
> People don't read the doco, they just follow intellisense.

@yvesgoeleven Agree, but it isn't our problem. What you are saying is: users choose configuration API to use "randomly". What can we do to handle that? Nothing.
`AddDirectSendNamespace` method explains well what it will do. If users don't read what intellisense suggests...

SeanFeldman:
This is how I see it: destination namespaces served a purpose of overcoming limitations of a single namespace (throughput). With partitioning, we have both, HA and ability to overcome limitation of a single namespace (with RoundRobin, FailOver, etc). Therefore it feels that we should _not_ encourage usage of the destination namespaces, unless it's needed. Perhaps a solution would be to move `.AddDestinationNamespace()` into a "routeing category", which it is. Here are two options for example. What do you think?

### Option 1
```c#
var extensions = new TransportExtensions<AzureServiceBusTransport>(...);
extensions.Routing()
    .AddDestinationNamespace("directNamespaceName", "directConnectionString");

extensions.AddNamespace("namespace1name", "connectionString1");
extensions.AddNamespace("namespace2name", "connectionString2");
extensions.ConnectionString("connString"); //`default` as the name 
```

### Option 2
```c#
var extensions = new TransportExtensions<AzureServiceBusTransport>(...);

extensions.NamespacePartitioning()
    .UseStrategy<AnyStrategy>()
   .AddNamespace("namespace1name", "connectionString1");
   .AddNamespace("namespace2name", "connectionString2");

extensions.Routing()
    .AddDestinationNamespace("directNamespaceName", "directConnectionString");

extensions.ConnectionString("connString"); //`default` as the name 
```

melkio:
Great @SeanFeldman: I vote for the second option!
Doesn't it clash with some `Routing` extension from core?

SeanFeldman:
@melkio core is exposing `.UnicastRouting()` right now.

SimonCropp:
`.Routing()` is likely to clash in the future. perhaps `.AsbRouting()`?

SeanFeldman:
Can we extend on top of Routing() from core?

yvesgoeleven:
@SeanFeldman probably we can, that return type must be public to be useable anyway

yvesgoeleven:
We could call it NamespaceRouting as well?

melkio:
I vote for what @yvesgoeleven suggested (`NamespaceRouting`). It's something directly related to ASB without using `Asb` prefix as suggested by @SimonCropp and with this option we avoid any kind of clashing (or noise in general) with core.

src/Transport/Config/ExtensionPoints/AzureServiceBusNamespaceRoutingSettings.cs
```diff
@@ -0,0 +1,33 @@
...
+        /// <summary>
+        /// Adds a namespace for routing.
+        /// </summary>
+        public void AddNamespace(string name, string connectionString)
+        {
+            NamespaceConfigurations namespaces;
+            if (!settings.TryGet(WellKnownConfigurationKeys.Topology.Addressing.Partitioning.Namespaces, out namespaces))
+            {
+                namespaces = new NamespaceConfigurations();
+                settings.Set(WellKnownConfigurationKeys.Topology.Addressing.Partitioning.Namespaces, namespaces);
```
melkio: Nitpicking here: since transport registers namespaces for different purposes (`Partitioning` and `Routing`) with `WellKnownConfigurationKeys.Topology.Addressing.Partitioning.Namespaces` key, should we change the name of the key to be more general?

src/Transport/Addressing/Partitioning/Strategies/FailOverNamespacePartitioning.cs
```diff
@@ -11,9 +11,16 @@ public class FailOverNamespacePartitioning : INamespacePartitioningStrategy
 
         public FailOverNamespacePartitioning(ReadOnlySettings settings)
         {
-            if (!settings.TryGet(WellKnownConfigurationKeys.Topology.Addressing.Partitioning.Namespaces, out namespaces) || namespaces.Count != 2)
+            if (!settings.TryGet(WellKnownConfigurationKeys.Topology.Addressing.Partitioning.Namespaces, out namespaces))
             {
-                throw new ConfigurationErrorsException("The 'FailOver' namespace partitioning strategy requires exactly two namespaces to be configured");
+                throw new ConfigurationErrorsException($"The '{nameof(FailOverNamespacePartitioning)}' strategy requires exactly two namespaces to be configured");
```
melkio: For others exception message transport provides to users the way to fix the issue, reporting snippets or something similar to be more helpful. Should we improve the ex. message here?

yvesgoeleven: I'll add that

src/Transport/Config/ExtensionPoints/AzureServiceBusNamespaceRoutingSettings.cs
yvesgoeleven: I'll change the key

yvesgoeleven:
I've implemented the NamespaceRouting API & incorporated the suggested changes to exception messages. Time for a final review

melkio:
It's fine by me. 
Waiting for build and @SeanFeldman 's review then we can merge it

SeanFeldman:
Merged

Issue #188, opened 2016-04-19
Missing: DiscriminatorBasedIndividualization SetDiscriminatorGenerator


yvesgoeleven:
The DiscriminatorBasedIndividualization (to be used for backward compat as our old discriminator logic differs slightly from the new one in the core), misses config option to set the value for discriminatorGenerator

SeanFeldman:
Fixed via https://github.com/Particular/NServiceBus.AzureServiceBus/commit/3e04860ec7afd5486e6b88eca41a578d87444295

PR #250, opened 2016-06-07
Allow to set discriminator generator for discriminator based individualization

SeanFeldman:
Connect to #188


SeanFeldman:

Issue #252, opened 2016-06-08
Validation and sanitization API friction


melkio:
Sanitization strategies provided by transport (`EndpointOrientedTopologySanitization` and `ThrowOnFailingSanitization`) have a dependency on configured `IValidationStrategy`.

Sanitization strategies use configured validation strategy to evaluate if provided value is valid or not to drive sanitization behavior. This check is useless since `AddressingLogic` [invokes sanitization only if validation fails](https://github.com/Particular/NServiceBus.AzureServiceBus/blob/release-7.0.0/src/Transport/Addressing/AddressingLogic.cs#L20)

### Plan of Attack

- [x] Change API 
- [x] Update documentation  => [PR #1629](https://github.com/Particular/docs.particular.net/pull/1629)
  - [x] API doco + existing RC issue Particular/docs.particular.net#1262
  - [x] Upgrade guide
  - [x] Samples
- End To End tests
  - [x] Transports compatability tests [here](https://github.com/Particular/EndToEnd/blob/master/src/TransportCompatibilityTests/AzureServiceBusV7/EndpointFacade.cs#L32)
  - [x] Performance tests [here](https://github.com/Particular/EndToEnd/blob/master/src/PerformanceTests/Transport.V6.AzureServiceBus.V7/AzureServiceBusProfile.cs)
- [x] Publish new Beta7 to nuget

SeanFeldman:
> This check is useless since AddressingLogic invokes sanitization only if validation fails

Not quite. Addressing logic is:

```c#
public string Apply(string entityname, EntityType entityType)
{
	var path = composition.GetEntityPath(entityname, entityType);
	var validationResult = validationStrategy.IsValid(path, entityType); 
	if (!validationResult.IsValid)
	{
		path = sanitizationStrategy.Sanitize(path, entityType);
	}
	return path;
}
```

For `ThrowOnFailingSanitization` strategy we could drop the dependency on the validation strategy. But if we'd do that, we will be loosing the guidance in the error message in the exception thrown to the users. We _could_ pass in `validationResult` from the `AddressingLogic` rather than depending on a validation strategy.

Assume `validationResult.IsValid` is `false when `EndpointOrientedTopologySanitization` is running`. Here's the catch: the strategy doesn't know _what_ was invalid. Looking at `EndpointOrientedTopologySanitization`:

```c#
public string Sanitize(string entityPathOrName, EntityType entityType)
{
    // ...
	// remove invalid characters
[...]

PR #255, opened 2016-06-08
[SPIKE] Do not perform validation in sanitization code

SeanFeldman:
Connect to #252 

The issue what @melkio brought up in #252 (Check validation during sanitization is useless) should be not about if check is useless since, but about unnecessary duplicated validation.

This spike is to demonstrate how we can remove double validation and remove dependency on validation strategy when creating a sanitization strategy by flowing in validation result..

@Particular/azure-service-bus-maintainers what do you think? Should we proceed this route? 

**UPDATE**:
Alternative spike: #257 (sanitization and validation combined into `ISanitizationStrategy`)

### DIFF
--- a/src/Transport/Addressing/Sanitization/Strategies/EndpointOrientedTopologySanitization.cs
+++ b/src/Transport/Addressing/Sanitization/Strategies/EndpointOrientedTopologySanitization.cs
@@ -4,21 +4,13 @@
 
     public class EndpointOrientedTopologySanitization : ISanitizationStrategy
     {
-        IValidationStrategy validationStrategy;
-
-        public EndpointOrientedTopologySanitization(IValidationStrategy validationStrategy)
-        {
-            this.validationStrategy = validationStrategy;
-        }
-
-        public string Sanitize(string entityPathOrName, EntityType entityType)
+        public string Sanitize(string entityPathOrName, EntityType entityType, ValidationResult validationResult)
         {
             // remove invalid characters
             var rgx = new Regex(@"[^a-zA-Z0-9\-\._]");
             entityPathOrName = rgx.Replace(entityPathOrName, "");
 
-            var validationResult = validationStrategy.IsValid(entityPathOrName, entityType);
-            if (!validationResult.IsValid)
+            if (!validationResult.LengthIsValid)
             {
                 // turn long name into a guid
                 entityPathOrName = MD5DeterministicNameBuilder.Build(entityPathOrName);
--- a/src/Tests/API/APIApprovals.ApproveAzureServiceBusTransport.approved.txt
+++ b/src/Tests/API/APIApprovals.ApproveAzureServiceBusTransport.approved.txt
@@ -13,8 +13,8 @@
     }
     public class EndpointOrientedTopologySanitization : NServiceBus.AzureServiceBus.Addressing.ISanitizationStrategy
     {
-        public EndpointOrientedTopologySanitization(NServiceBus.AzureServiceBus.Addressing.IValidationStrategy validationStrategy) { }
-        public string Sanitize(string entityPathOrName, NServiceBus.AzureServiceBus.EntityType entityType) { }
+        public EndpointOrientedTopologySanitization() { }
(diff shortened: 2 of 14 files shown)


melkio:
I don't like putting explicit properties about errors inside `ValidationResult` (`CharactersAreValid` and `LengthIsValid`). When transport needs to handle a "new" error we have to change `ValidationResult` shape. What about the scenario in which user wants to introduce a custom validation/sanitization step? They'd need another property inside `ValidationResult`

`Sanitization` and `Validation` are very coupled. Are they the same thing perhaps?

SeanFeldman:
> When transport needs to handle a "new" error we have to change ValidationResult shape.

Feel a little like "what if" :)

Validation of an entity is always about 
1. Characters
2. Maximum path/name length

SeanFeldman:
@melkio and myself have discussed the options. The discussion should be captured in the [issue itself](https://github.com/Particular/NServiceBus.AzureServiceBus/issues/252) in case we'll go with another spike and abandon this one.

src/Transport/Addressing/Sanitization/Strategies/EndpointOrientedTopologySanitization.cs
```diff
@@ -4,21 +4,13 @@
...
-        public string Sanitize(string entityPathOrName, EntityType entityType)
+        public string Sanitize(string entityPathOrName, EntityType entityType, ValidationResult validationResult)
         {
             // remove invalid characters
             var rgx = new Regex(@"[^a-zA-Z0-9\-\._]");
             entityPathOrName = rgx.Replace(entityPathOrName, "");
 
-            var validationResult = validationStrategy.IsValid(entityPathOrName, entityType);
-            if (!validationResult.IsValid)
+            if (!validationResult.LengthIsValid)
```
melkio: This logic seems different than in the current release: `validationResult` is the result produced by `ValidationStrategy` and is created when `AddressingLogic` invokes validation. Inside `SanitizationStrategy` we use provided value without executing again validation to evaluate if length is fine after removing invalid chars.
In the current release transport checks again validity after removing invalid chars

Is it true?

src/Transport/Addressing/Sanitization/Strategies/EndpointOrientedTopologySanitization.cs
```diff
@@ -4,21 +4,13 @@
...
-        public EndpointOrientedTopologySanitization(IValidationStrategy validationStrategy)
-        {
-            this.validationStrategy = validationStrategy;
-        }
-
-        public string Sanitize(string entityPathOrName, EntityType entityType)
+        public string Sanitize(string entityPathOrName, EntityType entityType, ValidationResult validationResult)
         {
             // remove invalid characters
             var rgx = new Regex(@"[^a-zA-Z0-9\-\._]");
```
melkio: `EntityNameValidationV6Rules` and `EndpointOrientedTopologySanitization` share the same `regex` to check and sanitize `pathOrName`. This is another symptom that validation and sanitization are more tiled than we think.
Does make sense validate with a regex and sanitize with another? IMO no

SeanFeldman: Not quite the same...
`EntityNameValidationV6Rules`:              "^[0-9A-Za-z_\.\-]+$"
`EndpointOrientedTopologySanitization`: "[^a-zA-Z0-9\-\._]"

`EntityNameValidationV6Rules` uses the regex to validate characters (with requirement of length more than zero, "+" at the end)
Regex in `EndpointOrientedTopologySanitization` represents all the characters that are not legal. Hence the "^" in the beginning of the character group.

src/Transport/Addressing/Sanitization/Strategies/EndpointOrientedTopologySanitization.cs
SeanFeldman: Correct, that's again highlighting the issue [here](https://github.com/Particular/NServiceBus.AzureServiceBus/issues/252#issuecomment-225677505) :(

SeanFeldman:
Closed in favour of PR #265

PR #257, opened 2016-06-10
[SPIKE] Validation and sanitization combined into ISanitizationStrategy

SeanFeldman:
Connect to #252 

An alternative spike to the original spike #255. 

- `IValidationStrategy` is removed
- `ValidationResult` is removed
- `ISanitizationStrategy` is responsible for sanitization along with validation
- Validation settings on topologies removed


SeanFeldman:
@melkio this is a simplistic version of what we've discussed. Separate spike for a fine-grained (separate sanitization for chars and length) sanitization would be required. Concern with that one - order of operations matters. Necessary order: first, the characters, then the length. Length can become valid once invalid characters are sanitized.

SeanFeldman:
Closed in favour of PR #265

PR #260, opened 2016-06-11
[SPIKE] Sanitization strategy per entityt type

SeanFeldman:
Connect to #252 

A second alternative spike to the original spike #255 that takes spike #257 a step further by introducing a strategy per type.

- `IValidationStrategy` is removed
- `ValidationResult` is removed
- Validation settings on topologies removed
- `ISanitizationStrategy` is replaced with an abstract `SanitizationStrategy` class
- Each instance of a `SanitizationStrategy` sanitizes solely a single entity type it registered against (has `CanSanitize` to determine what a give strategy can sanitize)\*
 - Strategies are registered as object instances (NOTE: requires a little more of polishing)

\* Defaults have changed: `ForwardingTopology` is using `ThrowOnFailingSanitizationStrategyFor[entitytype]` and `EndpointOrientedTopology` is using `EndpointOrientedTopologySanitizationFor[entitytype]`.


SeanFeldman:
@melkio this is a spike on top of #257 spike. 
See my comment [here](https://github.com/Particular/NServiceBus.AzureServiceBus/pull/257#issuecomment-225312957) in regards to splitting character and length validations.

SeanFeldman:
Closed in favour of PR #265

Issue #180, opened 2016-04-13
Improve exception message: No publishers configured for `X`


ramonsmits:
> System.InvalidOperationException: No publishers configured for `GatedPublishRunner+Event`

This message could be improved to identify what to do, as the following:

> System.InvalidOperationException: No publishers configured for `GatedPublishRunner+Event`. Please add mappings via `endpointConfiguration.UseTransport<AzureServiceBusTransport>().UseTopology<EndpointOrientedTopology>().RegisterPublisherForType("endpointname", typeof(MyEvent))`

Based on #177

PR #262, opened 2016-06-15
Improve exception message for missing publishers

melkio:
Connects to #180 

The only concern I have related to add suggestions about how to fix issue inside the exception message, is that configuration API could change in the future and we'll have to remember to upgrade the exception message too. Without any help from compiler or tests

@Particular/azure-service-bus-maintainers ping

### DIFF
--- a/src/Tests/Topology/When_configuring_publishers.cs
+++ b/src/Tests/Topology/When_configuring_publishers.cs
@@ -32,9 +32,7 @@
         {
             configuration.Map("publisher", typeof(MyType));
 
-            var exception = Assert.Throws<InvalidOperationException>(() => configuration.GetPublishersFor(typeof(MyType)));
-            StringAssert.Contains($"No publishers configured for `{typeof(MyType).FullName}`", exception.Message);
-            StringAssert.Contains($"Register mappings via `endpointConfiguration.UseTransport<AzureServiceBusTransport>().UseTopology<EndpointOrientedTopology>().RegisterPublisherForType(\"endpointname\", typeof({typeof(MyType).FullName}))`", exception.Message);
+            Assert.Throws<InvalidOperationException>(() => configuration.GetPublishersFor(typeof(MyType)));
         }
 
         [Test]
--- a/src/Tests/Topology/When_configuring_publishers.cs
+++ b/src/Tests/Topology/When_configuring_publishers.cs
@@ -24,7 +24,7 @@
         {
             var exception = Assert.Throws<InvalidOperationException>(() => configuration.GetPublishersFor(typeof(MyType)));
             StringAssert.Contains($"No publishers configured for `{typeof(MyType).FullName}`", exception.Message);
-            StringAssert.Contains($"Please add mappings via `endpointConfiguration.UseTransport<AzureServiceBusTransport>().UseTopology<EndpointOrientedTopology>().RegisterPublisherForType(\"endpointname\", typeof({typeof(MyType).FullName}))`", exception.Message);
+            StringAssert.Contains($"Register mappings via `endpointConfiguration.UseTransport<AzureServiceBusTransport>().UseTopology<EndpointOrientedTopology>().RegisterPublisherForType(\"endpointname\", typeof({typeof(MyType).FullName}))`", exception.Message);
         }
 
         [Test]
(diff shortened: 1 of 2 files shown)


src/Transport/Topology/PublishersConfiguration.cs
```diff
@@ -48,7 +48,8 @@ public IEnumerable<string> GetPublishersFor(Type type)
         {
             if (!HasPublishersFor(type))
             {
-                throw new InvalidOperationException($"No publishers configured for `{type.FullName}`");
+                throw new InvalidOperationException($"No publishers configured for `{type.FullName}`{Environment.NewLine}" +
+                                                    $"Please add mappings via `endpointConfiguration.UseTransport<AzureServiceBusTransport>().UseTopology<EndpointOrientedTopology>().RegisterPublisherForType(\"endpointname\", typeof({type.FullName}))`");
```
SeanFeldman: > Please add mappings via...

"Register mappings via..."

SeanFeldman:
Valid concern. Though, if API is changing, the tests will break, won't they?

SeanFeldman:
Needs rebasing on top of master to get rid of the inspection issues.

melkio:
> Though, if API is changing, the tests will break, won't they?

I don't follow you. How tests can fail if we change configuration API without changing exception message?

src/Tests/Topology/When_configuring_publishers.cs
```diff
@@ -34,7 +34,7 @@ public void Should_not_map_publisher_and_type_if_it_is_not_a_message()
 
             var exception = Assert.Throws<InvalidOperationException>(() => configuration.GetPublishersFor(typeof(MyType)));
             StringAssert.Contains($"No publishers configured for `{typeof(MyType).FullName}`", exception.Message);
-            StringAssert.Contains($"Please add mappings via `endpointConfiguration.UseTransport<AzureServiceBusTransport>().UseTopology<EndpointOrientedTopology>().RegisterPublisherForType(\"endpointname\", typeof({typeof(MyType).FullName}))`", exception.Message);
+            StringAssert.Contains($"Register mappings via `endpointConfiguration.UseTransport<AzureServiceBusTransport>().UseTopology<EndpointOrientedTopology>().RegisterPublisherForType(\"endpointname\", typeof({typeof(MyType).FullName}))`", exception.Message);
```
SeanFeldman: Is that what you want to test here? The name of the test implies that nothing should happen if registered type is not a message. If the intent is to validate that given it was the only type registered, therefore exception is thrown, perhaps just test for exception, not the exception message?

SeanFeldman:
@melkio in regards to the API, couldn't you use the public API in the `Should_throw_if_mapping_does_not_exist` to register a type that is not a message and check for exception? That way, when the public API is changing, the test will have to change and that way we'll know the exception has to be updated.

src/Tests/Topology/When_configuring_publishers.cs
melkio: Ok, you're right. I already test exception message with the test above. I can remove the assert about message from this test.

>The name of the test implies that nothing should happen if registered type is not a message.
Yes, but the only way to verify that a "not message" hasn't been registered is to try to "resolve" it, asserting against an exception. Others ideas?

melkio:
It's not easy. 
What I'm testing is `PublishersConfiguration`, that it's an `internal` class. What we report into the exception message is what a user has to do, using [configuration API](https://github.com/Particular/NServiceBus.AzureServiceBus/blob/develop/src/Transport/Config/ExtensionPoints/AzureServiceBusEndpointOrientedTopologySettingsExtensions.cs#L13), to fix the issue.

Using public API, I couldn't use `PublishersConfiguration` directly to test what I want, but I should use `EndpointOrientedTopologySectionManager` and call `DetermineResourcesToSubscribeTo` because it's the only method that uses `PublishersConfiguration`. 

I should move focus from the "core" component I want to test I prefer to stay with an easier test rather than making things complex and less readable. What do you think?

src/Tests/Topology/When_configuring_publishers.cs
SeanFeldman: I think I get it now. The test is to verify that a non-message type won't get registered, and the way to test it is to generate an exception. Then just removing the message assertion is good.

melkio:
@Particular/azure-service-bus-maintainers any other comment? otherwise we could merge this PR...

PR #265, opened 2016-06-17
Validation + Sanitization + Hasing strategy

SeanFeldman:
Connect to #252 

Spike #4 with only 2 strategies

| Strategy                               | Default | Validates | Sanitizes | Hashes |
|--------------------------------------|-----------|--------------|--------------|------------|
|`ThrowOnFailedValidation`   | Yes      | Yes          | Throws    | Throws  |
|`ValidateAndHashIfNeeded` | No       | Yes          | No-op       | No-Op   |

Default validation rules are for v7 rules
Validation, sanitization, and hasing rules can be customized via configuration API, otherwise no-op is applied


SeanFeldman:
Found one drawback of the API. It's the ability to set validation and sanitization rules with and without specifying the strategy (`.UseStrategy<T>()`). For hashing and sanitization this could be resolved by allowing `.Hash()` and `.[entitytype]Validation()` only after a strategy is specified. But validation doesn't follow the same path.

yvesgoeleven:
@SeanFeldman I have no problem that this api requires you to be explicit about the strategy, you need to know what you're replacing anyway

src/Transport/Config/DefaultConfigurationValues.cs
```diff
@@ -1,13 +1,21 @@
...
     using Settings;
 
     class DefaultConfigurationValues
     {
+        // Entity segments can contain only letters, numbers, periods (.), hyphens (-), and underscores (-), paths can contain slashes (/)
+        static Regex regex = new Regex(@"^[^\/][0-9A-Za-z_\.\-\/]+[^\/]$");
+
+        // Except for subscriptions and rules, these cannot contain slashes (/)
+        static Regex subscriptionAndRuleNameRegex = new Regex(@"^[0-9A-Za-z_\.\-]+$");
+
```
yvesgoeleven: I wonder if these 2 regex's shouldnt be publicly readable for use in config?

src/Transport/Config/ExtensionPoints/AzureServiceBusSanitizationExtensionPoint.cs
```diff
@@ -0,0 +1,25 @@
...
+    public class AzureServiceBusSanitizationExtensionPoint<T> : ExposeSettings where T : ISanitizationStrategy
+    {
+        SettingsHolder settings;
+
+        public AzureServiceBusSanitizationExtensionPoint(SettingsHolder settings) : base(settings)
+        {
+            this.settings = settings;
+        }
+
+        public AzureServiceBusSanitizationExtensionPoint<T> Hash(Func<string, string> hash)
```
SeanFeldman: Hashing, as sanitization rule, are only applicable when using a strategy other than the default (`ThrowOnFailedValidation`). As per discussion, we'll move all rules for validation & sanitization along with hasing rule under `AzureServiceBusSanitizationExtensionPoint<T>`.

src/Transport/Config/DefaultConfigurationValues.cs
SeanFeldman: @yvesgoeleven I thought about it and really can't come up with a reason to do so.
I'd rather have it documented (or provide a sample) then expose in the code and later regret that we cannot retire it if needed. I've [captured](https://github.com/Particular/NServiceBus.AzureServiceBus/issues/252#issuecomment-227333102) what was validate/sanitization for V6 (and validation for V7 we're currently using) in the issue.

SeanFeldman:
@Particular/azure-service-bus-maintainers please review.

yvesgoeleven:
:+1: good for me

PR #274, opened 2016-07-01
Ensuring backward compat after namespace naming introduction

yvesgoeleven:
Found a few backward compatibility issues when not using namespace names

### DIFF
--- a/src/AcceptanceTests/Routing/When_using_single_namespace.cs
+++ b/src/AcceptanceTests/Routing/When_using_single_namespace.cs
@@ -43,34 +43,9 @@ namespace NServiceBus.Azure.Transports.WindowsAzureServiceBus.AcceptanceTests.Ro
 
             Assert.IsTrue(context.RequestReceived, "context.RequestReceived");
             Assert.IsTrue(context.ReplyReceived, "context.ReplyReceived");
-            Assert.IsTrue(context.ReplyToContainsNamespace, "context.ReplyToContainsNamespace");
+            Assert.IsTrue(context.ReplyToContainsNamespaceName, "context.ReplyToContainsNamespaceName");
         }
-
-
-        [Test]
-        public async Task Should_not_append_namespace_name_to_reply_address_when_using_connection_strings()
-        {
-            var runSettings = new RunSettings();
-            runSettings.TestExecutionTimeout = TimeSpan.FromMinutes(1);
-
-            var context = await Scenario.Define<Context>()
-                .WithEndpoint<SourceEndpoint>(b =>
-                {
-                    b.When(async (bus, c) =>
-                    {
-                        var sendOptions = new SendOptions();
-                        sendOptions.SetDestination("usingsinglenamespace.targetendpoint");
-                        await bus.Send(new MyRequest(), sendOptions);
-                    });
-                })
-                .WithEndpoint<TargetEndpoint>()
-                .Done(c => c.ReplyReceived)
-                .Run(runSettings);
-
-            Assert.IsTrue(context.RequestReceived, "context.RequestReceived");
-            Assert.IsTrue(context.ReplyReceived, "context.ReplyReceived");
-            Assert.IsFalse(context.ReplyToContainsNamespace, "context.ReplyToContainsNamespace");
-        }
+      
 
         public class SourceEndpoint : EndpointConfigurationBuilder
         {
@@ -106,7 +81,7 @@ namespace NServiceBus.Azure.Transports.WindowsAzureServiceBus.AcceptanceTests.Ro
                 public async Task Handle(MyRequest message, IMessageHandlerContext context)
                 {
                     Context.RequestReceived = true;
-                    Context.ReplyToContainsNamespace = context.ReplyToAddress.Contains("@");
+                    Context.ReplyToContainsNamespaceName = context.ReplyToAddress.Contains("@default");
                     await context.Reply(new MyResponse());
                 }
             }
@@ -124,7 +99,7 @@ namespace NServiceBus.Azure.Transports.WindowsAzureServiceBus.AcceptanceTests.Ro
         {
             public bool RequestReceived { get; set; }
             public bool ReplyReceived { get; set; }
-            public bool ReplyToContainsNamespace { get; set; }
+            public bool ReplyToContainsNamespaceName { get; set; }
         }
     }
 }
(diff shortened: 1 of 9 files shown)


src/AcceptanceTests/Routing/When_using_single_namespace.cs
```diff
@@ -0,0 +1,105 @@
...
+
+            runSettings.Set("AzureServiceBus.AcceptanceTests.TransportConfigContext", ctx);
+
+            var context = await Scenario.Define<Context>()
+                .WithEndpoint<SourceEndpoint>(b =>
+                {
+                    b.When(async (bus, c) =>
+                    {
+                        var sendOptions = new SendOptions();
+                        sendOptions.SetDestination("usingsinglenamespace.targetendpoint");
```
SeanFeldman: Just a question - is there a way to get the endpoint name w/o hard-coding a string? We're doing this at several places. Not the end of the world, but curious since ATTs are doing so :)

src/AcceptanceTests/Routing/When_using_single_namespace.cs
```diff
@@ -0,0 +1,105 @@
...
+
+            class MyRequestHandler : IHandleMessages<MyRequest>
+            {
+                public Context Context { get; set; }
+
+                public async Task Handle(MyRequest message, IMessageHandlerContext context)
+                {
+                    Context.RequestReceived = true;
+                    Context.ReplyToContainsNamespaceName = context.ReplyToAddress.Contains("@default");
+                    await context.Reply(new MyResponse());
```
SeanFeldman: `await` is not needed, just return a task?

SeanFeldman:
Looking good :dancers:

src/AcceptanceTests/Routing/When_using_single_namespace.cs
yvesgoeleven: not that I know off

src/Transport/Sending/DefaultBatchedOperationsToBrokeredMessagesConverter.cs
```diff
@@ -81,8 +80,16 @@ void SetReplyToAddress(OutgoingMessage outgoingMessage, BrokeredMessage brokered
 
                     if (selected != null)
                     {
-                        replyTo = mapper.Map(new EntityAddress(replyTo.Name, selected.Name));
+                        if (mapper is DefaultNamespaceNameToConnectionStringMapper)
```
melkio: Checking against types is a smell. I don't like it :)

Anyway...the logic here isn't clear for me. I try to summarize what transport has to accomplish:
- it extracts `replyTo` header value
- if address doesn't contain a suffix (`@something`), transport adds ones
- transport maps built value using configured `mapper`

Is it right? If true, could we change the code as following?

``` csharp
if (outgoingMessage.Headers.ContainsKey(Headers.ReplyToAddress))
{
    var replyTo = new EntityAddress(outgoingMessage.Headers[Headers.ReplyToAddress]);
    if (!replyTo.HasSuffix)
    {
        var namespaces = settings.Get<NamespaceConfigurations(WellKnownConfigurationKeys.Topology.Addressing.Namespaces);
        var defaultName = settings.Get<string(WellKnownConfigurationKeys.Topology.Addressing.DefaultNamespaceName);
        var selected = namespaces.FirstOrDefault(ns => ns.Name == defaultName || 
                                                       ns.Purpose == NamespacePurpose.Partitioning);
        if (selected != null)
            replyTo = new EntityAddress(replyTo, selected.Name);
    }

    var convertedValue = mapper.Map(replyTo);
    outgoingMessage.Headers[Headers.ReplyToAddress] = convertedValue;
    brokeredMessage.ReplyTo = convertedValue;
}
```

yvesgoeleven: I didn't like it either, I wanted to check on the boolean that indicated whether UseNamespaceNames is on or not, but that config method registers a type (so I checked on that type instead)

The logic:
if 'UseName' is on -> always append a name
if 'UseName' is off -> always append a connectionstring

melkio: > var convertedValue = mapper.Map(replyTo); does not translate name into connectionstring, it does the inverse and blow up if the name is not found, that is not what is needed for backward compat.

It isn't true. `Mapper` injected into `DefaultBatchedOperationsToBrokeredMessagesConverter` could be: 
- `DefaultNamespaceNameToConnectionStringMapper` if `UseNamespaceNameInsteadOfConnectionString` is off. This mapper converts back `namespace name` to `connection string`
- `PassThroughNamespaceNameToConnectionStringMapper`  if `UseNamespaceNameInsteadOfConnectionString` is enabled and it doesn't do anything on provided value (that should be a namespace name)

> It's also problematic if the suffix is a connectionstring from the beginning (if user does .Send("ep@fullconnectionstring") and not a name), in that case no mapping is registered and it would also blow up.

I didn't think about this scenario, but in that case I suspect that [this line](https://github.com/Particular/NServiceBus.AzureServiceBus/blob/2e577f3e44f96bf5a8db8872b7b7f71ae3fc2a2c/src/Transport/Sending/DefaultBatchedOperationsToBrokeredMessagesConverter.cs#L69) will throw

Anyway...how do we want to behave in this "mixed scenario": user enables `UseNamespaceNameInsteadOfConnectionString` to improve security and then they'll use `.Send("ep@fullConnectionString")` without configuring namespace name?
It's inconsistent. I'd prefer to throw rather than send it out

yvesgoeleven: DefaultNamespaceNameToConnectionStringMapper should allow connectionstring suffxes without trying to map them, because the content of the reply to header must match the rules of the other endpoint, not the sending endpoint's, it does not necessarliy adhere to the local send/receive/partitioning rules

Imagine there are 3 endpoints, 2 belong to the same 'partitioning set' and share the notion of namespace names. A third is some V5 endpoint that needs to be sent to and is replying back, but it is hosted in some other namespace, like ServiceControl, then you need to be able to send to it directly using a connectionstring regardless of the local setting.

PassThroughNamespaceNameToConnectionStringMapper's behavior is not correct either for the reply to header when the replying endpoint is in another namespace, it does not append anything so the replyer will reply to it's own namespace, while it should reply to the namespace with current endpoint's input queue. In this case the connectionstring should be appended if the replier is V5, or if it is V6 with UseNames off.

yvesgoeleven:
regarding bus.SetReplyAddress("someendpoint@name"). if user explicitly sets SetReplyAddress("someendpoint@name") it means that he wants the reply to be redirected to somewhere else and we need to respect the user's specification and not change it.if he specifies nothing behind @ sign, we can assume it's in same partition set and append based on UseNamespace setting

Issue #277, opened 2016-07-08
Implement recoverability seem


yvesgoeleven:
A new part of the transport seem will be introcuded, which allows native recoverability features (FLR counter)

Instead of core deciding when the error behavior should kick in, this task is delegated to the transport But the core still decides how it's done (send to errorQ, SLR logic etc...).

In order to get this behavior  an OnError callback will be introduced that transport needs to invoke when the maximum retry count has been exceeded, and this callback will return a result that indicates whether core handled the message (by sending to errorQ or SLR)

The behavior needs to be implemented differently depending on transaction mode though:

When transaction mode < AtomicSendWithReceive

onError needs to be invoked after an execution occurs in the pipeline, https://github.com/Particular/NServiceBus.AzureServiceBus/blob/release-7.0.0/src/Transport/Receiving/MessageReceiverNotifier.cs#L199, passing in the exception info

* When core says that the exception is handled, the messages should be completed
* When core says the exception is not handled, the message should be abandoned

When transaction mode = AtomicSendWithReceive then 'Completing after an exception' (bullet 1 above) could lead to pending send operations being triggered which should not be triggered (because the message has been sent to error queue or SLR), so in this case

* Should NOT invoke on error callback after processing the pipeline
* instead always Abandon the message and put exception information in message header (truncate if needed) using Abandon overload (add overload to the adapter)
* on next receive, check the retryAttempts and invoke on error instead of invoking the pipeline https://github.com/Particular/NServiceBus.AzureServiceBus/blob/release-7.0.0/src/Transport/Receiving/MessageReceiverNotifier.cs#L182 passing in the necessary exception information

SeanFeldman:
@melkio could you please update the issue with the details from the sync with the TF? Thank you.

melkio:
> In order to get this behavior an OnError callback will be introduced that transport needs to invoke when the maximum retry count has been exceeded, and this callback will return a result that indicates whether core handled the message (by sending to errorQ or SLR)

It's not right. Transport has to invoke _OnError_ callback each time a message fails. Core handles the retry count exceeded scenario.

_OnError_ callback output could be:

- `Handled`: transport has to acknowledge the message
- `RetryRequired`: transport has to abandon the message and retry later

/ @tmasternak @danielmarbach

yvesgoeleven:
@melkio core handlers the retry count exceeded scenario: but we do have control on the number used (last parameter on ErrorContext), that's what we need

[...]

PR #281, opened 2016-07-21
new seam recoverability 

melkio:
Connects to #277 

~~A first second third~~ fourth  attempt to introduce new seam recoverability for ASB transport.

How to configure "local environment" to compile the solution:
- download nuget packages from [TC](https://builds.particular.net/viewLog.html?buildId=176346&buildTypeId=NServiceBusCore_Build&tab=artifacts#!12q45qa)
- configure (from Visual Studio) a new nuget source to point to foder that contain previously downloaded packages

There is a new package named `NServiceBus.TransportTests.Sources` that contains all tests related to new seam recoverability feature. This package has been added to `NServiceBus.AzureServiceBus.AcceptanceTests` project. 

`NServiceBus.TransportTests.Sources` provides a class (`NServiceBusTransportTest`) to setup tests before running them. I had to change the implementation because transport works in a different ways than other transports and it needs some settings configured properly to work well.

**It works the same way as other transports now**

Using a transactionscope around the core pipeline invocation, combined with a suppress scope inside a section of the pipeline, allows to send outgoing messages and complete the brokered message in the same trasnaction without flowing the transaction into the user code. If an exception occurs the transaction is rolledback though, preventing outbound messages to be sent.
Then the error pipeline is invoked and message is abandoned/completed based on the result without using a transaction at all.

**~~Different solution then originally suggested~~**

~~When an exception we occurs we always invoke the error context immediately, but we mark the receive context as 'recovering'. This recovering flag will indicate the sending infrastructure to bypass the send via queue so that SLR & ErrorQ messages are dispatched immediately and won't rollback if we abandon the origin message.~~

~~When the result says to retry, we abandon so retry will happen
When the result says handled and send via is off, we complete (sent messages have gone out anyway)
When the result says handled and send via is on, we abandon and mark the message with a Recovery header. On the next receive any message that has a recovery header is completed without invoking the pipeline.~~

~~There are 3 tests that fail, all with `TransportTransactionMode = SendsAtomicWithReceive`:~~
- ~~`When_on_message_throws.Should_call_on_error` and `When_on_message_throws_after_immediate_retry.Should_persist_message_delivery_count_between_on_error_calls` fail because to handle failure when `TransportTransactionMode = SendsAtomicWithReceive` transport has to handle message twice and the native counter is 2x than expected.~~
- ~~`When_failure_happens_after_send.Should_not_emit_messages`: it seems receive and send aren't atomic as I thought. Transport receives the message and it fails. Then transport abandons the message but sent messages are delivered anyway.~~

~~`Another point to cover is when `onError` callback fails [here](https://github.com/Particular/NServiceBus.AzureServiceBus/blob/new_seam_recoverability/src/Transport/Receiving/MessageReceiverNotifier.cs#L239). Is it a `critical error` or is it a transient error? Should transport trigger the circuit breaker or should transport try later again (with the risk to retry indefinitely)?~~`


melkio:
@Particular/azure-service-bus-maintainers @tmasternak @danielmarbach

src/Tests/Seam/When_dispatching_messages_in_receive_context.cs
```diff
@@ -461,7 +461,7 @@ public class When_dispatching_messages_in_receive_context
         async Task<ITopologySectionManager> SetupEndpointOrientedTopology(TransportPartsContainer container, string enpointname, SettingsHolder settings)
         {
             container.Register(typeof(SettingsHolder), () => settings);
-            settings.SetDefault<EndpointName>(new EndpointName(enpointname));
+            settings.SetDefault("NServiceBus.Routing.EndpointName", enpointname);
```
SeanFeldman: @danielmarbach is `EndpointName` gone? Not a big fan of string keys...

danielmarbach: Ask the routing TF ;)

SeanFeldman:
The one that doesn't exist? (https://github.com/orgs/Particular/teams?utf8=✓&query=routing)
😎

@tmasternak any ideas?

danielmarbach:
It is @bording @timbussmann and @davidboike

SeanFeldman:
Thank you @danielmarbach

PR #285, opened 2016-07-22
[WIP] Stream adieu

danielmarbach:
See https://github.com/Particular/NServiceBus/pull/3956

Does not address all other release branch changes from core


SeanFeldman:
Can't be merged until newer core is merged. `PushContext` is still accepts `Stream` and not `byte[]`.

yvesgoeleven:
This won't work, always doing brokeredMessage.GetBody<byte[]>() will throw when doing native integration, we need to get the body from the brokered message in the configured way (byte[] or stream) and then convert it to a byte[] to satisfy core needs

yvesgoeleven:
Applying this manually to #281 as v6 package restore includes both changes

PR #289, opened 2016-07-26
Shows the usage of transport transaction

danielmarbach:
Shows how transport transaction was meant to be used.

@Particular/azure-service-bus-maintainers

### DIFF
--- a/src/Tests/Seam/When_dispatching_messages_in_receive_context.cs
+++ b/src/Tests/Seam/When_dispatching_messages_in_receive_context.cs
@@ -57,12 +57,15 @@ namespace NServiceBus.Azure.WindowsAzureServiceBus.Tests.Seam
 
             await pump.Init(async context =>
             {
+                // normally the core would do that
+                context.Context.Set(context.TransportTransaction);
+
                 received = true;
 
                 var bytes = Encoding.UTF8.GetBytes("Whatever");
                 var outgoingMessage = new OutgoingMessage("Id-1", new Dictionary<string, string>(), bytes);
 
-                var ctx = context.Context.Get<ReceiveContext>();
+                var ctx = context.TransportTransaction.Get<ReceiveContext>();
                 ctx.OnComplete.Add(() =>
                 {
                     completed.Set();
@@ -141,12 +144,15 @@ namespace NServiceBus.Azure.WindowsAzureServiceBus.Tests.Seam
 
             await pump.Init(async context =>
             {
+                // normally the core would do that
+                context.Context.Set(context.TransportTransaction);
+
                 received = true;
 
                 var bytes = Encoding.UTF8.GetBytes("Whatever");
                 var outgoingMessage = new OutgoingMessage("Id-1", new Dictionary<string, string>(), bytes);
 
-                var ctx = context.Context.Get<ReceiveContext>();
+                var ctx = context.TransportTransaction.Get<ReceiveContext>();
                 ctx.OnComplete.Add(async () =>
                 {
                     await Task.Delay(1); // makes sure the compiler creates the statemachine that handles exception propagation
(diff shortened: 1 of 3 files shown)


yvesgoeleven:
The test When_dispatching_messages_in_receive_context.Should_dispatch_message_in_receive_context appears to be hanging now

danielmarbach:
@yvesgoeleven pushed an update

SeanFeldman:
@danielmarbach is everything committed? `When_dispatching_messages_in_receive_context` hangs locally (on my sandboxed namespace)

src/Transport/Seam/MessagePump.cs
```diff
@@ -42,12 +42,7 @@ public Task Init(Func<PushContext, Task> pump, CriticalError criticalError, Push
 
                 circuitBreaker.Success();
 
-                var context = new ContextBag();
-
-                context.Set(receiveContext);
-
-                //todo, figure out what the TransportTransaction parameter is about
-                return messagePump(new PushContext(incoming.MessageId, incoming.Headers, incoming.BodyStream, new NoTransaction(), tokenSource, context));
+                return messagePump(new PushContext(incoming.MessageId, incoming.Headers, incoming.BodyStream, new ReceiveContextAwareTransaction(receiveContext), tokenSource, new ContextBag()));
```
SeanFeldman: There's a some sort of mixup here... the `ContextBag` is empty, the receive context is sent via transaction. So the [test](https://github.com/Particular/NServiceBus.AzureServiceBus/blob/7032515acfa946a11dbb869564ebcf287162c115/src/Tests/Seam/When_dispatching_messages_in_receive_context.cs#L65) will throw an exception and abandon the message, causing an infinite loop.

src/Tests/Seam/When_dispatching_messages_in_receive_context.cs
```diff
@@ -62,7 +62,7 @@ public class When_dispatching_messages_in_receive_context
                 var bytes = Encoding.UTF8.GetBytes("Whatever");
                 var outgoingMessage = new OutgoingMessage("Id-1", new Dictionary<string, string>(), bytes);
 
-                var ctx = context.Context.Get<ReceiveContext>();
+                var ctx = context.Context.Get<ReceiveContextAwareTransaction>().ReceiveContext;
```
SeanFeldman: The `ReceiveContext` is not longer in the `ContextBag`, but in the `TransportTransaction`.

src/Transport/Seam/Dispatcher.cs
```diff
@@ -20,24 +20,24 @@ public Dispatcher(IRouteOutgoingBatches routeOutgoingBatches, IBatcher batcher)
 
         public Task Dispatch(TransportOperations operations, ContextBag context)
         {
-            ReceiveContext receiveContext;
+            ReceiveContextAwareTransaction transaction;
             var outgoingBatches = batcher.ToBatches(operations);
 
-            context.TryGet(out receiveContext);
-            if (receiveContext == null) // not in a receive context, so send out immediately
+            context.TryGet(out transaction);
```
SeanFeldman: This is wrong. `ContextBag` is no longer set with the `ReceiveContext`.

SeanFeldman:
@danielmarbach `IDispatchMessages.Dispatch() receives `TransportOperations` and `ContextBag`. The `ReceiveContext` is no longer set in the `ContextBag, but instead is set on the Transport operation. My quick fix is to set `ReceiveContext` on both, but that smells. Should `IDispatchMessages` receive tranport transaction rather than a context bag? I don't follow what the context bag should do here...

danielmarbach:
@Particular/azure-service-bus-maintainers green and ready to be merged

PR #290, opened 2016-07-26
EntityAddress improvements

danielmarbach:
* Removed the implicit conversation operators since to lead to unnecessary string allocations. In on one of the tests it even created a side effect because a string was compared to `EntityAddress` with `!=` and the type was boxed to a string and then the string equality was used. Call `ToString` when you need it.
* Removed computed properties and only parse it once and cache the results
* Made it a struct. Currently, it is 24 bytes which is slightly larger than the recommended struct size of 16 bytes but since this type is used during runtime this tradeoff is OK (confirmed with @Scooletz )

While doing this I think I found a bug in the mapping code. Will raise another PR.

### DIFF
--- a/src/Tests/Receiving/When_converting_brokered_messages_to_incoming_messages.cs
+++ b/src/Tests/Receiving/When_converting_brokered_messages_to_incoming_messages.cs
@@ -4,11 +4,11 @@ namespace NServiceBus.Azure.WindowsAzureServiceBus.Tests.Receiving
     using System.IO;
     using System.Linq;
     using System.Text;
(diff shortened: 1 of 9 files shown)


SeanFeldman:
needs to be rebased @danielmarbach

danielmarbach:
@SeanFeldman it didn't need a rebase. But I did it anyway. Updated and force pushed

danielmarbach:
Green light

SeanFeldman:
Looks good. @yvesgoeleven we're good to merge it (you wanted to do some work on the mappers).

PR #294, opened 2016-07-27
AsyncAutoResetEvent supports now cancellation

danielmarbach:
Refactored the AsyncAutoResetEvent to support cancellation. All tests that are using that event will now automatically cancel after 60 seconds when the event was not received. Will make sure the tests don't hang indefinitely.

I didn't bother to cleanup more. Arguably those tests could be structured better. The timeout could be controlled in a single place but I think this is outside of the scope of this improvement. 

@Particular/azure-service-bus-maintainers please review and merge


danielmarbach:
@Particular/azure-service-bus-maintainers this is green and ready to merge

PR #296, opened 2016-07-27
[WIP] Batcher

danielmarbach:
Proposal to simplify the dispatching logic.

* Avoid unnecessary closure allocations over the ReceiveContext Completed list
* Aligns with other transports
* Fixes a bug where failing service bus operations of DispatchConsistency.Default would not be passed into recoverability (old seam)

Commented out tests which would need to be rewritten

### DIFF
--- a/src/Transport/Seam/Dispatcher.cs
+++ b/src/Transport/Seam/Dispatcher.cs
@@ -3,14 +3,17 @@ namespace NServiceBus.AzureServiceBus
     using System.Collections.Generic;
     using System.Linq;
     using System.Threading.Tasks;
+    using System.Transactions;
     using Extensibility;
     using Microsoft.ServiceBus.Messaging;
+    using Settings;
     using Transports;
 
     class Dispatcher : IDispatchMessages
     {
-        public Dispatcher(IRouteOutgoingBatches routeOutgoingBatches, IBatcher batcher)
+        public Dispatcher(ReadOnlySettings settings, IRouteOutgoingBatches routeOutgoingBatches, IBatcher batcher)
         {
+            this.settings = settings;
             this.routeOutgoingBatches = routeOutgoingBatches;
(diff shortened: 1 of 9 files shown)


yvesgoeleven:
rebased

src/Transport/Seam/Dispatcher.cs
```diff
@@ -46,11 +49,30 @@ Task DispatchBatches(IList<Batch> outgoingBatches, BrokeredMessageReceiveContext
...
+            return DispatchAccordingToIsolationLevel(outgoingBatches, receiveContext);
+        }
+
+        async Task DispatchAccordingToIsolationLevel(IList<Batch> outgoingBatches, BrokeredMessageReceiveContext receiveContext)
+        {
+            var batchesWithIsolatedDispatchConsistency = outgoingBatches.Where(t => t.RequiredDispatchConsistency == DispatchConsistency.Isolated);
+            var batchesWithDefaultConsistency = outgoingBatches.Where(t => t.RequiredDispatchConsistency == DispatchConsistency.Default);
+
+            await routeOutgoingBatches.RouteBatches(batchesWithIsolatedDispatchConsistency, receiveContext).ConfigureAwait(false);
+            await DispatchWithTransactionScopeIfRequired(batchesWithDefaultConsistency, receiveContext).ConfigureAwait(false);
```
SeanFeldman: If we can get multiple dispatcher invocations for isolated consistency messages, and only one invocation for default (batched) level, why do we call `DispatchWithTransactionScopeIfRequired()` each time? Should we first see if it's needed? Or within the method only invoke a router if collection contains any messages?

yvesgoeleven: I'll add a check

SeanFeldman:
Beyond the comments, looks good (and green :D). :+1:

Issue #292, opened 2016-07-26
When_receiving_a_message_with_unknown_transport_encoding done condition


danielmarbach:
The core ATT framework tries to evaluate the done condition of an AT every one millisecond. The done condition is supposed to only check the context. No heavylifting should be done in there. `When_receiving_a_message_with_unknown_transport_encoding` creates a receive infrastructure and does a "blocking" receive on the deadletter queue. Essentially deadlocking the done condition checks. This test should be refactored to use EndpointNamingConventions to get the endpoint name and potentially spin up another endpoint listening on  receivingamessagewithunknowntransportencoding.receiver/$DeadLetterQueue like we do in https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.AcceptanceTests/Recoverability/When_message_is_moved_to_error_queue.cs#L101

SeanFeldman:
Endpoint name is created using [`Conventions.EndpointNamingConventions`](https://github.com/Particular/NServiceBus/blob/09c94279a068ea0282c5d7bead34badd870d080b/src/NServiceBus.AcceptanceTesting/Customization/Conventions.cs#L6) func.

SeanFeldman:
@danielmarbach the approach with a spy won't work. The spy is leveraging NSB endpoint to receive a message. In this case, message is poisonous and cannot be handled by the transport. The only option I see here is to have some sort of native processing that would need to have access to the `Context`.

SeanFeldman:
fixed via https://github.com/Particular/NServiceBus.AzureServiceBus/commit/e74bdf5cf5cd68e3e843ddad1d69b05373076e12

PR #297, opened 2016-08-01
Do not block in Done block

SeanFeldman:
Connect to https://github.com/Particular/NServiceBus.AzureServiceBus/issues/292

1. Remove hardcoded queue name
1. Receive message outside of `.Done()` in a non-blocking fashion

### DIFF
--- a/src/AcceptanceTests/TransportEncoding/When_receiving_a_message_with_unknown_transport_encoding.cs
+++ b/src/AcceptanceTests/TransportEncoding/When_receiving_a_message_with_unknown_transport_encoding.cs
@@ -16,7 +16,7 @@
     public class When_receiving_a_message_with_unknown_transport_encoding : NServiceBusAcceptanceTest
     {
         [Test]
-        public async Task Should_receive_the_message()
+        public async Task Should_deadletter_message()
         {
             var context = await Scenario.Define<Context>()
                     .WithEndpoint<Sender>(b => b.When((bus, ctx) =>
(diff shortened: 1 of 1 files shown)


SeanFeldman:
@danielmarbach I couldn't apply message spy (reasons are [here](https://github.com/Particular/NServiceBus.AzureServiceBus/issues/292#issuecomment-236706454)). Instead, I'm still doing a native receive in a non-blocking fashion outside of the `.Done()` block.

Is there a better way in your opinion? Thanks.

src/AcceptanceTests/TransportEncoding/When_receiving_a_message_with_unknown_transport_encoding.cs
```diff
@@ -26,21 +27,25 @@ public class When_receiving_a_message_with_unknown_transport_encoding : NService
...
-                        var receivedMessageIdMatchesTheOriginal = message.Properties["NServiceBus.MessageId"].ToString() == ctx.OriginalMessageId;
-                        var testRunIdMatchesTheCurrectTestRun = message.Properties["$AcceptanceTesting.TestRunId"] as string == ctx.TestRunId.ToString();
-                        var deliveredOnceOnly = message.DeliveryCount == 1;
-                        ctx.MessageWasMovedToDlq = receivedMessageIdMatchesTheOriginal && testRunIdMatchesTheCurrectTestRun && deliveredOnceOnly;
-
-                        return ctx.MessageWasMovedToDlq;
-                    })
-                    .Run(new RunSettings() {TestExecutionTimeout = TimeSpan.FromMinutes(2)});
+                        // can't apply a message spy pattern here since message will be always poisonous
+                        return Task.Run(async () =>
```
danielmarbach: Ok see why you do it here because the context is created by `Define<TContext>` and this means you can't get access before the test starts.

src/AcceptanceTests/TransportEncoding/When_receiving_a_message_with_unknown_transport_encoding.cs
```diff
@@ -26,21 +27,25 @@ public class When_receiving_a_message_with_unknown_transport_encoding : NService
...
+                        {
+                            var connectionString = Environment.GetEnvironmentVariable("AzureServiceBusTransport.ConnectionString");
+                            var namespaceManager = new NamespaceManagerAdapter(NamespaceManager.CreateFromConnectionString(connectionString));
+                            var factory = MessagingFactory.CreateAsync(namespaceManager.Address, namespaceManager.Settings.TokenProvider).GetAwaiter().GetResult();
```
danielmarbach: `await`

src/AcceptanceTests/TransportEncoding/When_receiving_a_message_with_unknown_transport_encoding.cs
```diff
@@ -26,21 +27,25 @@ public class When_receiving_a_message_with_unknown_transport_encoding : NService
...
+                            var dlqPath = Conventions.EndpointNamingConvention(typeof(Receiver)) + "/$DeadLetterQueue";
+                            var receiver = await factory.CreateMessageReceiverAsync(dlqPath, ReceiveMode.ReceiveAndDelete); 
+                            var message = await receiver.ReceiveAsync();
```
danielmarbach: Wouldn't this need a timeout?

src/AcceptanceTests/TransportEncoding/When_receiving_a_message_with_unknown_transport_encoding.cs
danielmarbach: While this is certainly a solution I would have put it into a feature startup task or something, but here is fair enough and pragmatic.

danielmarbach: @Particular/nservicebus-maintainers Could we provide a better way for this? Maybe allow Define<TContext>(TContext context)` again? Then the ATT could create the context and start the receive task before the scenario starts.

timbussmann: sorry I can't follow this code at a quick glance without digging deeper. what's the problem here? What do you want to solve?

danielmarbach: It is basically an ErrorSpy on steroids. For some messages they need to go raw receive. Spinning up a raw receiver and still have access to the context. The hack @SeanFeldman did now is starting that raw receive in the When. But I'd say it would be cleaner to do something like

```
var context = new ScenarioContext();
var rawReceive = Task.Run(async => { // rawReceiveWithAccessToContext ));
var scenarioTask = Scenario.Define(context)...
await Task.WhenAll();
```

SeanFeldman:
@danielmarbach updated.
@Particular/azure-service-bus-maintainers please merged if gets :+1: from @danielmarbach

src/AcceptanceTests/TransportEncoding/When_receiving_a_message_with_unknown_transport_encoding.cs
```diff
@@ -27,20 +30,25 @@ public class When_receiving_a_message_with_unknown_transport_encoding : NService
...
+
+            // can't apply a message spy pattern here since message will be always poisonous
+            var rawReceiveTask = Task.Run(async () =>
+            {
+                var connectionString = Environment.GetEnvironmentVariable("AzureServiceBusTransport.ConnectionString");
+                var namespaceManager = new NamespaceManagerAdapter(NamespaceManager.CreateFromConnectionString(connectionString));
+                var factory = MessagingFactory.CreateAsync(namespaceManager.Address, namespaceManager.Settings.TokenProvider).GetAwaiter().GetResult();
+                var dlqPath = Conventions.EndpointNamingConvention(typeof(Receiver)) + "/$DeadLetterQueue";
+                var receiver = await factory.CreateMessageReceiverAsync(dlqPath, ReceiveMode.ReceiveAndDelete);
+                var message = await receiver.ReceiveAsync();
```
danielmarbach: I still think we should use a Receive with timeout

danielmarbach:
Cheecky trick @seanfeldmann with setting a context which is then overwritten

SeanFeldman:
:) is there a better way in this context?
I'm trying to re-implement this test with a `FeatureStartupTask` as you've suggested, but once a startup task is executing, the send on the Sender endpoint is never executed.

src/AcceptanceTests/TransportEncoding/When_receiving_a_message_with_unknown_transport_encoding.cs
SeanFeldman: I could constrain it to the time the test is running. But is it necessary? Wouldn't the task get terminated with test is done?

SeanFeldman: Probably would be better to do
```c#
Context context;
var scenarioTask = Scenario.Define<Context>(contextInitializer => context = contextInitializer)...
```

SeanFeldman:
`FeatureStartupTask` won't work due to how startup tasks are executed. This is a final version
@Particular/azure-service-bus-maintainers please review.

Issue #309, opened 2016-08-24
Reply-To header is not set correctly when using namespace aliases


SeanFeldman:
## Description

When the transport is instructed to replace namespace connections strings, `reply-to` header should be populated with alias and not the connection string. [`DefaultBatchedOperationsToBrokeredMessagesConverter`](https://github.com/Particular/NServiceBus.AzureServiceBus/blob/release-7.0.0/src/Transport/Sending/DefaultBatchedOperationsToBrokeredMessagesConverter.cs#L81) does the opposite. 

```c#
if (useAliases)
{
   replyTo = new EntityAddress(replyTo.Name, selected.ConnectionString);
}
else
{
   replyTo = new EntityAddress(replyTo.Name, selected.Alias);
}
```

Assignments in If/else should be inverted.

## PoA
- [x] Fix the code
- [x] Fix the [test `When_converting_outgoing_messages_to_brokered_messages`](https://github.com/Particular/NServiceBus.AzureServiceBus/blob/212d068b2ac8ab11f26c128b6e68e6f8697eee10/src/Tests/Sending/When_converting_outgoing_messages_to_brokered_messages.cs#L232) to verify alias/connString is applied properly

PR #311, opened 2016-08-25
Reply-To header is not set correctly when using namespace aliases

SeanFeldman:
Connect to https://github.com/Particular/NServiceBus.AzureServiceBus/issues/309


SeanFeldman:

PR #312, opened 2016-08-25
Lower noise when transient exceptions occur with OnMessage callback

SeanFeldman:
Currently logging non-transient and transient exceptions with INFO threshold.

With the change log
 * Transient exception with DEBUG threshold with full exception
 * Non-transient exception with INFO threshold with message only

### DIFF
--- a/src/Transport/Receiving/MessageReceiverNotifier.cs
+++ b/src/Transport/Receiving/MessageReceiverNotifier.cs
@@ -72,8 +72,6 @@ namespace NServiceBus.Transport.AzureServiceBus
 
         void OptionsOnExceptionReceived(object sender, ExceptionReceivedEventArgs exceptionReceivedEventArgs)
         {
-            // todo respond appropriately
-
             // according to blog posts, this method is invoked on
             //- Exceptions raised during the time that the message pump is waiting for messages to arrive on the service bus
             //- Exceptions raised during the time that your code is processing the BrokeredMessage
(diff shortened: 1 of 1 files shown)


src/Transport/Receiving/MessageReceiverNotifier.cs
```diff
@@ -72,18 +72,26 @@ public void Initialize(EntityInfo entity, Func<IncomingMessageDetails, ReceiveCo
...
-
             // according to blog posts, this method is invoked on
             //- Exceptions raised during the time that the message pump is waiting for messages to arrive on the service bus
             //- Exceptions raised during the time that your code is processing the BrokeredMessage
             //- It is raised when the receive process successfully completes. (Does not seem to be the case)
 
             if (!stopping) //- It is raised when the underlying connection closes because of our close operation
             {
-                logger.InfoFormat("OptionsOnExceptionReceived invoked, action: {0}, exception: {1}", exceptionReceivedEventArgs.Action, exceptionReceivedEventArgs.Exception);
+                var messagingException = (MessagingException)exceptionReceivedEventArgs.Exception;
```
yvesgoeleven: shouldn't we do an 'AS' cast here, what if it's not a MessagingException

SeanFeldman: According to the ASB team it is, but we could do

```c#
var messagingException = exceptionReceivedEventArgs.Exception as MessagingException;
if (messagingException?.IsTransient)
{
   //...
}

yvesgoeleven: yes, that's much better

yvesgoeleven: I'll make the change, go sleep!

SeanFeldman: :sleeping:

yvesgoeleven: The elvis operator did not work, it turns the result int a bool? not a bool

PR #316, opened 2016-08-31
[WIP] Fixed ToTransportAddress by making the translation explicit

SzymonPobiega:
The previous version with `ToString` would include all instance properties (if provided) which is not something we want in the address.

### DIFF
--- a/src/Transport/Config/AzureServiceBusTransportInfrastructure.cs
+++ b/src/Transport/Config/AzureServiceBusTransportInfrastructure.cs
@@ -2,7 +2,7 @@
 {
     using System;
     using System.Collections.Generic;
-    
+    using System.Text;
     using DelayedDelivery;
     using Performance.TimeToBeReceived;
     using Routing;
@@ -27,7 +27,19 @@
 
         public override string ToTransportAddress(LogicalAddress logicalAddress)
         {
-            return logicalAddress.ToString();
+            var queue = new StringBuilder(logicalAddress.QueueName);
+
+            if (logicalAddress.Discriminator != null)
+            {
+                queue.Append("-" + logicalAddress.Discriminator);
+            }
+
+            if (logicalAddress.Qualifier != null)
+            {
+                queue.Append("." + logicalAddress.Qualifier);
+            }
+
+            return queue.ToString();
         }
 
         public override TransportReceiveInfrastructure ConfigureReceiveInfrastructure()


yvesgoeleven:
@SzymonPobiega it doesn't appear to compile on the buildserver due to missing QueueName & Discriminator properties, is there some version dependency mismatch?

SzymonPobiega:
@yvesgoeleven it is WIP because the parent one is not yet merged. I'll create a non-WIP version that does not depend on that change in the core.

src/Transport/Config/AzureServiceBusTransportInfrastructure.cs
```diff
@@ -27,7 +27,19 @@ public override EndpointInstance BindToLocalEndpoint(EndpointInstance instance)
 
         public override string ToTransportAddress(LogicalAddress logicalAddress)
         {
-            return logicalAddress.ToString();
+            var queue = new StringBuilder(logicalAddress.QueueName);
```
SeanFeldman: Why couldn't this be encapsulated in `LogicalAddress.ToString()`?
Or, alternatively, not to have a `LogicalAddress.ToToTransportAddress()`?

SzymonPobiega: Because the actual code depends on the transport. I mean, different transport can use different separators for qualifiers and discriminators (e.g. SQL) and can use additional properties (MSMQ, SQL)

SeanFeldman: Thank you @SzymonPobiega.
So discriminator and qualifier are mutually exclusive? Sorry for pestering with these questions, want to understand :)

SzymonPobiega: Yes, in the current core code qualifier and discriminator are mutually exclusive. There is even a check in `LogicalAddress` which prevents creating an address with both. The idea is that the satellite queues (which have qualifier) are always shared by all instances.

I am not entirely sure if this is a permanent property but it seems unlikely that we'll have a per-instance satellites any time soon.

SeanFeldman: Thank you @SzymonPobiega

SzymonPobiega:
Replaced by #318 which does not require changes in the core.

PR #327, opened 2016-09-13
Core RC2

yvesgoeleven:
- Update ASB to Core RC2
- Fix the issue of message pumps (main input queue pump, that has to be reused between `SubscriptionManager` and `MessagePump`, and satellite input queues pumps)


SeanFeldman:

src/Transport/Config/AzureServiceBusTransport.cs
```diff
@@ -28,7 +29,7 @@ public override TransportInfrastructure Initialize(SettingsHolder settings, stri
 
             SetConnectivityMode(settings);
 
-            return new AzureServiceBusTransportInfrastructure(topology, settings);
+            return new AzureServiceBusTransportInfrastructure(topology, settings, settings.Get<SatelliteTransportAddressCollection>());
```
yvesgoeleven: why inject the collection if the settings holder is already injected?

src/Transport/Config/DefaultConfigurationValues.cs
```diff
@@ -18,9 +19,16 @@ public SettingsHolder Apply(SettingsHolder settings)
             ApplyDefaultValuesForRules(settings);
             ApplyDefaultValuesForSerialization(settings);
             ApplyDefaultValuesForBrokeredMessageConventions(settings);
+            ApplyDefaultValuesForSatelliteTransportAddresses(settings);
             return settings;
         }
 
+        static void ApplyDefaultValuesForSatelliteTransportAddresses(SettingsHolder settings)
+        {
+            // TODO: review if we can improve this
```
yvesgoeleven: you could check on consumption if the settings key is configured

src/Transport/Seam/MessagePump.cs
```diff
@@ -17,28 +19,19 @@ class MessagePump : IPushMessages, IDisposable
         RepeatedFailuresOverTimeCircuitBreaker circuitBreaker;
         ILog logger = LogManager.GetLogger(typeof(MessagePump));
         string inputQueue;
+        SatelliteTransportAddressCollection satelliteTransportAddresses;
 
         public MessagePump(ITopologySectionManager topologySectionManager, ITransportPartsContainer container)
         {
             this.topologySectionManager = topologySectionManager;
             this.container = container;
+            satelliteTransportAddresses = container.Resolve<ReadOnlySettings>().Get<SatelliteTransportAddressCollection>();
```
yvesgoeleven: wouldn't it be better to inject ReadOnlySettings instead of querying the container here?

yvesgoeleven:
@SeanFeldman added a few comments, but besides that looks good

src/Transport/Config/DefaultConfigurationValues.cs
andreasohlund: I think the current way is good enough for now. I'll remove the todo?

Issue #329, opened 2016-09-16
Message processed multiple times when sending from a SendOnly endpoint


carlbm:
When a message is sent from a send only endpoint the handler is invoked multiple times. There is no limit to the number of times the handler is invoked (the first line of the stack trace would suggest otherwise, but it's not the case), and it is invoked again immediately after it returns - I've only tested it with one message however, ignoring the immediate and delayed retry rules. 

Here's the stack trace:

`2016-09-16 11:55:12.849 WARN  NServiceBus.RecoverabilityExecutor Delayed Retry will reschedule message 'fd0fb6cd-b939-4882-9ceb-a68300c1ff8c' after a delay of 00:00:30 because of an exception:
System.ArgumentException: Entity address value can't be empty
Parameter name: value
   at NServiceBus.Transport.AzureServiceBus.EntityAddress..ctor(String value) in C:\Build\src\Transport\Topology\MetaModel\EntityAddress.cs:line 49
   at NServiceBus.Transport.AzureServiceBus.DefaultBatchedOperationsToBrokeredMessagesConverter.SetReplyToAddress(OutgoingMessage outgoingMessage, BrokeredMessage brokeredMessage, RuntimeNamespaceInfo destinationNamespace) in C:\Build\src\Transport\Sending\DefaultBatchedOperationsToBrokeredMessagesConverter.cs:line 64
   at NServiceBus.Transport.AzureServiceBus.DefaultBatchedOperationsToBrokeredMessagesConverter.Convert(BatchedOperation outgoingOperation, RoutingOptions routingOptions) in C:\Build\src\Transport\Sending\DefaultBatchedOperationsToBrokeredMessagesConverter.cs:line 36
   at NServiceBus.Transport.AzureServiceBus.DefaultBatchedOperationsToBrokeredMessagesConverter.<>c__DisplayClass2_0.<Convert>b__0(BatchedOperation x) in C:\Build\src\Transport\Sending\DefaultBatchedOperationsToBrokeredMessagesConverter.cs:line 23
   at System.Linq.Enumerable.WhereSelectListIterator`2.MoveNext()
   at System.Collections.Generic.List`1..ctor(IEnumerable`1 collection)
   at System.Linq.Enumerable.ToList[TSource](IEnumerable`1 source)
[...]

PR #330, opened 2016-09-16
Issue 329

yvesgoeleven:
Fix for issue #329

### DIFF
--- a/src/Transport/Sending/DefaultBatchedOperationsToBrokeredMessagesConverter.cs
+++ b/src/Transport/Sending/DefaultBatchedOperationsToBrokeredMessagesConverter.cs
@@ -64,7 +64,9 @@ namespace NServiceBus.Transport.AzureServiceBus
             if (outgoingMessage.Headers.ContainsKey(Headers.ReplyToAddress))
             {
                 var replyToAddress = outgoingMessage.Headers[Headers.ReplyToAddress];
-                if (string.IsNullOrWhiteSpace(replyToAddress)) return;
+                // Read-only endpoints have no reply-to value
+                if (string.IsNullOrWhiteSpace(replyToAddress))
+                    return;
 
                 var replyTo = new EntityAddress(replyToAddress);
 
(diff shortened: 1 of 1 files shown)


yvesgoeleven:
@Particular/azure-service-bus-maintainers this PR is ready for review

src/Transport/Sending/DefaultBatchedOperationsToBrokeredMessagesConverter.cs
```diff
@@ -63,7 +63,10 @@ void SetReplyToAddress(OutgoingMessage outgoingMessage, BrokeredMessage brokered
         {
             if (outgoingMessage.Headers.ContainsKey(Headers.ReplyToAddress))
             {
-                var replyTo = new EntityAddress(outgoingMessage.Headers[Headers.ReplyToAddress]);
+                var replyToAddress = outgoingMessage.Headers[Headers.ReplyToAddress];
+                if (string.IsNullOrWhiteSpace(replyToAddress)) return;
```
SeanFeldman: I'd leave a comment saying "Read-only endpoint case has no reply-to value".

src/Transport/Sending/DefaultBatchedOperationsToBrokeredMessagesConverter.cs
```diff
@@ -155,7 +158,10 @@ void ApplyDeliveryConstraints(BrokeredMessage brokeredMessage, BatchedOperation
 
         static void CopyHeaders(OutgoingMessage outgoingMessage, BrokeredMessage brokeredMessage)
         {
-            brokeredMessage.Properties[Headers.ReplyToAddress] = brokeredMessage.ReplyTo;
+            if (!string.IsNullOrEmpty(brokeredMessage.ReplyTo))
```
SeanFeldman: So if brokeredMessage.ReplyTo is empty, we don't want even to have a header (empty)?

yvesgoeleven: correct, it serves no purpose

src/Transport/Sending/DefaultBatchedOperationsToBrokeredMessagesConverter.cs
yvesgoeleven: the empty header was actually created by the code on the other comment. But nothing says other causes could exist

Issue #308, opened 2016-08-24
Delivery count to respect Immediate retries and user settings


SeanFeldman:
## Description

Before the new retries seam implementation, retries feature did not respect the native `DeliveryCount`. 
NSB.ASB is setting the default value to [10 and we expect customers to adjust it](http://docs.particular.net/nservicebus/azure-service-bus/configuration/full#controlling-entities-queues) according to what is configured with Immediate and Delayed retries.

With the new seam in place, native delivery counter is used by immediate and delayed retries and we we could configure `DeliveryCount` to match Immediate retries count (Delayed retries are not relevant since  SLRed messages are brand new message).

## Proposal

Leave the default and reconfigure `DeliveryCount` to match Immediate retries (ex-FLR) and allow user override with validation to ensure user setting is no less than Immediate retries count is set to.

## Technical Info

Delivery count is configured for [queues](https://github.com/Particular/NServiceBus.AzureServiceBus/blob/release-7.0.0/src/Transport/Config/DefaultConfigurationValues.cs#L72) and [subscriptions](https://github.com/Particular/NServiceBus.AzureServiceBus/blob/release-7.0.0/src/Transport/Config/DefaultConfigurationValues.cs#L114).

Core exposes the setting via [`NumberOfImmediateRetries`](https://github.com/Particular/NServiceBus/blob/e3a465651022f4ecb2ad56909495bf324dea8e6a/src/NServiceBus.Core/Recoverability/Recoverability.cs#L214) settings key.

## PoA
- [x] Implement changes PR #333
- [x] [Update documentation and upgrade guide](http://docs.particular.net/nservicebus/azure-service-bus/retries#first-level-retries-vs-maxdeliverycount) [PR #1945](https://github.com/Particular/docs.particular.net/pull/1945)

SeanFeldman:
@Particular/azure-service-bus-maintainers should we get this into RC?

SeanFeldman:
@yvesgoeleven ping

yvesgoeleven:
@SeanFeldman did we miss this one? I think we can do it on maintainer budget, has no impact on api

SeanFeldman:
[...]

PR #333, opened 2016-09-20
Delivery count as immediate retries

SeanFeldman:
Connect to #308 

A few questions raised in the issue. @Particular/azure-service-bus-maintainers please answer those first.

### DIFF
--- a/src/Tests/Creation/When_creating_queues.cs
+++ b/src/Tests/Creation/When_creating_queues.cs
@@ -1,8 +1,11 @@
 namespace NServiceBus.Azure.WindowsAzureServiceBus.Tests.Creation
 {
     using System;
+    using System.Collections.Generic;
     using System.Threading.Tasks;
     using AzureServiceBus;
+    using Config;
+    using Config.ConfigurationSource;
     using Microsoft.ServiceBus;
     using Microsoft.ServiceBus.Messaging;
     using TestUtils;
@@ -452,6 +455,50 @@ namespace NServiceBus.Azure.WindowsAzureServiceBus.Tests.Creation
 
             //cleanup
             await namespaceManager.DeleteQueue("existingqueue");
+        }
+
+        [Test]
+        public async Task Should_not_update_properties_of_an_existing_shared_queue()
+        {
+            var queuePath = "errorQ";
+
+            var namespaceManager = new NamespaceManagerAdapter(NamespaceManager.CreateFromConnectionString(AzureServiceBusConnectionString.Value));
+            if (!await namespaceManager.QueueExists(queuePath).ConfigureAwait(false))
+            {
+                await namespaceManager.CreateQueue(new QueueDescription(queuePath)).ConfigureAwait(false);
+            }
+
+            var settings = new DefaultConfigurationValues().Apply(new SettingsHolder());
+            settings.Set("errorQueue", queuePath);
+            settings.Set("TypesToScan", new List<Type> {typeof(ConfigureAuditQueue)});
+            settings.Set<IConfigurationSource>(null);
+            settings.Set(WellKnownConfigurationKeys.Core.RecoverabilityNumberOfImmediateRetries, 2);
+
+            var extensions = new TransportExtensions<AzureServiceBusTransport>(settings);
+            extensions.Queues().MaxDeliveryCount(2);
+
+            var creator = new AzureServiceBusQueueCreator(settings);
+
+            await creator.Create(queuePath, namespaceManager);
+
+            var real = await namespaceManager.GetQueue(queuePath);
+
+            Assert.That(real.MaxDeliveryCount, Is.EqualTo(10));
+
+            //cleanup
+            await namespaceManager.DeleteQueue(queuePath);
+        }
+
+
+        class ConfigureAuditQueue : IProvideConfiguration<AuditConfig>
+        {
+            public AuditConfig GetConfiguration()
+            {
+                return new AuditConfig
+                {
+                    QueueName = "auditQ"
+                };
+            }
         }
     }
 }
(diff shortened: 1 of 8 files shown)


src/Tests/Creation/When_creating_queues.cs
```diff
@@ -453,5 +456,49 @@ await namespaceManager.CreateQueue(new QueueDescription("existingqueue")
...
+
+            var namespaceManager = new NamespaceManagerAdapter(NamespaceManager.CreateFromConnectionString(AzureServiceBusConnectionString.Value));
+            if (!await namespaceManager.QueueExists(queuePath).ConfigureAwait(false))
+            {
+                await namespaceManager.CreateQueue(new QueueDescription(queuePath)).ConfigureAwait(false);
+            }
+
+            var settings = new DefaultConfigurationValues().Apply(new SettingsHolder());
+            settings.Set("errorQueue", queuePath);
+            settings.Set("TypesToScan", new List<Type> {typeof(ConfigureAuditQueue)});
```
andreasohlund: This sucks, I'll take a look in the core to see why we need to do this

SeanFeldman: That's because TryGet() will throw, had to...

andreasohlund: Ie why we need ConfigureAuditQueue in there

andreasohlund: Yea, we should fix this in the core

SeanFeldman: Are you going to open an issue or should I @andreasohlund ?

andreasohlund:
Lemme take a look first, stay tuned

src/Tests/Creation/When_creating_queues.cs
andreasohlund: Problem is that `GetConfigSection` assumes that there are "available types" in the settings. We could make that a TryGetAvailableTypes or just make it return new List<Type>

https://github.com/Particular/NServiceBus/blob/release-6.0.0/src/NServiceBus.Core/SettingsExtensions.cs#L20

@Particular/nservicebus-maintainers thoughts?

SeanFeldman:
@yvesgoeleven @andreasohlund I've [worked around it](https://github.com/Particular/NServiceBus.AzureServiceBus/pull/333/commits/fc083745d5fd1566f5504c1bf4337c047b9d6665) for now.

SeanFeldman:
The build is green. @Particular/azure-service-bus-maintainers please review

src/Transport/Creation/AzureServiceBusQueueCreator.cs
```diff
@@ -103,6 +133,11 @@ public AzureServiceBusQueueCreator(ReadOnlySettings settings)
             return description;
         }
 
+        bool IsSharedQueue(string queuePath)
```
yvesgoeleven: I'd rename this to IsSystemQueue

src/Transport/Creation/AzureServiceBusQueueCreator.cs
```diff
@@ -40,6 +43,28 @@ public AzureServiceBusQueueCreator(ReadOnlySettings settings)
             }
         }
 
+        void TryGetSystemQueuessAddresses()
```
andreasohlund: I'm starting to doubt that we need this since I think that all the queues passed in the "sending" collection should be consider as shared/system queues? (Audit+Forwarding+Error) 

In short: Treat all queues in https://github.com/Particular/NServiceBus/blob/release-6.0.0/src/NServiceBus.Core/Transports/QueueBindings.cs#L18 as system/shared

Side note: I don't think forwarding should be there, I'll raise a core issue for that

@Particular/nservicebus-maintainers what do you all think?

danielmarbach: Agree with @andreasohlund 
Sending destinations should not be created. We should also stop doing it https://github.com/Particular/NServiceBus/blob/release-6.0.0/src/NServiceBus.Core/Transports/Msmq/QueueCreator.cs#L23 but of course this has other implications for F5ability

src/Tests/Creation/When_creating_queues.cs
timbussmann: sounds ok to me

src/Transport/Creation/AzureServiceBusQueueCreator.cs
andreasohlund: Yea, that's a different issue that we need to deal with in the future

src/Tests/Creation/When_creating_queues.cs
andreasohlund: Let's see, given the below discussion we might not need this here at all

src/Transport/Creation/AzureServiceBusQueueCreator.cs
SeanFeldman: Thank you @andreasohlund for separating "present" from the "future" :)

src/Transport/Creation/AzureServiceBusQueueCreator.cs
SeanFeldman: Good catch!

src/Tests/Creation/When_creating_queues.cs
SeanFeldman: @andreasohlund I worked around the issue. Saying that, `TryGet` should never fail. A different issue :)

SeanFeldman:
Updated. @Particular/azure-service-bus-maintainers ready to be merged.

SeanFeldman:
Hold please. More refactoring coming.

SeanFeldman:
👍 to merge once green.

Issue #340, opened 2016-09-22
Disable prefetching by default in combination with transaction mode none


ramonsmits:
While talking about prefetching I was wondering what the behavior is during graceful shutdown specifically to this question:

- Are message that are already 'prefetched' but not yet processed still processed until the prefetch buffer/queue is depleted

If the answer is no and the customer is running with transaction mode 'None' than we have potential message loss during a 'graceful shutdown' even though no errors occur as the messages in the prefetch buffer are then discarded.

If this behavior is our actual behavior we should write a WARN log entry stating something like the following at startup

> WARN Prefetchsize is larger than 0 while transaction mode is 'None'. This can result in message loss during normal operation.

Also, during shutdown, we should log a similar message for each discarded message if possible.

## Update 2016-09-26

The transport will not set a default `PrefetchCount` if receive mode is `ReceiveAndDelete`, unless user has explicitly configured `PrefetchCount` using an API.

## PoA
- [x] Implement the change PR #346 
- [x] Update documentation to make it clear that `PrefetchCount` **will** be used if configured despite the `ReceiveAndDelete` mode. [PR #1964](https://github.com/Particular/docs.particular.net/pull/1964)

yvesgoeleven:
@ramonsmits they will be gone, running in transaction mode none is very dangerous regardless of prefetch as stated in our docs. I'm ok with adding a warn statement (I'd rather see this mode gone completely)

SeanFeldman:
I wonder if we should automatically disable prefetching if mode is Receive-and-Delete

bording:
> I wonder if think we should automatically disable prefetching if mode is Receive-and-Delete

An alternative would be to not actually use a Receive-and-Delete mode, which is what I ended up doing in Rabbit, for a number of reasons.

SeanFeldman:
What we'll do is disable prefetch by default with `ReceiveAndDelete`, but allow it to be configured through API if needed. While `ReceiveAndDelete` is dangerous, it does allow faster throughput and if loosing individual messages is not a problem, this mode can be helpful. Wherever it fits the overall NSB offer, is a different issue that should be discussed. [All transport support unreliable mode](https://docs.particular.net/nservicebus/transports/transactions) today. Are you saying the doco is incorrect @bording?

bording:
>  All transport support unreliable mode today. Are you saying the doco is incorrect @bording?

[...]

Issue #345, opened 2016-09-26
TransactionModeSettingsExtensions.SupportedTransactionMode() doesn't take into account TransportTransactionMode


SeanFeldman:
Core is setting `settings.Get<TransportTransactionMode>()` to be the required transaction mode that is not used in `TransactionModeSettingsExtensions.SupportedTransactionMode() `

SeanFeldman:
What was missing [utilization of the `ReceiveMode`](https://github.com/Particular/NServiceBus.AzureServiceBus/pull/346/commits/945d7c39d0c77976d3d690824af253a54747be63#diff-f630819b553d88e547573c94435f95bdR15).

Fixed via https://github.com/Particular/NServiceBus.AzureServiceBus/pull/346/commits/945d7c39d0c77976d3d690824af253a54747be63

Issue #344, opened 2016-09-26
transport.Transactions(TransportTransactionMode.None) doesn't set receive mode to ReceiveAndDelete


SeanFeldman:
Expectation would be to see to see receive mode (`WellKnownConfigurationKeys.Connectivity.MessageReceivers.ReceiveMode`) to be lowered to `ReceiveAndDelete`.

Also, is the logic of having the following API makes sense?

```c#
var transport = c.UseTransport<AzureServiceBusTransport>();
transport.MessageReceivers().ReceiveMode(ReceiveMode.ReceiveAndDelete);
transport.MessageReceivers().PrefetchCount(500);
```

Is it expected that transaction mode will be lowered?

There seems to be a disconnect between receive mode and transport transaction mode.

SeanFeldman:
Fixed via https://github.com/Particular/NServiceBus.AzureServiceBus/pull/346/commits/945d7c39d0c77976d3d690824af253a54747be63.  

TransactionModeSettingsExtensions.TransportTransactionMode did not used `ReceiveMode`.

PR #346, opened 2016-09-26
Control prefetch count

SeanFeldman:
Connects to #340 

When transport transaction mode is None or receive mode is ReceiveAndDelete, `PrefetchCount` should not be set, unless explicitly configured by  user code.

### DIFF
--- a/src/Transport/Connectivity/MessageReceiverCreator.cs
+++ b/src/Transport/Connectivity/MessageReceiverCreator.cs
@@ -1,14 +1,12 @@
 namespace NServiceBus.Transport.AzureServiceBus
 {
     using System.Threading.Tasks;
-    using Logging;
     using Microsoft.ServiceBus;
     using Microsoft.ServiceBus.Messaging;
     using Settings;
 
     class MessageReceiverCreator : ICreateMessageReceivers
     {
-        static ILog logger = LogManager.GetLogger(typeof(MessageReceiverCreator));
         IManageMessagingFactoryLifeCycle factories;
         ReadOnlySettings settings;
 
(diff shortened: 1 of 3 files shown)


SeanFeldman:

src/Transport/Connectivity/MessageReceiverCreator.cs
```diff
@@ -22,12 +24,20 @@ public MessageReceiverCreator(IManageMessagingFactoryLifeCycle factories, ReadOn
             var factory = factories.Get(namespaceAlias);
             var receiveMode = settings.Get<ReceiveMode>(WellKnownConfigurationKeys.Connectivity.MessageReceivers.ReceiveMode);
 
-            var receiver = await factory.CreateMessageReceiver(entityPath, receiveMode).ConfigureAwait(false);
+            var prefetchCount = settings.GetOrDefault<int>(WellKnownConfigurationKeys.Connectivity.MessageReceivers.PrefetchCount);
+            var userHasNotProvidedPrefetchCount = !settings.HasExplicitValue(WellKnownConfigurationKeys.Connectivity.MessageReceivers.PrefetchCount);
+            var transportTransactionMode = settings.HasExplicitValue<TransportTransactionMode>() ? settings.Get<TransportTransactionMode>() : settings.SupportedTransactionMode();
```
yvesgoeleven: maybe we should create an extension method on settings for this as well, since we now have this check in multiple places?

SeanFeldman: Update: not needed. Another bug fixed (#344, #345) and now we can accomplish disabling prefetch count (if not specified by user) as part of the default value update during [transport initialization](https://github.com/Particular/NServiceBus.AzureServiceBus/pull/346/commits/945d7c39d0c77976d3d690824af253a54747be63#diff-abc21d576eebe5908bcca158905c4878R73).

SeanFeldman:
@Particular/azure-service-bus-maintainers are we good with these changes?

Issue #347, opened 2016-09-27
Sending messages in batch throws a TransactionSizeExceededException


HEskandari:
This is based on a case (00021184). Looks like we don't take message count when sending a batch. Here's the stack trace:

<details>
<summary>Exception details</summary>
[...]

PR #348, opened 2016-09-27
Transaction with batch > 100 msgs

SeanFeldman:
Connect to #347


SeanFeldman:

PR #349, opened 2016-09-28
MessageReceiverNotifier should not invoke recoverability when receiving

danielmarbach:
This is a proposal PR based on my limited understanding of ASB, related to https://github.com/Particular/NServiceBus/issues/4169. This doesn't have to be merged for this release since it is not really critical.

## Assumptions

While analyzing the code with @tmasternak we came to following conclusions:

* You can only close the internal receiver after all pending receives are completed or failed but
* The azure SDK has no way of removing the callback function OnMessage which means while shutting down messages that are prefetched and potentially more messages will still invoke the `ProcessMessageAsync` as well as add tasks to the to the `pipelineInvocationTasks` dictionary. 
* If we use `ReceiveMode.ReceiveAndDelete` calling `AbandonInternal` has no effect
* If we use `ReceiveMode.PeekLock` we are trying to call `AbondonInternal` to hopefully make the messages earlier available in the input queue
* BrokeredMessages is over its internal receive context directly connected to the message receiver. That's why we cannot close the internal receiver earlier in the stop phase

If those assumptions above are correct then I would like to propose the following change:

## Change

* Since the `stopping `and the `IsRunning `flag is checked by multiple threads in parallel we need to mark the field volatile (minimum we should do)
* Introduced an `ShouldReceiveMessages` phase based on `stopped` and `isRunning` as soon as we are no need to receive messages (when `stopping` is `true`) the callbacks should just return and the recoverability should only be executed when we are in the receive message phase

This will mean the following:

### RecieveAndDelete

Any messages coming in during shutdown will be consumed (like before)

### PeekLock

* Any messages coming in during shutdown will released when the peek lock time is over (change of behavior)

These changes will prevent the following from happen:

1. We are shutting down, pipeline invocations are in flight
1. We asynchronously wait until current snapshot of pipeline invocations is done or the timeout has been reached
1. In the meantime during these 30 seconds messages still fly in and we will try to abondon it under peek lock. This will still track pipeline invocations in the concurrent dictionary! But as soon as the internal receiver is closed (eventually because we received the timeout or because the snapshot only container a few number of pending receives) the Abandon calls will start throwing on SafeAbandonAsync or scope.Complete() / scope.Dispose()
1. Any successful receives that are still pending will call `HandleCompletion` (like before)
1. But now comes the caveat. The previous behavior could means if we still have pending receives that throw, we'll land inside the catch and we would call the core and for FLRs we would try to andandon (although the receiver is already closed) go into the outer catch and try to abandon again (on the closed receiver). 

Overall I would say we are trying to be unnecessary helpful by early abandoning messages while shutting down and therefore run into the exceptions that @ramonsmits saw and I would propose to change it.

### DIFF
--- a/src/Transport/Receiving/MessageReceiverNotifier.cs
+++ b/src/Transport/Receiving/MessageReceiverNotifier.cs
@@ -11,7 +11,7 @@ namespace NServiceBus.Transport.AzureServiceBus
     using Logging;
     using Settings;
     using Transport;
-    
+
     class MessageReceiverNotifier : INotifyIncomingMessages
     {
         IManageMessageReceiverLifeCycle clientEntities;
@@ -25,7 +25,8 @@ namespace NServiceBus.Transport.AzureServiceBus
         ConcurrentDictionary<Task, Task> pipelineInvocationTasks;
         string fullPath;
         EntityInfo entity;
-        bool stopping;
+        volatile bool stopping;
+        volatile bool isRunning;
         ConcurrentStack<Guid> locksTokensToComplete;
         CancellationTokenSource batchedCompletionCts;
         Task batchedCompletionTask;
@@ -41,8 +42,11 @@ namespace NServiceBus.Transport.AzureServiceBus
             batchedCompletionCts = new CancellationTokenSource();
         }
 
-        public bool IsRunning { get; private set; }
+        public bool IsRunning => isRunning;
+
         public int RefCount { get; set; }
+
+        bool ShouldReceiveMessages => !stopping || isRunning;
 
         public void Initialize(EntityInfo entity, Func<IncomingMessageDetails, ReceiveContext, Task> callback, Func<Exception, Task> errorCallback, Func<ErrorContext, Task<ErrorHandleResult>> processingFailureCallback, int maximumConcurrency)
         {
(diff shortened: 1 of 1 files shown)


src/Transport/Receiving/MessageReceiverNotifier.cs
```diff
@@ -198,7 +202,7 @@ void PerformBatchedCompletionTask()
                     }
                 }
             }
-            catch (Exception exception)
+            catch (Exception exception) when(ShouldReceiveMessages)
```
SeanFeldman: Very slick

src/Transport/Receiving/MessageReceiverNotifier.cs
```diff
@@ -163,9 +167,9 @@ void PerformBatchedCompletionTask()
 
         async Task ProcessMessageAsync(BrokeredMessage message)
         {
-            if (stopping || !IsRunning)
+            if (!ShouldReceiveMessages)
             {
-                await AbandonAsync(message).ConfigureAwait(false);
+                logger.Info("Received message while shutting down, abandoning it so we can process it later.");
```
SeanFeldman: Who are "we"? 😃

SeanFeldman: > abandoning it so we can process it later.

Perhaps should say 

>$"Received message with ID {message.MessageId} while shutting down. Message will not be processed and will be retried after {message.LockedUntilUtc}."

SeanFeldman:
@Particular/azure-service-bus-maintainers assuming change is approved, do we want to take this in as part of RC5 (with Core RC3 changes)?

PR #353, opened 2016-09-29
Outbox bug issue

### LINKED ISSUE #352: Outbox doesn't work with ASB transport
## Who's affected

Anyone using Azure Service Bus transport in conjunction with the Outbox feature

## Symptoms 

Handlers process messages and then NHibernate.Exceptions.GenericADOException is thrown with a message `could not execute query, System.ComponentModel.Win32Exception: The wait operation timed out`, eventually moving incoming messages into the error queue.

## Workaround

Set transport transaction mode to `ReceiveOnly` explicitly.

```c#
var transport = endpointConfiguration.UseTransport<AzureServiceBusTransport>();
transport.Transactions(TransportTransactionMode.ReceiveOnly); // lower transport transaction mode
```

SeanFeldman:
Fixes #352 

## Issue

When Outbox is enabled, transport transaction mode is set as the [default setting](https://github.com/Particular/NServiceBus/blob/d061342e8ed69da242514a03e9ca9ce1f96db373/src/NServiceBus.Core/Reliability/Outbox/OutboxConfigExtensions.cs#L20). ASB transport was expecting to find an explicit value instead. As a result of that, the setting was ignored and the mode was not lowered as expected.

### DIFF
--- a/src/Transport/Config/AzureServiceBusTransport.cs
+++ b/src/Transport/Config/AzureServiceBusTransport.cs
@@ -47,7 +47,7 @@
 
         void MatchSettingsToConsistencyRequirements(SettingsHolder settings)
         {
-            if (settings.HasExplicitValue<TransportTransactionMode>())
+            if (settings.HasSetting<TransportTransactionMode>())
             {
                 var required = settings.Get<TransportTransactionMode>();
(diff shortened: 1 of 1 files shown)


bording:
LGTM 👍 You want to be checking for possible default values there as well, and that's what `HasSetting` does.

PR #354, opened 2016-10-04
Performance tuning

yvesgoeleven:
- Client per factory/connection (original design re-used single factory for multiple clients)
- Total concurrency addressed (original design used total concurrency per client rather than distributing among all clients)
- Topology section caching (originally wasn't cached)
- ConnectionStrings caching (originally wasn't cached)

### DIFF
--- a/src/Transport/Receiving/MessageReceiverNotifier.cs
+++ b/src/Transport/Receiving/MessageReceiverNotifier.cs
@@ -17,7 +17,7 @@ namespace NServiceBus.Transport.AzureServiceBus
         IManageMessageReceiverLifeCycle clientEntities;
         IConvertBrokeredMessagesToIncomingMessages brokeredMessageConverter;
         ReadOnlySettings settings;
-        IMessageReceiver internalReceiver;
+        IList<IMessageReceiver> internalReceivers = new List<IMessageReceiver>();
         ReceiveMode receiveMode;
         OnMessageOptions options;
         Func<IncomingMessageDetails, ReceiveContext, Task> incomingCallback;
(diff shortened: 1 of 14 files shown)


SeanFeldman:
@yvesgoeleven looks like things got broken

src/Transport/Receiving/MessageReceiverNotifier.cs
```diff
@@ -104,33 +107,41 @@ public void Start()
...
+                    pipelineInvocationTasks.TryAdd(processTask, processTask);
+                    processTask.ContinueWith(t =>
+                    {
+                        Task toBeRemoved;
+                        pipelineInvocationTasks.TryRemove(t, out toBeRemoved);
+                    }, TaskContinuationOptions.ExecuteSynchronously);
+                    return processTask;
+                };
+
+                isRunning = true;
```
SeanFeldman: This is not part of this PR, but I wonder if we better off to set `isRunning` to `true` when we're done with the receiver? I.e. after `internalReceiver.OnMessage(callback, options);` is invoked?

src/Transport/Seam/MessagePump.cs
```diff
@@ -20,6 +20,7 @@ class MessagePump : IPushMessages, IDisposable
         ILog logger = LogManager.GetLogger(typeof(MessagePump));
         string inputQueue;
         SatelliteTransportAddressCollection satelliteTransportAddresses;
+        SemaphoreSlim throttler = new SemaphoreSlim(1);
```
SeanFeldman: Not needed since will be re-assigned in [`OnStart()`](https://github.com/Particular/NServiceBus.AzureServiceBus/pull/354/files#diff-f62072a55a1af73f0df134b65d530d8cR103)

SeanFeldman:
I've looked into the build again. Purged namespaces, but it didn't help. Random tests fail for both topologies.

SeanFeldman:
https://github.com/Particular/NServiceBus.AzureServiceBus/pull/354/commits/f44348296b9c5362cbf971eaa9727bc2a7ed2a99 has fixed the random failing tests. The code is good.

PR #357, opened 2016-10-07
Concurrency set to zero

yvesgoeleven:
## Who's affected

Any user specifying concurrency limit on NServiceBus endpoint and number of client per entity for Azure Service Bus transport.

## Symptoms

When concurrency limit was lower then number of clients, concurrency could be set to zero and result in exception.

### DIFF
--- a/src/Transport/Receiving/MessageReceiverNotifier.cs
+++ b/src/Transport/Receiving/MessageReceiverNotifier.cs
@@ -67,11 +67,12 @@ namespace NServiceBus.Transport.AzureServiceBus
 
             numberOfClients = settings.Get<int>(WellKnownConfigurationKeys.Connectivity.NumberOfClientsPerEntity);
             var concurrency = maximumConcurrency / (double)numberOfClients;
+            var maxConcurrentCalls = concurrency > 0 ? (int) Math.Round(concurrency, MidpointRounding.AwayFromZero) : 1;
             options = new OnMessageOptions
             {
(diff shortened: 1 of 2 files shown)


src/Transport/Receiving/MessageReceiverNotifier.cs
```diff
@@ -67,11 +67,12 @@ public void Initialize(EntityInfo entity, Func<IncomingMessageDetails, ReceiveCo
 
             numberOfClients = settings.Get<int>(WellKnownConfigurationKeys.Connectivity.NumberOfClientsPerEntity);
             var concurrency = maximumConcurrency / (double)numberOfClients;
+            var maxConcurrentCalls = concurrency > 1 ? (int) Math.Round(concurrency, MidpointRounding.AwayFromZero) : 1;
```
SeanFeldman: Example: concurrency is set to 2 and number of clients to 10. It will cause concurrency to be lifted to 10. I think the least we could do to address it is output a warning message explaining what happened.

SeanFeldman:
In general, if we override user provided values, probably a good think to log a warning with a reason we do so.

yvesgoeleven:
@SeanFeldman in hindsight, there really is no need to raise a warning as we're throttling out the maximumconcurrency anyway, I'll add an informational statement

SeanFeldman:
Replaced by https://github.com/Particular/NServiceBus.AzureServiceBus/pull/364

PR #360, opened 2016-10-10
Speeding up startup

yvesgoeleven:
- removed `lock` statements
- creation of message receivers in parallel

### DIFF
--- a/src/Transport/Connectivity/MessageReceiverLifeCycleManager.cs
+++ b/src/Transport/Connectivity/MessageReceiverLifeCycleManager.cs
@@ -1,5 +1,6 @@
 namespace NServiceBus.Transport.AzureServiceBus
 {
+    using System;
     using System.Collections.Concurrent;
     using System.Threading.Tasks;
     using Settings;
@@ -21,12 +22,22 @@ namespace NServiceBus.Transport.AzureServiceBus
             var buffer = MessageReceivers.GetOrAdd(entityPath + namespaceAlias, s =>
             {
                 var b = new CircularBuffer<EntityClientEntry>(numberOfReceiversPerEntity);
+                var exceptions = new ConcurrentQueue<Exception>();
                 Parallel.For(0, numberOfReceiversPerEntity, i =>
                 {
-                    var e = new EntityClientEntry();
-                    e.ClientEntity = receiveFactory.Create(entityPath, namespaceAlias).GetAwaiter().GetResult();
-                    b.Put(e);
+                    try
+                    {
+                        var e = new EntityClientEntry();
+                        e.ClientEntity = receiveFactory.Create(entityPath, namespaceAlias).GetAwaiter().GetResult();
+                        b.Put(e);
+                    }
+                    catch (Exception ex)
+                    {
+                        exceptions.Enqueue(ex);
+                    }
+                   
                 });
+                if (exceptions.Count > 0) throw new AggregateException(exceptions);
                 return b;
             });
 
--- a/src/Transport/Connectivity/MessageReceiverLifeCycleManager.cs
+++ b/src/Transport/Connectivity/MessageReceiverLifeCycleManager.cs
@@ -21,10 +21,10 @@ namespace NServiceBus.Transport.AzureServiceBus
             var buffer = MessageReceivers.GetOrAdd(entityPath + namespaceAlias, s =>
             {
                 var b = new CircularBuffer<EntityClientEntry>(numberOfReceiversPerEntity);
-                Parallel.For(0, numberOfReceiversPerEntity, async i =>
+                Parallel.For(0, numberOfReceiversPerEntity, i =>
                 {
                     var e = new EntityClientEntry();
-                    e.ClientEntity = await receiveFactory.Create(entityPath, namespaceAlias).ConfigureAwait(false);
+                    e.ClientEntity = receiveFactory.Create(entityPath, namespaceAlias).GetAwaiter().GetResult();
                     b.Put(e);
                 });
                 return b;
--- a/src/Transport/Connectivity/MessageReceiverLifeCycleManager.cs
+++ b/src/Transport/Connectivity/MessageReceiverLifeCycleManager.cs
@@ -1,8 +1,6 @@
 namespace NServiceBus.Transport.AzureServiceBus
 {
-    using System;
     using System.Collections.Concurrent;
-    using System.Threading.Tasks;
     using Settings;
 
     class MessageReceiverLifeCycleManager : IManageMessageReceiverLifeCycle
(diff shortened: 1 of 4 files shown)


ramonsmits:
@yvesgoeleven I've tested [build 8 of this PR](https://builds.particular.net/viewLog.html?buildId=191172&tab=artifacts&buildTypeId=NServiceBusAzure_NServiceBusAzureServiceBus_1Build#!12q45qa) and this reduced the startup time from over 2m30s to 9 seconds when setting both *NumberOfMessagingFactoriesPerNamespace* and *NumberOfClientsPerEntity* to 64.

src/Transport/Connectivity/MessageReceiverLifeCycleManager.cs
```diff
@@ -20,13 +20,10 @@ public IMessageReceiver Get(string entityPath, string namespaceAlias)
...
                 var b = new CircularBuffer<EntityClientEntry>(numberOfReceiversPerEntity);
-                for (var i = 0; i < numberOfReceiversPerEntity; i++)
+                for(var i = 0; i < numberOfReceiversPerEntity; i++)
                 {
                     var e = new EntityClientEntry();
-                    lock (e.Mutex)
-                    {
-                        e.ClientEntity = receiveFactory.Create(entityPath, namespaceAlias).GetAwaiter().GetResult();
-                    }
+                    e.ClientEntity = receiveFactory.Create(entityPath, namespaceAlias).GetAwaiter().GetResult();
```
SeanFeldman: Since `MessageReceivers` is a concurrent dictionary, lock is not needed I presume @yvesgoeleven

SeanFeldman: Should we expose a synchronous `Create()` method rather than forcing async version w/o awaiting?

src/Transport/Receiving/MessageReceiverNotifier.cs
```diff
@@ -108,38 +108,45 @@ public void Start()
...
+                    if (internalReceiver == null)
                     {
-                        Task toBeRemoved;
-                        pipelineInvocationTasks.TryRemove(t, out toBeRemoved);
-                    }, TaskContinuationOptions.ExecuteSynchronously);
-                    return processTask;
-                };
-
-                isRunning = true;
+                        throw new Exception($"MessageReceiverNotifier did not get a MessageReceiver instance for entity path {fullPath}, this is probably due to a misconfiguration of the topology");
```
SeanFeldman: I have to admit that reading this message it's more confusing than helping. 

> this is probably due to a misconfiguration of the topology

Error message should either help to understand what exactly the problem is or how to fix it. Don't think this error message does either well.

src/Transport/Receiving/MessageReceiverNotifier.cs
```diff
@@ -108,38 +108,45 @@ public void Start()
             stopping = false;
             pipelineInvocationTasks = new ConcurrentDictionary<Task, Task>();
 
-            
-            for(var i = 0; i < numberOfClients; i++)
+            var exceptions = new ConcurrentQueue<Exception>();
+            Parallel.For(0, numberOfClients, i =>
```
SeanFeldman: Why do we use `Parallel.For()` and not `Task.WhenAll()`?

yvesgoeleven: If we change e.ClientEntity to Task, then it can become async as well, will do that change

src/Transport/Receiving/MessageReceiverNotifier.cs
yvesgoeleven: what message do you suggest?

src/Transport/Receiving/MessageReceiverNotifier.cs
yvesgoeleven: Can't do this as the interfaces IManageMessageSenderLifeCycle and IManageMessageReceiverLifeCycle are public it would need to change to return a task instead, so need to wait until our next major

ramonsmits: @yvesgoeleven even then its not possible as `CircularBuffer` isn't async. Maybe the only thing that works is to make the circular buffer contain Lazy items of Task<IMessageReceiver>. So something like:

`CircularBuffer<Lazy<Task<IMessageReceiver>>>`

That would resolve the item async on first access but I'm not sure if that even works :-) @danielmarbach any guidance? Does my idea make sense?

src/Transport/Connectivity/MessageSenderLifeCycleManager.cs
```diff
@@ -20,15 +20,13 @@ public IMessageSender Get(string entitypath, string viaEntityPath, string namesp
             var buffer = MessageSenders.GetOrAdd(entitypath + viaEntityPath + namespaceName, s =>
             {
                 var b = new CircularBuffer<EntityClientEntry>(numberOfSendersPerEntity);
-                for (var i = 0; i < numberOfSendersPerEntity; i++)
+                for(var i = 0; i < numberOfSendersPerEntity; i++)
                 {
                     var e = new EntityClientEntry();
-                    lock (e.Mutex)
```
ramonsmits: As the Mutex is gone, you can probably remove the whole `EntityClientEntry` and just create a `CircularBuffer<IMessageReceiver>`

ramonsmits: I see that this is not possible as you are replacing instances in the buffer when 'closed'.

yvesgoeleven: @ramonsmits no, entities can break so need to be replaced, you can't update the ringbuffer

yvesgoeleven: will still need to lock when repairing btw, don't want to give out a closed entity

ramonsmits: In that case, wouldn't you want to 'Get' the next and evict the current? Not sure how expensive it is and how frequent this happens.

src/Transport/Receiving/MessageReceiverNotifier.cs
yvesgoeleven: @ramonsmits you can await a task more then once, it just returns the result the second time, no need for lazy

src/Transport/Connectivity/MessageSenderLifeCycleManager.cs
yvesgoeleven: ring buffers are fixed in size, you can't evict the entry in it

src/Transport/Receiving/MessageReceiverNotifier.cs
danielmarbach: @ramonsmits I brought this up in last weeks review. We can definitely improve here. In general @yvesgoeleven is right though you can await the cached task multiple times. @yvesgoeleven @SeanFeldman maybe we can pair on this one today

src/Transport/Connectivity/MessageReceiverLifeCycleManager.cs
SeanFeldman: Separate issue is here #362

SeanFeldman:
@ramonsmits should go out as 7.0.1

PR #366, opened 2016-10-12
hotfix-7.0.1

### LINKED ISSUE #365: Unclean transport shutdown where not all batched message completions are awaited
## Who's affected

Anyone using Azure Service Bus transport version 7.0.0

## Symptoms

When shutting down, the transport is supposed to complete all messages in-flights within 30 seconds from shutdown request. When multiple receivers are used, only a single receiver is awaited, causing transport to close receiver connections w/o waiting for the rest of batched completion tasks to handle their work. This could results in interrupting native client work upon ASB transport shutdown which would lead to native ASB client throwing exceptions.

SeanFeldman:
fixes #365 
fixes #364 (cherry picked from develop)


SeanFeldman:
@Particular/azure-host-maintainers please review (addition of #365)

yvesgoeleven:
👍

PR #373, opened 2016-10-14
Removed Math.Max with 0

ramonsmits:


src/Transport/Config/DefaultConfigurationValues.cs
```diff
@@ -48,7 +48,7 @@ void ApplyDefaultValuesForAddressing(SettingsHolder settings)
 
         void ApplyDefaultsForConnectivity(SettingsHolder settings)
         {
-            var numberOfLogicalCores = Math.Max(0, Environment.ProcessorCount);
+            var numberOfLogicalCores = Environment.ProcessorCount;
```
SeanFeldman: We probably want to do something like Core does: `Max(2, Environment.ProcessorCount)`.

SeanFeldman:
@yvesgoeleven please update the description using the [standard format](https://github.com/Particular/PlatformDevelopment/blob/master/guidelines/release-notes.md)

ramonsmits:
@SeanFeldman What happens if the customer explicitly sets the max concurrency limit to 1 on the core API? Seems that `numberOfLogicalCores` now has an incorrect value. Can't this value be retrieved from the core settings?

yvesgoeleven:
@ramonsmits we only set the default value, any explicit value set by the user will be overriding ours

ramonsmits:
@yvesgoeleven I'm talking about this line : https://github.com/Particular/NServiceBus.AzureServiceBus/pull/373/files#diff-617257f585c77c4d07bc31bc7ca44fd0R51

What if the user doesn't set any ASB specific setting but does set the max concurrency to 1 on the core API?

yvesgoeleven:
@ramonsmits the message pump throttles as well, so the additional clients will kindof behave like a prefetch, but no parallel execution of pipeline will occur

ramonsmits:
And in this example:

- Lease duration is 30s
- core API set to sequential processing (max concurrency is1)
- average message processing duration is 20 seconds

Will we now lease issues because NumberOfClientsPerEntity is set to 2?

SeanFeldman:
@ramonsmits yes, message lock tokens will be lost. Increase lock duration?

ramonsmits:
From the customer perspective I find it strange that this happens. So this means the customer requires to call NumberOfClientsPerEntity too and set it to 1?

Issue #337, opened 2016-09-21
ObjectDisposedException is thrown when shutting endpoint down


ramonsmits:
## Who's affected

* Anyone using Azure Service Bus

## Symptoms

An exception `System.ObjectDisposedException` with message "Cannot access a disposed object." is thrown when endpoint is shutting down

## Original issue

Getting the following exception at endpoint shutdown with 7.0.0-rc0003, this doesnt seem to be happening with 7.0.0-beta0013

> System.ObjectDisposedException: Cannot access a disposed object.
Object name: 'Microsoft.ServiceBus.Messaging.Sbmp.RedirectBindingElement+RedirectContainerChannelFactory`1[System.ServiceModel.Channels.IRequestSessionChannel]'.

I'm getting this primarily when using transaction modes None and Atomic.

<details>
  <summary>Show stacktrace</summary>
```
2016-09-21 10:51:01.6582|INFO|4|NServiceBus.Transport.AzureServiceBus.MessagePump|Messagepump stopped
2016-09-21 10:51:01.6582|FATAL|1|Host.Program|Main
System.ObjectDisposedException: Cannot access a disposed object.
Object name: 'Microsoft.ServiceBus.Messaging.Sbmp.RedirectBindingElement+RedirectContainerChannelFactory`1[System.ServiceModel.Channels.IRequestSessionChannel]'.
Server stack trace: 
   at System.ServiceModel.Channels.CommunicationObject.ThrowIfDisposed()
   at System.ServiceModel.Channels.ChannelFactoryBase`1.ValidateCreateChannel()
   at System.ServiceModel.Channels.ChannelFactoryBase`1.InternalCreateChannel(EndpointAddress address, Uri via)
   at Microsoft.ServiceBus.Messaging.Channels.SharedChannel`1.CreateChannelAsyncResult.<GetAsyncSteps>d__6.MoveNext()
   at Microsoft.ServiceBus.Messaging.IteratorAsyncResult`1.EnumerateSteps(CurrentThreadType state)
   at Microsoft.ServiceBus.Messaging.IteratorAsyncResult`1.Start()
Exception rethrown at [0]: 
   at Microsoft.ServiceBus.Common.AsyncResult.End[TAsyncResult](IAsyncResult result)
   at Microsoft.ServiceBus.Messaging.Channels.SharedChannel`1.OnEndCreateInstance(IAsyncResult asyncResult)
[...]

PR #377, opened 2016-10-15
[WIP][SPIKE] Closing factories only once

SeanFeldman:
**WARNING**: Do NOT merge!

Related to #337, ObjectDisposedException thrown during shutdown, (first option described in https://github.com/Particular/NServiceBus.AzureServiceBus/issues/337#issuecomment-253947633).

Added `IStoppableTopology` interface to allow `AzureServiceBusTransport` during `Stop()` operation invoke topology to close factories. Operation that will take place once, after message receivers are closed.

@Particular/azure-service-bus-maintainers this is a breaking change and modification to the design we'll have to analyze and review. A candidate for a minor or next major as a breaking change.

### DIFF
--- a/src/Transport/Config/AzureServiceBusTransportInfrastructure.cs
+++ b/src/Transport/Config/AzureServiceBusTransportInfrastructure.cs
@@ -3,6 +3,7 @@
     using System;
     using System.Collections.Generic;
     using System.Text;
+    using System.Threading.Tasks;
     using DelayedDelivery;
     using NServiceBus.AzureServiceBus.Topology.MetaModel;
     using Performance.TimeToBeReceived;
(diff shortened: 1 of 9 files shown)


src/Transport/Config/AzureServiceBusTransportInfrastructure.cs
```diff
@@ -14,6 +15,11 @@ class AzureServiceBusTransportInfrastructure : TransportInfrastructure
         ITopology topology;
         SatelliteTransportAddressCollection satelliteTransportAddresses;
 
+        public override Task Stop()
+        {
+            return ((IStoppableTopology) topology).Stop();
```
yvesgoeleven: why not just resolve the factory lifecycle manager here and stop all? Why introduce a new interface and then implement it twice?

yvesgoeleven: right, we'd have to expose the container from the topology, that may not be a good idea

danielmarbach:
Replaced by https://github.com/Particular/NServiceBus.AzureServiceBus/pull/380 
Cherry picked @SeanFeldman commit

danielmarbach:
Closed with the blessing of @yvesgoeleven

PR #386, opened 2016-10-21
Make IStoppableTopology public

SeanFeldman:
## Who's affected

- Anyone implementing a custom topology with Azure Service Bus

## Symptoms

- `ForwardingTopology` and `EndpointOrientedTopology` introduced `Stop()` in a hotfix w/o a taking into consideration custom topologies. Therefore, custom topologies 

## Original issue

In version 7.0.2 both topologies introduced a public `Stop()` method which was a breaking SemVer change. In case a topology has special needs when a transport is stopping, there should be a contract to allow so. As `ITopology` cannot be modified until the next major, `IStoppableTopology` is introduced to allow the required behaviour.


SeanFeldman:

PR #393, opened 2016-10-26
Use first topic in events bundle

### LINKED ISSUE #391: First topic in events bundle is never used
## Who's affected

- Anyone using `ForwardingTopology` and working with events on Azure Service Bus 

## Symptoms

With the `ForwardingTopology` when events are sent, they are randomly distributed between bundled topics for better performance. By default, there are 2 topics in a bundle, but only one is utilized. 

## Details

Bundles are used for performance. This bug does not require an immediate upgrade.

SeanFeldman:
Fixes #391 


yvesgoeleven:
👍

PR #398, opened 2016-10-27
Unsubscribe from exceptions

danielmarbach:
When we are shutting down we want to immediately unsubscribe from OnMessage API\`s `ExceptionReceived` event since we are closing the message receivers and we don't want to unnecessarily trigger the circuit breaker as well as risking to run into double shutdown over the critical error action.


danielmarbach:

PR #400, opened 2016-11-01
Remove physical subscriptions on unsubscribe

SeanFeldman:
Removing physical subscription entity on unsubscribe.

### DIFF
--- a/src/Transport/Creation/AzureServiceBusForwardingSubscriptionCreator.cs
+++ b/src/Transport/Creation/AzureServiceBusForwardingSubscriptionCreator.cs
@@ -2,12 +2,13 @@
 {
     using System;
     using System.Collections.Concurrent;
+    using System.Linq;
     using System.Threading.Tasks;
     using Microsoft.ServiceBus.Messaging;
     using Logging;
     using Settings;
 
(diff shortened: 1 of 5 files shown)


src/Transport/Creation/AzureServiceBusForwardingSubscriptionCreator.cs
```diff
@@ -2,12 +2,13 @@
...
     using System;
     using System.Collections.Concurrent;
+    using System.Linq;
     using System.Threading.Tasks;
     using Microsoft.ServiceBus.Messaging;
     using Logging;
     using Settings;
 
-    class AzureServiceBusForwardingSubscriptionCreator : ICreateAzureServiceBusSubscriptions
+    class AzureServiceBusForwardingSubscriptionCreator : ICreateAzureServiceBusSubscriptions, ICreateAzureServiceBusSubscriptionsAbleToDeleteSubscriptions
```
yvesgoeleven: This interface shouldn't be public, right? https://github.com/Particular/NServiceBus.AzureServiceBus/blob/develop/src/Transport/Creation/ICreateAzureServiceBusSubscriptions.cs#L12

SeanFeldman: Correct, fixed.

PR #388, opened 2016-10-24
Internalizing public types

SeanFeldman:
## Who's affected
- Anyone extending and customizing Azure Service Bus transport version 7 and above
## Symptoms

Version 7.0 of the ASB transport has exposed internal types that shouldn't have been exposed.
## Original issue

<details><summary>

Details</summary>

The following types have been exposed and broken down into several categories for a review.

**Types that should be exposed**

✔️ IIndividualizationStrategy
✔️ IHandleOversizedBrokeredMessages
✔️ ISanitizationStrategy
✔️ ICompositionStrategy
✔️ INamespacePartitioningStrategy

**Types that should NOT have been exposes at all**

❌ ITopology
❌ IStoppableTopology
❌ ITransportPartsContainer
❌ IResolveTransportParts  
❌ IRegisterTransportParts
❌ INotifyIncomingMessages
❌ IClientSideSubscriptionFilter
❌ INamespaceManager ==|> NamespaceManagerAdapter

**Types that require intimate knowledge of topologies and/or transport**

❗️ ICreateAzureServiceBusSubscriptions ==|> AzureServiceBusSubscriptionCreator + AzureServiceBusSubscriptionCreatorV6 / AzureServiceBusForwardingSubscriptionCreator (creator per topology)
❗️  ITopologySectionManager ==|> EndpointOrientedTopologySectionManager / ForwardingTopologySectionManager
❗️ IOperateTopology ==|> TopologyOperator
❗️ IBatcher ==|> Batcher / Batch / BatchedOperation
❗️ IRouteOutgoingBatches ==|> DefaultOutgoingBatchRouter / ReceiveContext
❗️ IManageMessageSenderLifeCycle ==|> MessageSenderLifeCycleManager
❗️ ICreateMessageSenders ==|> MessageSenderCreator
❗️ IManageMessagingFactoryLifeCycle ==|> MessagingFactoryLifeCycleManager
❗️ ICreateMessagingFactories ==|> MessagingFactoryCreator
❗️ ICreateMessageReceivers ==|> MessageReceiverCreator
❗️ IManageMessageReceiverLifeCycle ==|> MessageReceiverLifeCycleManager
❗️ IConvertBrokeredMessagesToIncomingMessages ==|> DefaultBrokeredMessagesToIncomingMessagesConverter is the default. Custom implementation not possible since it would require ICanMapConnectionStringToNamespaceAlias which is an internal contract
[...]

PR #403, opened 2016-11-02
RuntimeNamespaceInfo should not be internal

SeanFeldman:
If `RuntimeNamespaceInfo` is internalized, no custom partitioning strategy can be created.
Creating custom partitioning strategies should still be allowed in version 7.1.0


SeanFeldman:
`RuntimeNamespaceInfo` was internalized In #388, this PR reverts that.

boblangley:
Merged #403

danielmarbach:
Why has this been merged? I think we should take our time to discuss and come up with a long-term approach. I'll send in a proposal and a revert PR

Issue #404, opened 2016-11-02
Creating custom partitioning strategies


SeanFeldman:
#@internal: changes associated with internalization of types done in #388 

As part of ASB tranport API, we allow custom partitioning strategies to be [registered](https://docs.particular.net/nservicebus/azure-service-bus/configuration/full#physical-addressing-logic-namespace-partitioning).
To be able to access the namespaces registered with the transport, internally `NamespaceConfigurations` type is used, which is also exposed in a [sample we provide](https://docs.particular.net/samples/azure/custom-partitioning-asb/#subscriber-creating-a-custom-partitioning-strategy) how to create custom partitioning strategies. 

Should we revert internalization of `NamespaceConfigurations` and make it publicly available again or
review `namespacePartitioning.UseStrategy<T>()` API? The API doesn't allow registration of an object that could be initialized with the needed information about namespaces.

@Particular/azure-service-bus-maintainers @danielmarbach this is blocking release 7.1.0

**Update**:

Decision is to not obsolete this with a warning message. Release 8.0.0 will have a different way of registering partitioning strategy and API for building custom partitioning strategies. At that point we'll obsolete the involved types with errors and appropriate guidance.

boblangley:
Perhaps we should just handle the conversion for the user and provide them only RuntimeNamespaceInfo via an extension method in ReadOnlySettings? Here is a PR #405

SeanFeldman:
@yvesgoeleven I agree we should revert internalization of `NamespaceConfigurations` for now and leave it as-is for v7.x 

@danielmarbach attempted a different approach in PR #407 , which is probably a good way for vNext, but not the current solution.

danielmarbach:
#437 was merged. Closing

PR #405, opened 2016-11-03
[SPIKE] Only expose RuntimeNamespaceInfo through ReadOnlySettings

boblangley:
One possible solution to #404

### DIFF
--- a/src/Transport/Addressing/Partitioning/ReadOnlySettingsExtensions.cs
+++ b/src/Transport/Addressing/Partitioning/ReadOnlySettingsExtensions.cs
@@ -1,0 +1,22 @@
+﻿namespace NServiceBus.AzureServiceBus.Addressing.Partitioning
+{
+    using System.Collections.Generic;
+    using System.Configuration;
+    using System.Linq;
+    using Settings;
+    using Transport.AzureServiceBus;
+
+    public static class ReadOnlySettingsExtensions
+    {
+        public static List<RuntimeNamespaceInfo> GetPartitioningNamespaces(this ReadOnlySettings settings)
+        {
+            NamespaceConfigurations namespaces;
+            if (!settings.TryGet(WellKnownConfigurationKeys.Topology.Addressing.Namespaces, out namespaces))
+            {
+                throw new ConfigurationErrorsException($"No namespaces have been defined, please configure the connection string to your azure servicebus namespace. Please use {nameof(AzureServiceBusTransportExtensions.NamespacePartitioning)}().{nameof(AzureServiceBusNamespacePartitioningSettings.AddNamespace)}() to register one or more namespaces.");
+            }
+
+            return namespaces.Where(n => n.Purpose == NamespacePurpose.Partitioning).Select(n => new RuntimeNamespaceInfo(n)).ToList();
+        }
+    }
+}
--- a/src/Transport/Addressing/Partitioning/Strategies/FailOverNamespacePartitioning.cs
+++ b/src/Transport/Addressing/Partitioning/Strategies/FailOverNamespacePartitioning.cs
@@ -3,21 +3,17 @@
     using System.Collections.Generic;
     using System.Configuration;
     using System.Linq;
+    using AzureServiceBus.Addressing.Partitioning;
     using Transport.AzureServiceBus;
     using Settings;
 
     public class FailOverNamespacePartitioning : INamespacePartitioningStrategy
     {
-        NamespaceConfigurations namespaces;
+        List<RuntimeNamespaceInfo> namespaces;
 
         public FailOverNamespacePartitioning(ReadOnlySettings settings)
         {
-            if (!settings.TryGet(WellKnownConfigurationKeys.Topology.Addressing.Namespaces, out namespaces))
-            {
-                throw new ConfigurationErrorsException($"The '{nameof(FailOverNamespacePartitioning)}' strategy requires exactly two namespaces to be configured, please use {nameof(AzureServiceBusTransportExtensions.NamespacePartitioning)}().{nameof(AzureServiceBusNamespacePartitioningSettings.AddNamespace)}() to register the namespaces.");
-            }
-
-            namespaces = new NamespaceConfigurations(namespaces.Where(n => n.Purpose == NamespacePurpose.Partitioning).ToList());
+            namespaces = settings.GetPartitioningNamespaces();
 
             if (namespaces.Count < 2)
             {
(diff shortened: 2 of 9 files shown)


src/Transport/Addressing/Partitioning/ReadOnlySettingsExtensions.cs
```diff
@@ -0,0 +1,22 @@
...
+{
+    using System.Collections.Generic;
+    using System.Configuration;
+    using System.Linq;
+    using Settings;
+    using Transport.AzureServiceBus;
+
+    public static class ReadOnlySettingsExtensions
+    {
+        public static List<RuntimeNamespaceInfo> GetPartitioningNamespaces(this ReadOnlySettings settings)
```
SeanFeldman: This will internalize `NamespaceConfigurations`, but not `ReadOnlySettings`. 
What the usage code will look like?
```c#
endpointConfiguration.GetSettings().GetPartitioningNamespaces()
```
Then it's counter-intuitive to the extension point `transport.NamespacePartitioning().xyz`

boblangley:
Forced push API settings

src/Transport/Addressing/Partitioning/ReadOnlySettingsExtensions.cs
SeanFeldman: Custom topology will still rely on injected `ReadOnlySettings` and if we still do that, I am not sure what's the benefit of hiding `NamespaceConfigurations`. Might just ship it as public for now and introduce a proper (breaking) API in the next major.

boblangley: I think we are looking at it from a different perspective? I wasn't thinking `transport.NamespacePartitioning().UseStrategy<T>()` needed to change.

The only change would need to be our example of how to implement a custom partitioning strategy, which would change to something like this :

```
public class ReplicatedNamespacePartitioningStrategy : INamespacePartitioningStrategy
{
    static ILog log = LogManager.GetLogger<ReplicatedNamespacePartitioningStrategy>();
    List<RuntimeNamespaceInfo> namespaces;

    public ReplicatedNamespacePartitioningStrategy(ReadOnlySettings settings)
    {
        namespaces = settings.GetPartitioningNamespaces();
        if ( namespaces.Count <= 1)
        {
           throw new Exception("The 'Replicated' namespace partitioning strategy requires more than one namespace. Configure additional connection strings");
        }
    }

    public IEnumerable<RuntimeNamespaceInfo> GetNamespaces(PartitioningIntent partitioningIntent)
    {
        log.Info($"Determining namespace for {partitioningIntent}");
        return namespaces.Select(
            selector: namespaceInfo =>
            {
                log.Info($"Choosing namespace {namespaceInfo.Alias} ({namespaceInfo.ConnectionString})");
                return namespaceInfo;
            });
    }
}
```

SeanFeldman: Looking at the sample above, why do we even need this strategy? All it does is returning the found `RuntimeNamespaceInfo` entities, doesn't it?

src/Transport/Addressing/Partitioning/Strategies/SingleNamespacePartitioning.cs
```diff
@@ -3,21 +3,17 @@ namespace NServiceBus
...
     public class SingleNamespacePartitioning : INamespacePartitioningStrategy
     {
-        NamespaceConfigurations namespaces;
+        List<RuntimeNamespaceInfo> namespaces;
 
         public SingleNamespacePartitioning(ReadOnlySettings settings)
         {
-            if (!settings.TryGet(WellKnownConfigurationKeys.Topology.Addressing.Namespaces, out namespaces))
-            {
-                throw new ConfigurationErrorsException($"The '{nameof(SingleNamespacePartitioning)}' strategy requires exactly one namespace, please configure the connection string to your azure servicebus namespace.");
```
SeanFeldman: We're losing the error message which would explain why strategy is not configured properly.

src/Transport/Addressing/Partitioning/Strategies/FailOverNamespacePartitioning.cs
```diff
@@ -3,21 +3,17 @@
...
     public class FailOverNamespacePartitioning : INamespacePartitioningStrategy
     {
-        NamespaceConfigurations namespaces;
+        List<RuntimeNamespaceInfo> namespaces;
 
         public FailOverNamespacePartitioning(ReadOnlySettings settings)
         {
-            if (!settings.TryGet(WellKnownConfigurationKeys.Topology.Addressing.Namespaces, out namespaces))
-            {
-                throw new ConfigurationErrorsException($"The '{nameof(FailOverNamespacePartitioning)}' strategy requires exactly two namespaces to be configured, please use {nameof(AzureServiceBusTransportExtensions.NamespacePartitioning)}().{nameof(AzureServiceBusNamespacePartitioningSettings.AddNamespace)}() to register the namespaces.");
```
SeanFeldman: We're losing the error message which would explain why strategy is not configured properly.

src/Transport/Addressing/Partitioning/Strategies/RoundRobinNamespacePartitioning.cs
```diff
@@ -3,30 +3,24 @@ namespace NServiceBus
...
     {
-        CircularBuffer<NamespaceInfo> namespaces;
+        CircularBuffer<RuntimeNamespaceInfo> namespaces;
 
         public RoundRobinNamespacePartitioning(ReadOnlySettings settings)
         {
-            NamespaceConfigurations namespaces;
-            if (!settings.TryGet(WellKnownConfigurationKeys.Topology.Addressing.Namespaces, out namespaces))
-            {
-                throw new ConfigurationErrorsException($"The '{nameof(RoundRobinNamespacePartitioning)}' strategy requires more than one namespace, please use {nameof(AzureServiceBusTransportExtensions.NamespacePartitioning)}().{nameof(AzureServiceBusNamespacePartitioningSettings.AddNamespace)}() to register multiple namespaces");
```
SeanFeldman: We're losing the error message which would explain why strategy is not configured properly.

src/Transport/Addressing/Partitioning/Strategies/SingleNamespacePartitioning.cs
boblangley: This exception was redundant. All strategies were doing something similar, throwing if namespaces are not configured so I moved that check and exception throw to GetPartitioningNamespaces().

The strategy specific configuration exception is still below this on line 20.

SeanFeldman:
@boblangley and myself have discussed this and we've identified that internalizing one type not going to address the true issue with #404 . Suggest we close this PR and revisit the approach for vNext when registration of dependencies will change. Thoughts?

boblangley:
Agreed, Closed #405

PR #407, opened 2016-11-03
[SPIKE] Namespace partitioning

danielmarbach:
This shows a potential approach to #404 as an addition to #405 

* Introduces `IPartitioningStrategy` which has an explicit lifecycle and doesn't need to rely on settings
 * In the initialization phase the strategy would get all the namespaces that have intent partitioning
 * In the runtime phase it will get called multiple times and can return the RuntimeNamespaceInfo
* Removed obsolete with warn for `NamespaceInfo`

Benefits:
- First move into readonly settings free world
- More intention revealing
- Ability for users to pass in their own instance
- Would address https://github.com/Particular/docs.particular.net/issues/2153
- Proves that it is possible to implement a strategy by just following the contract

Here comes the caveat:

- Because of backwards compatibility the existing types need to be working and we have to redundantly implement both paths
- Tests would only target the new interfaces which could be a problem when we need to hotfix something
- It might be complex to find the right entry point to initialize the strategy with the namespace information because currently it relies heavily on DI magic (btw. the type is not a singleton and will be created all over again for each topology manager entry point and therefore filter the namespace information over and over again)
- Vocabulary of `UseStrategy` is no longer used, instead we'd use `With` or anything similar (necessary evil)

The implementation is not finished to not waste too much time

### DIFF
--- a/src/Tests/Addressing/NamespacePartitioning/When_using_failover_namespace_strategy.cs
+++ b/src/Tests/Addressing/NamespacePartitioning/When_using_failover_namespace_strategy.cs
@@ -2,34 +2,25 @@ namespace NServiceBus.Azure.WindowsAzureServiceBus.Tests.Addressing.NamespacePar
 {
     using System.Configuration;
     using System.Linq;
-    using Settings;
     using NUnit.Framework;
     using Transport.AzureServiceBus;
 
-#pragma warning disable 618
     [TestFixture]
     [Category("AzureServiceBus")]
     public class When_using_failover_namespace_strategy
     {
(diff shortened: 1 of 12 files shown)


danielmarbach:
@Particular/azure-service-bus-maintainers let's discuss this and #405

SeanFeldman:
Good spike. What I'm questioning here is the "butterfly effect" it has on the entire transport with DI. 
How about we drop this PR from 7.1.0 and go with #404 only? For vNext we do this change as it's a bigger of a scope, going beyond partitioning strategies.

danielmarbach:
Good with that.

SeanFeldman:
Closing

PR #408, opened 2016-11-03
Using dedicated OnMessageOptions per receiver

danielmarbach:
Attempt to tackle the shutdown issues we see. I figured it cannot hurt to have dedicated message options per receiver. With that each receiver attaches the event handler to its options. When we stop we remove the event handler from all the options. This should avoid races with the handler unsubscription.

The implementation of message options uses the race to initialize pattern

```
    public event EventHandler<ExceptionReceivedEventArgs> ExceptionReceived
    {
      add
      {
        EventHandler<ExceptionReceivedEventArgs> eventHandler = this.ExceptionReceived;
        EventHandler<ExceptionReceivedEventArgs> comparand;
        do
        {
          comparand = eventHandler;
          eventHandler = Interlocked.CompareExchange<EventHandler<ExceptionReceivedEventArgs>>(ref this.ExceptionReceived, (EventHandler<ExceptionReceivedEventArgs>) Delegate.Combine((Delegate) comparand, (Delegate) value), comparand);
        }
        while (eventHandler != comparand);
      }
      remove
      {
        EventHandler<ExceptionReceivedEventArgs> eventHandler = this.ExceptionReceived;
        EventHandler<ExceptionReceivedEventArgs> comparand;
        do
        {
          comparand = eventHandler;
          eventHandler = Interlocked.CompareExchange<EventHandler<ExceptionReceivedEventArgs>>(ref this.ExceptionReceived, (EventHandler<ExceptionReceivedEventArgs>) Delegate.Remove((Delegate) comparand, (Delegate) value), comparand);
        }
        while (eventHandler != comparand);
      }
    }
```

Which technically should work but since the message options are intended by pump and we start them in parallel it makes sense to have a dedicated option per receiver any way

### DIFF
--- a/src/Transport/Receiving/MessageReceiverNotifier.cs
+++ b/src/Transport/Receiving/MessageReceiverNotifier.cs
@@ -43,23 +43,17 @@ namespace NServiceBus.Transport.AzureServiceBus
                 fullPath = SubscriptionClient.FormatSubscriptionPath(topic.Target.Path, entity.Path);
             }
 
+            autoRenewTimeout = settings.Get<TimeSpan>(WellKnownConfigurationKeys.Connectivity.MessageReceivers.AutoRenewTimeout);
             numberOfClients = settings.Get<int>(WellKnownConfigurationKeys.Connectivity.NumberOfClientsPerEntity);
             var concurrency = maximumConcurrency/(double) numberOfClients;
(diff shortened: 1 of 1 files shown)


SeanFeldman:
👍

boblangley:
👍 Merged #408

PR #409, opened 2016-11-03
Shutdown notifier when all references are gone

danielmarbach:
When RefCount is one we would just decrease the RefCount but not shutdown.

This will make sure we shut down when all references are gone.


danielmarbach:
Since we are currently not stopping concurrently we are fine with the current not thread safe ref counting. Unfortunately I was not able to find any test scenarios with ref counting

SeanFeldman:
Also, [correct count of refference](https://github.com/Particular/NServiceBus.AzureServiceBus/pull/409/commits/2bf1ba2c7f30ad6ded11c7a57bd407bae7b9c805#diff-446ccf5d2a05a3320ebaf7992c901021R22) by counting the first instance as well.

SeanFeldman:
Retargetted to go into `develop`.
Looks good. @Particular/azure-service-bus-maintainers a second pair of eyes and a merge?

yvesgoeleven:
Looks good

PR #442, opened 2016-12-02
ASB 3.4.3 update

SeanFeldman:
Update to the latest MSFT ASB


SeanFeldman:
@Particular/azure-maintainers at your leasure

tmasternak:
adding `WIP` unitil we make the build green

SeanFeldman:
@tmasternak don't think we need a `WIP` as there's absolutely nothing to work on 😄 
Build should be green shortly so should be good to go.

danielmarbach:
@SeanFeldman WIP is a convention most repos use to indicate it cannot be merged. There is even a chrome plugin which greys out the merge buttons

SeanFeldman:
@danielmarbach careful, that plugin not always works. I always saw WIP as a convention to communicate "should not be merged", not "cannot be merged". I.e. the authors of PR indicate that it's still in progress and should not be attempted to be merged. Semantics 😄 

Removing WIP, @Particular/azure-maintainers good to go.

boblangley:
merged #442

PR #454, opened 2016-12-12
Exception message points to UseEndpointOrientedTopology instead of UseTopology<T>

danielmarbach:
Small fix for the exception message on the publisher configuration


boblangley:
Merged #454

Issue #392, opened 2016-10-26
Unsubscribing from an event doesn't work


SeanFeldman:
## Who's affected

- Anyone that requires to be able to unsubscribe endpoints from events

## Symptoms

Unsubscribe requests are not persisted, causing unsubscribed endpoints continue receiving events after those endpoints were restarts.

## Details

Issue raised based on the SO question [here](http://stackoverflow.com/questions/40249886/nservicebus-pub-sub-on-azure-service-bus-queues)

The request to unsubscribe should be modifying topologies. [Currently](https://github.com/Particular/NServiceBus.AzureServiceBus/blob/e5f64e929ecf7ff2227541870ae9a2c243ed9222/src/Transport/Seam/SubscriptionManager.cs#L28) it doesn't.
For the `ForwardingTopology`, subscriber should remove the subscription rule from all the bundles/namespaces.
For the `EndpointOrientedTopology`, subscriber should remove the subscription from the publisher.events topic.

PR #461, opened 2016-12-20
Hotfix 7.1.1

SeanFeldman:
Candidates for this hotfix:
- #462 
- #463

PR #462, opened 2016-12-20
Subscription not removed when unsubscribing with EndpointOrientedTopology

### LINKED ISSUE #459: Unsubscribing from an event doesn't work with EndpointOrientedTopology
## Who's affected

- Anyone that requires to be able to unsubscribe endpoints from events and using `EndpointOrientedTopology`

## Symptoms

Unsubscribe requests are not persisted, causing unsubscribed endpoints continue receiving events after those endpoints were restarts.

## Details

Similar to the issue that was found in `ForwardingTopology` (#392).

The request to unsubscribe should be modifying topologies. [Currently](https://github.com/Particular/NServiceBus.AzureServiceBus/blob/e5f64e929ecf7ff2227541870ae9a2c243ed9222/src/Transport/Seam/SubscriptionManager.cs#L28) it doesn't.

For the `EndpointOrientedTopology`, subscriber should remove the subscription from the publisher.events topic.

This could cause events to accumulate under subscription that is no longer monitored. If there's a high volume of these events, subscription could eventually overflood and cause production degradation.

SeanFeldman:
## Who's affected

- Anyone that requires to be able to unsubscribe endpoints from events and using `EndpointOrientedTopology`

## Symptoms

Unsubscribe requests are not persisted, causing unsubscribed endpoints continue receiving events after those endpoints were restarts.

## Details

Similar to the issue that was found in `ForwardingTopology` (#392).

The request to unsubscribe should be modifying topologies. [Currently](https://github.com/Particular/NServiceBus.AzureServiceBus/blob/e5f64e929ecf7ff2227541870ae9a2c243ed9222/src/Transport/Seam/SubscriptionManager.cs#L28) it doesn't.

For the `EndpointOrientedTopology`, subscriber should remove the subscription from the publisher.events topic.

This could cause events to accumulate under subscription that is no longer monitored. If there's a high volume of these events, subscription could eventually overflood and cause production degradation.

#### TODO
- [ ] backport to `support-7.0`
- [ ] port changes to `develop`

### DIFF
--- a/src/AcceptanceTests/Routing/When_unsubscribing.cs
+++ b/src/AcceptanceTests/Routing/When_unsubscribing.cs
@@ -28,10 +28,10 @@
 
             if (context.IsForwardingTopology)
             {
-                var isSubscriptionFound = await namespaceManager.SubscriptionExistsAsync("bundle-1", $"{endpointName}");
+                var isSubscriptionFound = await namespaceManager.SubscriptionExistsAsync("bundle-1", endpointName);
                 Assert.IsFalse(isSubscriptionFound, "Subscription under 'bundle-1' should have been deleted, but it wasn't.");
 
-                isSubscriptionFound = await namespaceManager.SubscriptionExistsAsync("bundle-2", $"{endpointName}");
+                isSubscriptionFound = await namespaceManager.SubscriptionExistsAsync("bundle-2", endpointName);
                 Assert.IsFalse(isSubscriptionFound, "Subscription under 'bundle-2' should have been deleted, but it wasn't.");
             }
             else // EndpointOrientedTopology
--- a/src/AcceptanceTests/Routing/When_unsubscribing.cs
+++ b/src/AcceptanceTests/Routing/When_unsubscribing.cs
@@ -42,7 +42,7 @@
         {
             public Endpoint()
             {
-                EndpointSetup<DefaultServer>()
+                EndpointSetup<DefaultServer>(endpointConfiguration => endpointConfiguration.EnableFeature<DetermineWhatTopologyIsUsed>())
                     .AddMapping<MyEvent>(typeof(Endpoint));
             }
 
--- a/src/AcceptanceTests/Routing/When_unsubscribing.cs
+++ b/src/AcceptanceTests/Routing/When_unsubscribing.cs
@@ -24,10 +24,18 @@
 
             var connectionString = Environment.GetEnvironmentVariable("AzureServiceBusTransport.ConnectionString");
             var namespaceManager = NamespaceManager.CreateFromConnectionString(connectionString);
+            var endpointName = ConfigureEndpointAzureServiceBusTransport.NameForEndpoint<Endpoint>();
 
-            if (!context.IsForwardingTopology)
+            if (context.IsForwardingTopology)
             {
-                var endpointName = ConfigureEndpointAzureServiceBusTransport.NameForEndpoint<Endpoint>();
+                var isSubscriptionFound = await namespaceManager.SubscriptionExistsAsync("bundle-1", $"{endpointName}");
+                Assert.IsFalse(isSubscriptionFound, "Subscription under 'bundle-1' should have been deleted, but it wasn't.");
+
+                isSubscriptionFound = await namespaceManager.SubscriptionExistsAsync("bundle-2", $"{endpointName}");
+                Assert.IsFalse(isSubscriptionFound, "Subscription under 'bundle-2' should have been deleted, but it wasn't.");
+            }
+            else // EndpointOrientedTopology
+            {
                 var isSubscriptionFound = await namespaceManager.SubscriptionExistsAsync(endpointName + ".events", $"{endpointName}.{nameof(MyEvent)}");
                 Assert.IsFalse(isSubscriptionFound, "Subscription should have been deleted, but it wasn't.");
             }
--- a/src/AcceptanceTests/Routing/When_unsubscribing.cs
+++ b/src/AcceptanceTests/Routing/When_unsubscribing.cs
@@ -1,0 +1,95 @@
+﻿namespace NServiceBus.Azure.Transports.WindowsAzureServiceBus.AcceptanceTests.Routing
+{
+    using System;
+    using System.Threading.Tasks;
+    using AcceptanceTesting;
+    using Features;
+    using Microsoft.ServiceBus;
+    using NUnit.Framework;
+    using Settings;
+    using Transport.AzureServiceBus;
+    using AzureServiceBus;
+    using NServiceBus.AcceptanceTests;
+    using NServiceBus.AcceptanceTests.EndpointTemplates;
+
+    public class When_unsubscribing : NServiceBusAcceptanceTest
+    {
+        [Test]
+        public async Task Should_remove_subscription_from_topology()
+        {
+            var context = await Scenario.Define<Context>()
+                .WithEndpoint<Endpoint>(b => b.When((session, c) => session.Unsubscribe<MyEvent>()))
+                .Done(c => c.EndpointsStarted)
+                .Run();
+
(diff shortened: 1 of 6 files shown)


src/Transport/Creation/AzureServiceBusSubscriptionCreatorV6.cs
```diff
@@ -39,15 +39,15 @@ public AzureServiceBusSubscriptionCreatorV6(ReadOnlySettings settings)
 
             var subscriptionDescription = new SubscriptionDescription(topicPath, subscriptionName);
             
-            // check the if the subscription with event name only is the one we should delete, i.e. event with another namespace owns it
-            if (!await SubscriptionIsReusedAcrossDifferentNamespaces(subscriptionDescription, sqlFilter, namespaceManager).ConfigureAwait(false))
+            // Check subscription with event name only is the one we should delete. If it's reused, then we need to use event full name.
+            if (await SubscriptionIsReusedAcrossDifferentNamespaces(subscriptionDescription, sqlFilter, namespaceManager).ConfigureAwait(false))
```
danielmarbach: Difficult to follow for me why we inverse the bool check

danielmarbach:
According to our support policy I think we need to release a hotfix for 7.0.x as well, right?

src/AcceptanceTests/Routing/When_unsubscribing.cs
```diff
@@ -0,0 +1,95 @@
...
+        {
+            var context = await Scenario.Define<Context>()
+                .WithEndpoint<Endpoint>(b => b.When((session, c) => session.Unsubscribe<MyEvent>()))
+                .Done(c => c.EndpointsStarted)
+                .Run();
+
+            var connectionString = Environment.GetEnvironmentVariable("AzureServiceBusTransport.ConnectionString");
+            var namespaceManager = NamespaceManager.CreateFromConnectionString(connectionString);
+
+            if (!context.IsForwardingTopology)
```
SeanFeldman: Yeah, I realized that I didn't have anything after pushing, but by then needed to wrap up. Will add it second case as well, because it's actually more sophisticated (has 2 cases on its own).

src/Transport/Creation/AzureServiceBusSubscriptionCreatorV6.cs
SeanFeldman: If the was the same event from a different namespace, then the subscription entity we're interested in is using a full type name, rather than just a name. This was a bug. We did the inverse.

SeanFeldman:
Correct. I've updated the general description to have that as an actionable item. `hotfix-7.0.5` branch is in place now and based on the outcomes of #461 will update it.

SeanFeldman:
Update: both topologies are tested now. For `ForwardingTopology` all permutations are validated.

@Particular/azure-maintainers this is ready for final review.

SeanFeldman:
@danielmarbach tried to avoid test execution for the 2nd topology by skipping it in the transport configuration, but ATTs doesn't allow it (throws).

danielmarbach:
Hmm... that worked for me. Care to show me the code u used?

SeanFeldman:
Haven't committed as it wasn't working, but it was the following code in `ConfigureEndpointAzureServiceBusTransport.cs`

```c#
if (endpointName == NameForEndpoint<When_unsubscribing_from_one_of_the_events_for_ForwardingTopology.Endpoint>())
{
  Assert.Ignore("not applicable ...");
}
```

src/AcceptanceTests/Routing/When_unsubscribing_from_one_of_the_events_for_ForwardingTopology.cs
```diff
@@ -0,0 +1,138 @@
...
+            if (context.IsForwardingTopology)
+            {
+                var isSubscriptionFound = await namespaceManager.SubscriptionExistsAsync("bundle-1", endpointName);
+                Assert.IsTrue(isSubscriptionFound, "Subscription under 'bundle-1' should have been found, but it wasn't.");
+
+                isSubscriptionFound = await namespaceManager.SubscriptionExistsAsync("bundle-2", endpointName);
+                Assert.IsTrue(isSubscriptionFound, "Subscription under 'bundle-2' should have been found, but it wasn't.");
+
+                var rules = await namespaceManager.GetRulesAsync("bundle-1", endpointName);
+                Assert.That(rules.Any(rule => rule.Name == ruleName), Is.False);
```
Scooletz: I think I'd prefer to assert the content of the whole collection. Ensuring that one item does not exist seems to week to me.

src/AcceptanceTests/Routing/When_unsubscribing_from_one_of_the_events_for_ForwardingTopology.cs
```diff
@@ -0,0 +1,138 @@
...
+
+                rules = await namespaceManager.GetRulesAsync("bundle-2", endpointName);
+                Assert.That(rules.Any(rule => rule.Name == ruleName), Is.False);
```
Scooletz: Assert whole collection.

Scooletz:
Beside assertion comment, LGTM.

src/AcceptanceTests/Routing/When_unsubscribing_from_one_of_the_events_for_ForwardingTopology.cs
SeanFeldman: We only care for a specific event to be unsubscribed, therefore that one should not be found. The rest of the events - we shouldn't care about those.

src/AcceptanceTests/Routing/When_unsubscribing_from_one_of_the_events_for_ForwardingTopology.cs
SeanFeldman: Same as [here](https://github.com/Particular/NServiceBus.AzureServiceBus/pull/462#discussion_r93771549)

src/AcceptanceTests/Routing/When_unsubscribing_from_one_of_the_events_for_ForwardingTopology.cs
Scooletz: I don't agree. We care for an event being unsubscribed AND the other left as it was.

SeanFeldman: Disagree. 
We've got another tests that verifies that a single event subscription will cause removal of the subscription and rule. In this scenario, we only care about specific event subscription. If there are other events (which we shouldn't care about) , then subscription won't be removed. In case subscription is not found (tested here), then we'll fail as should have.
We care for the specific event to be removed and for the subscription to remain if there are other events. We've got another test that verifies

SeanFeldman: I.e. for this scenario it is known that there will be _other subscriptions_, one or many shouldn't matter. What should matter is that 
1. The subscription for the event we've unsubscribed from is not found
1. Subscription entity remains (due to other event subscriptions still remaining)

Since there are only two events subscribed to, one of them no longer exists, but subscription entity is there, backed up with the 2nd test (`when_unsubscribing`) it defacto asserts that the second event subscription was not removed.

Scooletz: It's not about the other test testing a singular subscription case. The current assertion is an assertion against the action result (removal of the subscription). Turning that into the assertion against the state (how it looks after) would make it explicit.
The reasoning that if a subscription is there it must contain the other event is implicitly based on the other test. Making the assertion against the state makes this state explicitly based on the test itself.
As I'll be off, so please don't treat my comment as a blocker.

SeanFeldman:
Assertion modified as per review request.

PR #463, opened 2016-12-20
Events hierarchy from different publishers with EndpointOrientedTopology not handled

### LINKED ISSUE #460: Polymorphic events from different publishers not handled properly with EndpointOrientedTopology
## Who's affected

- Anyone using `EndpointOrientedTopology` and has endpoints that subscribe to polymorphic events published by different endpoints, applying workaround demonstrated in [Polymorphic events with Azure Service Bus Transport](https://docs.particular.net/samples/azure/polymorphic-events-asb/) sample.

## Symptoms

Each publisher should have a subscription for the subscriber, but only one such subscription is created under a single publisher. This could lead to published events disappear and never be handled. While `EndpointOrientedTopology` doesn't support polymorphic events out-of-the-box, with disabled `AutoSubscribe` feature and manual subscription to the base event it should work as expected.

## Details

Verification `When_multi_subscribing_to_a_polymorphic_event` should work for this scenario when `AutoSubscribe` feature is disabled for the `Subscriber` endpoint and registered publishers are using the base event type (`IMyEvent`) for `EndpointOrientedTopology`.

SeanFeldman:
Fixes #460 (Polymorphic events from different publishers not handled properly with `EndpointOrientedTopology`).
Based on the changes made in [`entity-info` branch](https://github.com/Particular/NServiceBus.AzureServiceBus/tree/entity-info).
Details and description are in the original issue.

#### TODO
- [ ] backport to `support-7.0`
- [ ] port changes to `develop`

### DIFF
--- a/src/Transport/Topology/MetaModel/EntityInfo.cs
+++ b/src/Transport/Topology/MetaModel/EntityInfo.cs
@@ -1,7 +1,6 @@
 namespace NServiceBus.Transport.AzureServiceBus
 {
     using System.Collections.Generic;
-    using System.Linq;
 
     [ObsoleteEx(Message = ObsoleteMessages.WillBeInternalized, TreatAsErrorFromVersion = "8.0", RemoveInVersion = "9.0")]
     public class EntityInfo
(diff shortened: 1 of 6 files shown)


src/Transport/Topology/MetaModel/EntityInfo.cs
```diff
@@ -22,6 +23,16 @@ public EntityInfo()
 
         protected bool Equals(EntityInfo other)
         {
+            if (Type == EntityType.Subscription)
```
danielmarbach: I know that @tmasternak @Scooletz and I wrote that code as a quick hack when we debugged the above test. Still on the fence whether we should put that in in this way. Since we have `SubscriptionInfo` should we move and overridden equal into the subscription info?

danielmarbach:
According to our support policy I think we need to release a hotfix for 7.0.x as well, right?

src/Transport/Topology/MetaModel/EntityInfo.cs
SeanFeldman: That and also bringing `EntityInfo` to the level of ASB - I.e. all entities have a path, but not the name. A path for a subscription will be topic path + subscription name. That way, comparing entities` path value will always be correct action.

SeanFeldman:
Pending PR #465

PR #464, opened 2016-12-22
Subscription not removed when unsubscribing with EndpointOrientedTopology

SeanFeldman:
## Who's affected

- Anyone that requires to be able to unsubscribe endpoints from events and using `EndpointOrientedTopology`

## Symptoms

Unsubscribe requests are not persisted, causing unsubscribed endpoints continue receiving events after those endpoints were restarts.

## Details

Similar to the issue that was found in `ForwardingTopology` (#392).

The request to unsubscribe should be modifying topologies. [Currently](https://github.com/Particular/NServiceBus.AzureServiceBus/blob/e5f64e929ecf7ff2227541870ae9a2c243ed9222/src/Transport/Seam/SubscriptionManager.cs#L28) it doesn't.

For the `EndpointOrientedTopology`, subscriber should remove the subscription from the publisher.events topic.

This could cause events to accumulate under subscription that is no longer monitored. If there's a high volume of these events, subscription could eventually overflood and cause production degradation.

Backport of #459


SeanFeldman:

src/Transport/Connectivity/NamespaceManagerAdapter.cs
```diff
@@ -113,9 +113,15 @@ public Task<TopicDescription> GetTopic(string path)
...
         }
 
+        // TODO: delete this once NamespaceManagerAdapter is internalized
         public Task DeleteSubscriptionAsync(string topicPath, string subscriptionName)
         {
-            return manager.DeleteSubscriptionAsync(topicPath, subscriptionName);
+            return ((INamespaceManagerAbleToDeleteSubscriptions) this).DeleteSubscription(new SubscriptionDescription(topicPath, subscriptionName));
+        }
+
+        Task INamespaceManagerAbleToDeleteSubscriptions.DeleteSubscription(SubscriptionDescription subscriptionDescription)
```
danielmarbach: Note to reviewers: This is a trick to not expose the internal API method as a public method

src/Transport/Creation/AzureServiceBusForwardingSubscriptionCreator.cs
```diff
@@ -2,12 +2,13 @@
...
     using System;
     using System.Collections.Concurrent;
+    using System.Linq;
     using System.Threading.Tasks;
     using Microsoft.ServiceBus.Messaging;
     using Logging;
     using Settings;
 
-    class AzureServiceBusForwardingSubscriptionCreator : ICreateAzureServiceBusSubscriptions
+    class AzureServiceBusForwardingSubscriptionCreator : ICreateAzureServiceBusSubscriptions, ICreateAzureServiceBusSubscriptionsAbleToDeleteSubscriptions
```
danielmarbach: Note to reviewers: Since this class is internal the trick with explicit interface implementation is not needed

src/Transport/Creation/AzureServiceBusSubscriptionCreator.cs
```diff
@@ -7,7 +7,7 @@
     using Logging;
     using Settings;
 
-    class AzureServiceBusSubscriptionCreator : ICreateAzureServiceBusSubscriptions
+    class AzureServiceBusSubscriptionCreator : ICreateAzureServiceBusSubscriptions, ICreateAzureServiceBusSubscriptionsAbleToDeleteSubscriptions
```
danielmarbach: Same comment as https://github.com/Particular/NServiceBus.AzureServiceBus/pull/464/files#r93740028

src/Transport/Creation/AzureServiceBusSubscriptionCreatorV6.cs
```diff
@@ -6,7 +6,7 @@
     using Logging;
     using Settings;
 
-    class AzureServiceBusSubscriptionCreatorV6 : ICreateAzureServiceBusSubscriptions
+    class AzureServiceBusSubscriptionCreatorV6 : ICreateAzureServiceBusSubscriptions, ICreateAzureServiceBusSubscriptionsAbleToDeleteSubscriptions
```
danielmarbach: See https://github.com/Particular/NServiceBus.AzureServiceBus/pull/464/files#r93740028

src/Transport/Topology/TopologyCreator.cs
```diff
@@ -3,7 +3,7 @@ namespace NServiceBus.Transport.AzureServiceBus
     using System.Linq;
     using System.Threading.Tasks;
 
-    class TopologyCreator : ICreateTopology
+    class TopologyCreator : ICreateTopology, ITearDownTopology
```
danielmarbach: See https://github.com/Particular/NServiceBus.AzureServiceBus/pull/464/files#r93740028

src/AcceptanceTests/Routing/When_unsubscribing_from_one_of_the_events_for_ForwardingTopology.cs
```diff
@@ -0,0 +1,138 @@
...
+            if (context.IsForwardingTopology)
+            {
+                var isSubscriptionFound = await namespaceManager.SubscriptionExistsAsync("bundle-1", endpointName);
+                Assert.IsTrue(isSubscriptionFound, "Subscription under 'bundle-1' should have been found, but it wasn't.");
+
+                isSubscriptionFound = await namespaceManager.SubscriptionExistsAsync("bundle-2", endpointName);
+                Assert.IsTrue(isSubscriptionFound, "Subscription under 'bundle-2' should have been found, but it wasn't.");
+
+                var rules = await namespaceManager.GetRulesAsync("bundle-1", endpointName);
+                Assert.That(rules.Any(rule => rule.Name == ruleName), Is.False);
```
Scooletz: As in https://github.com/Particular/NServiceBus.AzureServiceBus/pull/462/files#r93742195 I'd prefer to assert whole collection.

src/AcceptanceTests/Routing/When_unsubscribing_from_one_of_the_events_for_ForwardingTopology.cs
```diff
@@ -0,0 +1,138 @@
...
+
+                rules = await namespaceManager.GetRulesAsync("bundle-2", endpointName);
+                Assert.That(rules.Any(rule => rule.Name == ruleName), Is.False);
```
Scooletz: As in https://github.com/Particular/NServiceBus.AzureServiceBus/pull/462/files#r93742195 I'd prefer to assert whole collection.

danielmarbach:
@yvesgoeleven 

> The async keyword is an implementation detail since it is just syntactic sugar. From an API perspective the following APIs are equivalent:
>
> - `async void` vs. `void`
> - `async Task` vs `Task`
> - `async Task<T>` vs `Task<T>`
> 
> From a behavior perspective you could argue that the runtime behavior from `void` to `async void` can change and therefore it could be a breaking change but from an API perspective both APIs are binary compatible. Since API approver's concern is only the public API the `async` keyword should be omitted. 

https://github.com/JakeGinnivan/ApiApprover/pull/26

src/AcceptanceTests/Routing/When_unsubscribing_from_one_of_the_events_for_ForwardingTopology.cs
SeanFeldman: Same response as in #462

src/AcceptanceTests/Routing/When_unsubscribing_from_one_of_the_events_for_ForwardingTopology.cs
SeanFeldman: Same response as in #462

PR #465, opened 2016-12-23
Keep SubscriptionIfo comparison out of EntityInfo

SeanFeldman:
This is just a small backwards compatible cleanup to perform subscription related comparison in `SubscriptionInfo` and not `EntityInfo`.

There's an issue with our entities abstraction and some features introduced in v7. I'll open a separate issue for that.

### DIFF
--- a/src/Transport/Topology/MetaModel/EntityInfo.cs
+++ b/src/Transport/Topology/MetaModel/EntityInfo.cs
@@ -22,12 +22,7 @@ namespace NServiceBus.Transport.AzureServiceBus
 
         protected bool Equals(EntityInfo other)
         {
-            return string.Equals(Path, other.Path) && Type == other.Type && Equals(Namespace, other.Namespace) && DerivedEqual(other);
-        }
-
-        internal virtual bool DerivedEqual(EntityInfo other)
-        {
-            return true;
+            return string.Equals(Path, other.Path) && Type == other.Type && Equals(Namespace, other.Namespace);
         }
 
         public override bool Equals(object obj)
--- a/src/Transport/Topology/MetaModel/EntityInfo.cs
+++ b/src/Transport/Topology/MetaModel/EntityInfo.cs
@@ -1,7 +1,6 @@
 namespace NServiceBus.Transport.AzureServiceBus
 {
     using System.Collections.Generic;
-    using System.Linq;
 
     [ObsoleteEx(Message = ObsoleteMessages.WillBeInternalized, TreatAsErrorFromVersion = "8.0", RemoveInVersion = "9.0")]
     public class EntityInfo
@@ -23,17 +22,12 @@ namespace NServiceBus.Transport.AzureServiceBus
 
         protected bool Equals(EntityInfo other)
         {
-            if (Type == EntityType.Subscription)
-            {
-                var entity = RelationShips.First(r => r.Type == EntityRelationShipType.Subscription);
-                var otherEntity = other.RelationShips.First(r => r.Type == EntityRelationShipType.Subscription);
-                var targetPathEquals = string.Equals(entity.Target.Path, otherEntity.Target.Path);
-                if (!targetPathEquals)
-                {
-                    return false;
-                }
-            }
-            return string.Equals(Path, other.Path) && Type == other.Type && Equals(Namespace, other.Namespace);
+            return string.Equals(Path, other.Path) && Type == other.Type && Equals(Namespace, other.Namespace) && DerivedEqual(other);
+        }
+
+        internal virtual bool DerivedEqual(EntityInfo other)
+        {
+            return true;
         }
 
         public override bool Equals(object obj)
--- a/src/Transport/Topology/MetaModel/SubscriptionInfo.cs
+++ b/src/Transport/Topology/MetaModel/SubscriptionInfo.cs
@@ -1,5 +1,7 @@
 ﻿namespace NServiceBus.Transport.AzureServiceBus
 {
+    using System.Linq;
+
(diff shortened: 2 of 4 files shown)


SeanFeldman:
Once this one is reviewed and merged, #463 can drop `wip` and be reviewed as well.

src/Transport/Topology/MetaModel/EntityInfo.cs
```diff
@@ -23,17 +22,12 @@ public EntityInfo()
...
-                var entity = RelationShips.First(r => r.Type == EntityRelationShipType.Subscription);
-                var otherEntity = other.RelationShips.First(r => r.Type == EntityRelationShipType.Subscription);
-                var targetPathEquals = string.Equals(entity.Target.Path, otherEntity.Target.Path);
-                if (!targetPathEquals)
-                {
-                    return false;
-                }
-            }
-            return string.Equals(Path, other.Path) && Type == other.Type && Equals(Namespace, other.Namespace);
+            return string.Equals(Path, other.Path) && Type == other.Type && Equals(Namespace, other.Namespace) && DerivedEqual(other);
```
danielmarbach: Wouldn't the normal pattern be to just override bool Equals(object) in the base classes instead of introducing another virtual member?

SeanFeldman: Yes it would. Reason I didn't do that is because we'd end up with two new method signatures  detected by API comparison - `Equals(object)` and `GetHashCode()` required when overriding equality method. That means we're breaking binary compatibility, isn't it @danielmarbach?

tmasternak: @SeanFeldmanwe could go with virtual members in 8.0 when we obsolete both `EntityInfo` and `SubscriptionInfo`, right?

SeanFeldman: @tmasternak why would be obsolete `SubscriptionInfo`? Sorry, not following.

tmasternak: https://github.com/Particular/NServiceBus.AzureServiceBus/pull/465/files#diff-05c4f32fb58222014c81775f7448984cR5 ?

SeanFeldman: Funny. Obsolete from public API because it will be _internalized_ 😄

SeanFeldman: @Particular/azure-maintainers can we please make a decision on this?
This is what API approval yields if `Equals(object)` and `GetHashCode()` in `SubscriptionInfo` is overridden:
image

So this could be a bug in APIApproval, not sure. Even if it's not an issue, overriding `GetHashCode()` for `SubscriptoinInfo` is now changing the method`s behavior as it's no longer going to be the one of the base class.

Scooletz: I'm ok with this change as it does not break the backward compatibility. 

_Reasoning:_
If somebody used the `SubscriptionInfo.Equals` before, it wasn't working properly. This changes the behavior making is proper. The change is not breaking until somebody emitted a call directly against `EntityInfo.Equals` which is highly unlikely. Having this said, we're not breaking anything.

SeanFeldman:
During today's maintainers call we've agreed to move forward with this PR and block at the hotfix level (for a final review). In case there's a strong opposition hotfix will be modified. Otherwise, released.

src/Transport/Topology/MetaModel/EntityInfo.cs
boblangley: Based on my conversation with @bording today, I have no objections to switching to an override of `Equals` (and `GetHashCode`) instead.

It seems the concern is API compatibility, Binary compatibility, and behavioral compatibility:

**API compatibility** is met by implementing `Equals` since a NuGet update and compile will still work without any user code changes needed..
**Binary compatibility** is met because the virtual methods are late bound.
**Behavioral compatibility** is met because `Equals` is currently a value comparison and this does not change that. It only corrects the value comparison to be correct for `SubscriptionInfo`. Implementing `GetHashCode` does not break our behavioral compatibility as `GetHashCode` [does not have behavioral guarantees](https://msdn.microsoft.com/en-us/library/system.object.gethashcode(v=vs.110).aspx#Anchor_1).

SeanFeldman: I.e. we're safe to override `Equals()` and `GetHashCode()` and ignore what APIApproval complains about?

boblangley: In my opinion, yes.

SeanFeldman: 👍 pushed updates

SeanFeldman:
@Particular/azure-maintainers code update to use `GetHashCode` and `Equals`. Please review and merge if no issues are found.

src/Transport/Topology/MetaModel/SubscriptionInfo.cs
```diff
@@ -9,5 +11,35 @@ public class SubscriptionInfo : EntityInfo
         public IClientSideSubscriptionFilter ClientSideFilter { get; set; }
 
         public SubscriptionMetadata Metadata { get; set; }
+
+        public override bool Equals(object obj)
+        {
+            if (!base.Equals(obj))
+                return false;
```
yvesgoeleven: I don't think this is right, the entire point of overriding equals is to check equality of different instances?

yvesgoeleven: Discussed 👍

Issue #466, opened 2016-12-24
Maximum concurrency message needs improvement


HaripraghashSubramaniam:
## Who's affected

This is a logging improvement and no one is identified as "affected".

## Symptoms

When `MessageReceiverNotifier` maximum concurrency is adjusted, the log message doesn't specify what entity is going to be affected.

## Original issue

NServiceBus.Transport.AzureServiceBus.MessageReceiverNotifier The maximum concurrency on this message receiver instance has been adjusted to '1', because the total maximum concurrency '1' wasn't divisable by the number of clients '4'

This needs to include for which entitty and the path it was not able to set the maximum concurrency. Without this I was not able to figure out for with entity it was not being set.

SeanFeldman:
@HaripraghashSubramaniam thank you for the suggestion. Definitely [logged message](https://github.com/Particular/NServiceBus.AzureServiceBus/blob/160c369e9cedd9beab531ec47d8ada7da8985833/src/Transport/Receiving/MessageReceiverNotifier.cs#L54) could provide entity`s [full path](https://github.com/Particular/NServiceBus.AzureServiceBus/blob/160c369e9cedd9beab531ec47d8ada7da8985833/src/Transport/Receiving/MessageReceiverNotifier.cs#L38).

PR #470, opened 2017-01-04
Improved maximum concurrency message

SeanFeldman:
As pointed out in #466, the message is not descriptive enough. Given that there can be more than 1 message notifier, where one of them would be `.retries` queue where concurrency doesn't matter, it's hard to know for what entity concurrency is being adjusted. This change will remove the confusion.


SeanFeldman:

Scooletz:
👍 LGTM

PR #449, opened 2016-12-08
[WIP] Update core

timbussmann:
- [x] Raise a core issue for the acceptance test fixes @danielmarbach (https://github.com/Particular/NServiceBus/pull/4432)

PR #479, opened 2017-01-20
[SPIKE] Test core 6.1 with custom publisher registrations

SeanFeldman:
**DO NOT MERGE**

Investigating if changes to the core ATTs are causing execution to hang.

Excluded commits:
* https://github.com/Particular/NServiceBus.AzureServiceBus/pull/449/commits/36327f86b6f2e8915b275e86b371916a3acf79e0
* https://github.com/Particular/NServiceBus.AzureServiceBus/pull/449/commits/cc78b8e4db2771ebd4d8bab7db54991b64d7234b


SeanFeldman:
The build is hanging, now with both topologies (for PR #449, only `ForwardingTopology` is hanging).

image

SeanFeldman:
Spent some time with @danielmarbach and we've found the issue, which is most probably affecting ASQ as well. The issue is related to the ATTs FW. More specifically, Whens excecution that is going in a tight while loop, causing some thread starvation to the underlying ASB/ASQ libabry in environments that are resource poor. We've tried to introduce a delay or force task exit with Task.Yield and that fixed the hanging.

Closing this PR as we don't need it anymore.

PR #484, opened 2017-01-31
[SPIKE] Bringing in tests

danielmarbach:
Connects to #449 

**Base line** for this spike is `master` that is green on build server.

**Change introduced**: single ATT test `When_unsubscribing_from_event` which is causing failures.

### DIFF
--- a/src/AcceptanceTests/App_Packages/NSB.AcceptanceTests.6.0.0/Routing/MessageDrivenSubscriptions/When_unsubscribing_from_event.cs
+++ b/src/AcceptanceTests/App_Packages/NSB.AcceptanceTests.6.0.0/Routing/MessageDrivenSubscriptions/When_unsubscribing_from_event.cs
@@ -1,0 +1,137 @@
+﻿namespace NServiceBus.AcceptanceTests.Routing.MessageDrivenSubscriptions
+{
+    using System.Threading.Tasks;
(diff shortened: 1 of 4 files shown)


SeanFeldman:
Updated general description to have a clue what's going on.

SeanFeldman:
I've tried running the test in sequence (10 times).
image

At certain execution test runner completely stalls. Not only that, we no longer receive messages. I.e. messages are not picked up from the queue they are forwarded to:

image

As far as I know, test run should be interrupted if we exceed 2 mins, but the last execution is hanging.

SeanFeldman:
I've tried to limit subscribers concurrency to 1 as per suggestion from @timbussmann, but that didn't help a lot. Both topologies started failing the test after running for more than once and hanging the build. 
@Particular/azure-maintainers let's sync on this to decide what to do.

SeanFeldman:
Rebased on the latest master and `When_unsubscribing_from_event` is now failing on the `EndpointOrientedTopology` while works on the `ForwardingTopology`. Can't really draw conclusions out of that.

danielmarbach:
The issue seems to be related to ASB client and the broker. See https://github.com/SeanFeldman/ASB.Forwarding.Issue Closing

SeanFeldman:
Superseded by #517

Issue #458, opened 2016-12-19
Topology migration


danielmarbach:
Could we provide a migration tool or code internally in the transport which bridges an endpoint-oriented topology over to a forwarding topology on the fly including guidance when the old topology can be removed?

Connects to https://github.com/Particular/NServiceBus.Azure/issues/310

SeanFeldman:
This is something that has been briefly discussed in the past and the biggest concern was always the fact that taking offline a system to perform an complete upgrade is not an option. Worth discussing to see if we could bypass that w/o having to require system to go down, _if_ possible. If not, at least we'll be able to document the reasons.

yvesgoeleven:
We can't do it automatically without causing downtime (main limitation is that it's impossible to both receive and forward from an entity at the same time), but we could document how to do it step by step without causing downtime.

SeanFeldman:
Note: there will be basic automatic interoperabitlity between Azure Service Bus (ASB) Forwarding topology and Azure Service Bus .NET Standard (ASBS, ASB transport for .NET Core).

PR #485, opened 2017-01-31
TopologyOperator per pump unless it is the main queue

danielmarbach:
Porting `hotfix-7.1.2` into `develop`

Relying on conventions for satellites is not working properly since users can add their custom satellites. This PR contains:

- Remove the need for satellite address collection
- Rely on shared queue (local address). If a pump belongs to the shared queue then a singleton topology operator is used which is shared with the subscription manager
- The singleton topology operator is necessary because we need to share the main pump delegates with message receive notifiers that are created on demand when the subscription manager is called


danielmarbach:
@yvesgoeleven port of the #483 fixes to develop

SeanFeldman:
Please provide general description.

src/Transport/Seam/MessagePump.cs
```diff
@@ -98,7 +97,7 @@ public void Start(PushRuntimeSettings limitations)
         /// </summary>
         IOperateTopologyInternal DetermineTopologyOperator(string pushSettingsInputQueue)
         {
-            if (satelliteTransportAddresses.Contains(pushSettingsInputQueue))
+            if (pushSettingsInputQueue != localAddress)
```
Scooletz: Could this change be described as:

> instead of searching for the address in a set of satellites, now it checks is it different from the local address. If yes, it creates a new operator. The condition is more inclusive, works for all the previous cases and potentially covers some new cases?

danielmarbach: @Scooletz correct

danielmarbach:
@SeanFeldman done, better?

SeanFeldman:
@Particular/azure-maintainers when we port hotfixes back into develop, please indicate that with a reference to the hotfix. Thanks 😃

boblangley:
Merged #458

PR #492, opened 2017-02-03
Auto reset event leak

### LINKED ISSUE #490: Potential memory leak?
@SeanFeldman I have uploaded a repro. Uses azure sdk 2.9.5 with vs2017RC ASB 7.1.1 and NSB 6.1.2. I have removed the ASB conn string and licence. Run the publisher solution to start publishing messages and  WorkerRole1 to receive them. Run it atleast for 15 mins to see the private bytes consumption increasing.

Publisher - https://drive.google.com/open?id=0B4SLN0ruS-dvVEJPSHByMmtnM0k
Subscriber - https://drive.google.com/open?id=0B4SLN0ruS-dvV3VVSjdsLUUxMFE

### LINKED ISSUE #493: Memory leak due to TaskCompletionSource never completing when push interval is reached before batchSize
Related to #490 

## Who's affected
* All customers with throughput lower than 1000 msg per second

## Symptoms

The multi producer concurrent completion (MPCC) structure uses internally a `AsyncAutoResetEvent`. Whenever the batch Size is reached it sets that `AsyncAutoResetEvent`. Under low throughput the push interval is always reached before the batch size interval. In those cases tasks will be leaking.

This is cause due to a wrong implementation of the `AsyncAutoResetEvent `in combination with a misuse in the multi producer concurrent consumer. 

## Solution

The implementation of MPCC can simply use a `TaskCompletionSource `to signal that the batch size is reached

danielmarbach:
Closes #490 and fixes #493 

More details can be found in #493


danielmarbach:

Carefully review the fact that the `batchSizeReached` TaskCompletionSource is potentially accessed by multiple threads inside the `Push` method and the field is atomically assigned inside the timer loop by one thread only. Does it require any additional synchronization (I and @tmasternak don't think so)?

danielmarbach:
@tmasternak I made a comment because I realized the cancellation token registration is not necessary.

Issue #508, opened 2017-03-10
ASB Transport may attempt to update immutable properties of existing queues / topics


onesadjam:
BUG: Transport will attempt to update immutable properties of entity, resulting in exception.

Encountered in version 7.1.4, but would appear in previous releases as well.

If a queue or topic exists and the requested entity definition is different from the existing definition, the transport will attempt to update the entity.  Some properties are immutable:

- RequiresDuplicateDetection
- EnablePartitioning
- RequiresSession

If the difference between the requested definition and existing definition is limited to these properties, the entity will not be updated.  However, if the difference includes these as well as other, mutable properties, the update attempt will proceed.  This will result in a 400 response from ASB.  For instance, if a queue exists with a maximum size of 1MB and is unpartitioned, and a request is made to define the queue as 5MB and partitioned, the update attempt will proceed because the change in size is allowed.

The resolution is to only update those mutable properties.  Immutable properties should be read from the existing entity and applied to the update request.

SeanFeldman:
Thank you @onesadjam.
Will look into this and PR #509

SeanFeldman:
Added a fix for `RequiresSession` in PR #511

SeanFeldman:
@onesadjam hotfixes are staged and will be released shortly.

SeanFeldman:
Closing this issue. For the relevant hotfixes, subscribe to #512 or #513

SeanFeldman:
@onesadjam 7.1.5 release has [gone out](https://www.nuget.org/packages/NServiceBus.Azure.Transports.WindowsAzureServiceBus/7.1.5).

onesadjam:
@SeanFeldman Fantastic, thanks for your help!

PR #509, opened 2017-03-10
Safe entity update (#508)

onesadjam:
Connect to https://github.com/Particular/NServiceBus.AzureServiceBus/issues/508

On startup the transport will attempt to create the entity (queue or topic) if it does not exist.
Some entity properties may only be set at creation time and may not be modified later (immutable).
This PR corrects an issue where the endpoint configuration might describe an entity in such a way that it is different from the existing entity, and includes changes to both mutable and immutable properties.  The previous behavior would attempt to modify all properties to match the endpoint configuration, which would be rejected by ASB with a 400 error.  The new behavior is to copy the value of existing, immutable properties from the existing entity into the requested entity for update.  This will eliminate the exception.

### DIFF
--- a/src/Tests/Creation/When_creating_queues.cs
+++ b/src/Tests/Creation/When_creating_queues.cs
@@ -482,24 +482,40 @@ namespace NServiceBus.Azure.WindowsAzureServiceBus.Tests.Creation
             await namespaceManager.CreateQueue(new QueueDescription("existingqueue")
             {
                 LockDuration = TimeSpan.FromSeconds(50),
+                MaxSizeInMegabytes = SizeInMegabytes.Size2048,
                 RequiresDuplicateDetection = true,
                 EnablePartitioning = true,
                 RequiresSession = true
             });
+
+            var queueDescription = await namespaceManager.GetQueue("existingqueue");
+
+            // partitioned topics will have a size that is 16x the requested max
+            Assert.AreEqual(2048 * 16, queueDescription.MaxSizeInMegabytes);
+            Assert.AreEqual(TimeSpan.FromSeconds(50), queueDescription.LockDuration);
+            Assert.IsTrue(queueDescription.EnablePartitioning);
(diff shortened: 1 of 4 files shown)


onesadjam:
I see that the acceptance tests are failing on the build server, but I am not able to login to the build server to see the specific acceptance test failure.  From my own machine, I do get failures on some acceptance tests, such as Should_remove_subscription_from_topology. However, if I revert to the 'develop' branch prior to my changes, these test continue to fail.  It is very likely that my environment is not configured properly, or that I may need to change the properties of the ASB namespace in my subscription.  This is my first attempt at providing a PR, so any guidance is appreciated.

SeanFeldman:
@onesadjam on the build server it's the unit tests that are failing, as a result of PR changing our assumptions. Probably need to modify `When_creating_queues.Should_be_able_to_update_an_existing_queue_with_new_property_values_without_failing_on_readonly_properties` and `When_creating_topics.Should_be_able_to_update_an_existing_topic_with_new_property_values_without_failing_on_readonly_properties` to reflect the changes. 

I'll be looking into this issue today.

SeanFeldman:
@onesadjam if you want to move forward with your PR, the tests should be modified to assert a change and not an exception. For example, for `_queues.Should_be_able_to_update_an_existing_queue_with_new_property_values_without_failing_on_readonly_properties` you could have a different `LockDuration` and assert that the change took place and new value is set instead of [this line](https://github.com/Particular/NServiceBus.AzureServiceBus/blob/2f24714fd27d266a54b37109694c908043eb798c/src/Tests/Creation/When_creating_queues.cs#L436).

onesadjam:
Thanks!  Doing that now.

SeanFeldman:
@onesadjam the test changes look good. One of the acceptance tests under EndpointOrientedTopology has timed out, but I don't think it's related tot he changes. Once the build is green, I'll merge the PR and issue hotfixes. Stay tuned.

onesadjam:
@SeanFeldman Excellent, thank you!

PR #511, opened 2017-03-13
Safe subscriptions update

SeanFeldman:
Connect to https://github.com/Particular/NServiceBus.AzureServiceBus/issues/508

`RequiresSession` is a read-only/immutable property for subscriptions that should not be changed.


SeanFeldman:

PR #518, opened 2017-03-30
[WIP] DO NOT MERGE - Core 6.2 for master branch

SeanFeldman:
**DO NOT MERGE!!!**

This PR is created to see that ATTs for the `master`-based version of the ASB can be upgraded to the Core 6.2. Should be included with the next hotfix/support branch release.

A similar PR should be created against the `develop` branch.

PR #519, opened 2017-04-03
Core 6.2 for develop branch

SeanFeldman:
Cherry picked commits from the branch to `master` PR #518.


SeanFeldman:
Note: there are 
- 2 flaky deferral tests 
- Pub/sub test that is explicit due to [issue with ASB](https://github.com/SeanFeldman/ASB.Forwarding.Issue)

SeanFeldman:

src/AcceptanceTests/TestSuiteConstraints.cs
```diff
@@ -0,0 +1,15 @@
+﻿namespace NServiceBus.AcceptanceTests
+{
+    using AcceptanceTesting.Support;
+
+    public partial class TestSuiteConstraints
+    {
+        public bool SupportsDtc { get; } = false;
+        public bool SupportsCrossQueueTransactions { get; } = true; // TODO: verify that this is an AtomicSendWithReceive. If not, set to false
```
danielmarbach: @SeanFeldman this is the only test that uses it https://github.com/Particular/NServiceBus/blob/f00c76bc26e70d6e5188473f79e38e51d7445927/src/NServiceBus.AcceptanceTests/Tx/When_receiving_with_native_multi_queue_transaction.cs#L11 

It is literally can a transaction spawn multiple queues

https://github.com/Particular/NServiceBus/blob/f00c76bc26e70d6e5188473f79e38e51d7445927/src/NServiceBus.AcceptanceTests/Tx/When_receiving_with_native_multi_queue_transaction.cs#L39

src/AcceptanceTests/App_Packages/NSB.AcceptanceTests.6.2.0/DelayedDelivery/When_Deferring_a_message.cs
```diff
@@ -29,7 +29,7 @@ public class When_deferring_a_message : NServiceBusAcceptanceTest
                 .Done(c => c.WasCalled)
                 .Run();
 
-            Assert.GreaterOrEqual(context.ReceivedAt - context.SentAt, delay);
+            Assert.That(context.ReceivedAt - context.SentAt, Is.EqualTo(delay).Within(1).Seconds);
```
danielmarbach: @Particular/nservicebus-maintainers please see this. The problem is that ASB has its own time on the broker. When clock skew occurs between the client and the broker then this test will fail. Unfortunately it is not even guaranteed that one second is enough. Should we come up with better tests?

timbussmann: > Should we come up with better tests?

yes, definitely. But so far we haven't found a good approach regarding all delayed delivery related tests, as they are very hard to verify without delaying the test runs to an eternity :(

danielmarbach: Wouldn't it be good enough to have two tests where we do the following:

- One test sets the delay to a fraction of the test run times and asserts that the message was received within the test runtime
- On test sets the delay to a multiple of the test run time and verifies that the message was not received within the test runtime

But not asserting the actual receivedAt or only that the timespan falls withing the test runtime which is actually automatically given when we received the message?

src/AcceptanceTests/TestSuiteConstraints.cs
yvesgoeleven: @danielmarbach so this flag means, cross queue tx support in receive only mode? we should set it to false then. And I wonder why the related test is green then... asb can't do this, it needs atomic sends with receive mode

timbussmann: > verify that this is an AtomicSendWithReceive

yes it is.

danielmarbach: @timbussmann this is misleading then. Look at the links above.  The flag is only used in `When_receiving_with_native_multi_queue_transaction` which specifies `TransportTransactionMode.ReceiveOnly`

yvesgoeleven: @danielmarbach just diagnosed why this test is green, and seen that we have a bug here, we are sending via even if we should not

yvesgoeleven: Follow up, this test is green because core does not invoke the dispatcher for the send right before the exception. We are not effectively sending via (we are logging that we do though, I'll fix the log separatly)

SeanFeldman: So does this flag represent AtomicSendWithReceive or not? I'm confused.

danielmarbach: @SeanFeldman https://github.com/Particular/NServiceBus/pull/4595 it is SendsWithAtomicReceive

src/AcceptanceTests/App_Packages/NSB.AcceptanceTests.6.2.0/DelayedDelivery/When_Deferring_a_message.cs
danielmarbach: @Particular/nservicebus-maintainers what is your thought on the above statement?

timbussmann: > One test sets the delay to a fraction of the test run times and asserts that the message was received within the test runtime

this of course depends on the delayed delivery mechanism/implementation. But too short delayed delivers are not going through the timeout storage but are short circuited. This might be just a problem with the timeoutmanager though.

> On test sets the delay to a multiple of the test run time and verifies that the message was not received within the test runtime

this would make a very long running test as you would have to account for enough time in the test run so that endpoints are able to spin up before the timeout is reached. Such long running tests is something we wanted to avoid.

@danielmarbach how about raising an issue to discuss this problem in the core?

danielmarbach: Raised https://github.com/Particular/NServiceBus/issues/4601

Issue #524, opened 2017-04-18
The bundles feature for the ForwardingTopology can be missconfigured on different endpoints causing message loss


SeanFeldman:
## Who's affected

You are affected if you are using ForwardingTopology and custom number of topics in bundle using [`NumberOfEntitiesInBundle(int)` API](https://docs.particular.net/nservicebus/azure-service-bus/configuration/full#configuring-the-topology-forwarding-topology).

## Symptoms

Misconfigured publisher and subscriber cause events to be posted to the topics that are not listened to, causing events to be lost.

## More details

As part of the conversation in #521 it was identified that the scenario for a message loss could exist when `ForwardingTopology` is used and number of topics in bundles are miss-configured on different endpoints (see [this visualization](https://github.com/Particular/NServiceBus.AzureServiceBus/issues/521#issuecomment-294614402)).

We could have a startup check to verify that number of topics per bundle is aligned with the number of actually found topics. If that's not the case, an exception should be thrown preventing endpoint from starting until bundles configuration is resolved.

## Affected versions
- [x] `support-7.2` --> PR #565 
- [ ] `develop` --> PR #570 
- [x] Announcement --> PR https://github.com/Particular/Announcements/pull/149

SeanFeldman:
At maintainers sync, the following was proposed as a part of potential hotfix:

1. Subscriber(s) to scan all the topics in the bundle and subscriber to all (forwarding, cheap)
1. Obsolete Bundle API with proper messages

yvesgoeleven:
Note: The above is not completely fool proof. If publisher reconfigures the number of bundles without upgrading then subscriber may still miss messages.

SeanFeldman:
@yvesgoeleven have a look at #3 in the https://github.com/Particular/NServiceBus.AzureServiceBus/issues/524#issuecomment-295322130 - it will address the edge case you referred to.

yvesgoeleven:
@SeanFeldman good idea, allows to drain the second topic before we stop using it

yvesgoeleven:
created a branch for this issue https://github.com/Particular/NServiceBus.AzureServiceBus/issues/524

yvesgoeleven:
> Subscriber(s) to scan all the topics in the bundle and subscriber to all

This requires the transport to have manage rights, what will we do if we don't have manage rights?

yvesgoeleven:
[...]

PR #529, opened 2017-04-25
[WIP] Hotfix 7.2.1

SeanFeldman:
Connects to #524 
Replaces branch `prevent-bundle-misconfiguration`

### DIFF
--- a/src/Transport/Topology/MetaModel/NamespaceBundleConfigurations.cs
+++ b/src/Transport/Topology/MetaModel/NamespaceBundleConfigurations.cs
@@ -1,0 +1,51 @@
+﻿namespace NServiceBus.AzureServiceBus.Topology.MetaModel
+{
+    using System;
+    using System.Collections;
+    using System.Collections.Generic;
+    using System.Linq;
+
+    class NamespaceBundleConfigurations : IEnumerable<NamespaceBunleInfo>
+    {
+        List<NamespaceBunleInfo> namespaceBundles;
+
+        public NamespaceBundleConfigurations()
+        {
+            namespaceBundles = new List<NamespaceBunleInfo>();
+        }
+
+        public void Add(string namespaceAlias, int numberOfTopicsInBundle)
+        {
+            var bunleInfo = namespaceBundles.SingleOrDefault(x => x.Alias.Equals(namespaceAlias, StringComparison.OrdinalIgnoreCase));
+            if (bunleInfo != null && bunleInfo.NumberOfTopicsInBundle < numberOfTopicsInBundle)
+            {
+                throw new Exception($"Duplicate namespace alias `{namespaceAlias}` was added to the {nameof(NamespaceBundleConfigurations)}. Namespace aliases should be unique.");
+            }
+
+            namespaceBundles.Add(new NamespaceBunleInfo(namespaceAlias, numberOfTopicsInBundle));
+        }
+
+        public int GetNumberOfTopicInBundle(string namespaceAlias)
+        {
+            try
+            {
+                var selected = namespaceBundles.Single(x => x.Alias.Equals(namespaceAlias, StringComparison.OrdinalIgnoreCase));
+                return selected.NumberOfTopicsInBundle;
+            }
+            catch (InvalidOperationException exception)
+            {
+                throw new KeyNotFoundException($"Namespace with alias `{namespaceAlias}` hasn't been registered", exception);
+            }
+        }
+
+        public IEnumerator<NamespaceBunleInfo> GetEnumerator()
+        {
+            return namespaceBundles.GetEnumerator();
+        }
+
+        IEnumerator IEnumerable.GetEnumerator()
+        {
+            return GetEnumerator();
+        }
+    }
+}
--- a/src/Transport/Topology/MetaModel/NamespaceBundleConfigurations.cs
+++ b/src/Transport/Topology/MetaModel/NamespaceBundleConfigurations.cs
@@ -5,40 +5,33 @@
     using System.Collections.Generic;
     using System.Linq;
 
-    class NamespaceBundleConfigurations : IEnumerable<NamespaceBunleInfo>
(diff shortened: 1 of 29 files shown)


src/Transport/Topology/MetaModel/NamespaceBundleConfigurations.cs
```diff
@@ -0,0 +1,44 @@
...
+                throw new Exception($"Duplicate namespace alias `{namespaceAlias}` was added to the {nameof(NamespaceBundleConfigurations)}. Namespace aliases should be unique.");
+            }
+
+            namespaceBundles.Add(new NamespaceBundleInfo(namespaceAlias, numberOfTopicsInBundle));
+        }
+
+        public int GetNumberOfTopicInBundle(string namespaceAlias)
+        {
+            var selected = namespaceBundles.SingleOrDefault(x => x.Alias.Equals(namespaceAlias, StringComparison.OrdinalIgnoreCase));
+            return selected?.NumberOfTopicsInBundle ?? 0;
```
SeanFeldman: It's ok to return 0 because we'll [take the maximum](https://github.com/Particular/NServiceBus.AzureServiceBus/blob/b58165d61aa661e2018c0dd3dbeb31913e775fc9/src/Transport/Topology/Topologies/ForwardingTopologySectionManager.cs#L280) between what is returned from this method and whatever is configured on the endopint.

src/Transport/Topology/PreStartupChecks/NumberOfTopicsInBundleCheck.cs
```diff
@@ -0,0 +1,39 @@
...
+            {
+                var namespaceManager = manageNamespaceManagerLifeCycle.Get(namespaceConfiguration.Alias);
+                var filter = $"startswith(path, '{bundlePrefix}') eq true";
+                var namespaceManagerThatCanQueryAndFilterTopics = namespaceManager as NamespaceManagerAdapter;
+                // if user has provided an implementation of INamespaceManager, skip the checks all together
+                if (namespaceManagerThatCanQueryAndFilterTopics == null)
+                {
+                    return;
+                }
+                var foundTopics = await namespaceManagerThatCanQueryAndFilterTopics.GetTopics(filter).ConfigureAwait(false);
```
SeanFeldman: If we get to this point, we have the rights as otherwise the endpoint would fail already, wouldn't it?
And if this code will throw, it will throw because we don't have the manage rights. Which is what we'd like.

src/Transport/Topology/Topologies/ForwardingTopology.cs
```diff
@@ -111,6 +112,18 @@ public Func<ICreateQueues> GetQueueCreatorFactory()
             return () => container.Resolve<ICreateQueues>();
         }
 
+        internal void FindOutHowManyTopicExistsForBundleToEnsureResourcesAreCreatedProperly()
```
SeanFeldman: Verbose a bit, agree. Truncating.

src/Transport/Topology/Topologies/ForwardingTopologySectionManager.cs
```diff
@@ -102,17 +102,17 @@ public TopologySection DeterminePublishDestination(Type eventType)
...
                 };
             });
         }
 
-        IEnumerable<EntityInfo> SelectSingleRandomTopicFromBundle(List<EntityInfo> entityInfos)
+        IEnumerable<EntityInfo> SelectFirstTopicFromBundle(List<EntityInfo> entityInfos)
         {
-            var index = randomGenerator.Next(0, entityInfos.Count);
+            const int index = 1;
             var selected = entityInfos[index];
```
SeanFeldman: good catch!

src/Transport/Topology/PreStartupChecks/NumberOfTopicsInBundleCheck.cs
yvesgoeleven: No, if this is the current behavior than it needs to be fixed. Endpoint must be able to run without manage rights as long as all entities are there

src/Transport/Topology/MetaModel/NamespaceBundleConfigurations.cs
yvesgoeleven: I disagree, this method should return the best value possible and not assume that it's caller will fix it. 0 is a bad value

SeanFeldman: The caller doesn't "fix it". 

For the called class, `NamespaceBundleConfigurations` to know what the most recent setting value is, I would have to pass in the number of configured topics in bundle. That is not know until settings are locked (yes, I do not want to pass the whole Settings object since we're trying to avoid that). Finding when that happens is far from trivial.

What the method does is absolutely legitimate. If this method cannot find the namespace alias (which is probably odd on it's own since aliases registered are the ones that user has already provided), the value returned is compared with the setting (either 2, which is configured by default, or whatever user specified).

src/Transport/Topology/PreStartupChecks/NumberOfTopicsInBundleCheck.cs
SeanFeldman: Only option would be to swallow exception + log a warning

SeanFeldman: Updated.

SeanFeldman:
@yvesgoeleven changes applied (where were needed)

SeanFeldman:
#532 might need to go out as `7.2.1` first and this will be `7.2.2`

src/Transport/Topology/MetaModel/NamespaceBundleConfigurations.cs
SeanFeldman: Discussed. 👍 to return 1 instead of 0.

PR #533, opened 2017-05-02
[WIP] Hotfix 7.2.2

SeanFeldman:
Do **NOT** merge

Connects to #524 
Replaces #529 

Intended to be deployed after `hotfix-7.2.1`

### DIFF
--- a/src/AcceptanceTests/Issues/Issue_524.cs
+++ b/src/AcceptanceTests/Issues/Issue_524.cs
@@ -29,10 +29,7 @@
                 }))
                 .WithEndpoint<Subscriber>(builder => builder.When((session, ctx) =>
                 {
-                    if (ctx.HasNativePubSubSupport)
-                    {
-                        ctx.SubscribedToEvent = true;
-                    }
+                    ctx.SubscribedToEvent = ctx.HasNativePubSubSupport;
                     return Task.FromResult(0);
                 }))
                 .Done(ctx => ctx.EventWasHandled)
@@ -55,6 +52,7 @@
             {
                 EndpointSetup<DefaultServer>(endpointConfiguration =>
                 {
+                    endpointConfiguration.Recoverability().DisableLegacyRetriesSatellite();
                     var transport = endpointConfiguration.UseTransport<AzureServiceBusTransport>();
 #pragma warning disable 618
                     var topology = transport.UseTopology<ForwardingTopology>();
@@ -70,10 +68,7 @@
             {
                 EndpointSetup<DefaultServer>(endpointConfiguration =>
                 {
-                    var transport = endpointConfiguration.UseTransport<AzureServiceBusTransport>();
-#pragma warning disable 618
-                    transport.UseTopology<ForwardingTopology>();
-#pragma warning restore 618
+                    endpointConfiguration.Recoverability().DisableLegacyRetriesSatellite();
                 });
             }
 
--- a/src/AcceptanceTests/Issues/Issue_524.cs
+++ b/src/AcceptanceTests/Issues/Issue_524.cs
@@ -2,6 +2,8 @@
 {
     using System.Threading.Tasks;
     using AcceptanceTesting;
+    using Microsoft.ServiceBus;
+    using Microsoft.ServiceBus.Messaging;
     using NServiceBus.AcceptanceTests;
     using NServiceBus.AcceptanceTests.EndpointTemplates;
     using NServiceBus.AcceptanceTests.ScenarioDescriptors;
@@ -17,6 +19,8 @@
             {
                 Assert.Inconclusive("The test is designed for ForwardingTopology only.");
             }
+
+            await MimicAnExistingEnvironmentWithAlreadyMisconfiguredBundlesPreCreated();
 
             await Scenario.Define<Context>()
                 .WithEndpoint<Publisher>(builder => builder.When(ctx => ctx.SubscribedToEvent, async session =>
@@ -34,6 +38,24 @@
                 }))
                 .Done(ctx => ctx.EventWasHandled)
                 .Run();
+        }
+
+        static async Task MimicAnExistingEnvironmentWithAlreadyMisconfiguredBundlesPreCreated()
+        {
+            var connectionString = EnvironmentHelper.GetEnvironmentVariable("AzureServiceBusTransport.ConnectionString");
+            var namespaceManager = NamespaceManager.CreateFromConnectionString(connectionString);
+
+            async Task createTopic(string name)
+            {
+                if (! await namespaceManager.TopicExistsAsync(name))
+                {
+                    await namespaceManager.CreateTopicAsync(new TopicDescription(name));
+                }
+            }
+
+            await createTopic("bundle-1");
+            await createTopic("bundle-2");
+            await createTopic("bundle-3");
         }
 
         public class Context : ScenarioContext
--- a/src/AcceptanceTests/Issues/Issue_524.cs
+++ b/src/AcceptanceTests/Issues/Issue_524.cs
@@ -12,7 +12,7 @@
     public class Issue_524 : NServiceBusAcceptanceTest
     {
         [Test]
-        public async Task When_publisher_and_subscriber_bundles_are_miscofigured_Should_not_loose_messages()
+        public async Task When_publisher_and_subscriber_bundles_are_misconfigured_Should_not_lose_messages()
         {
             var topology = EnvironmentHelper.GetEnvironmentVariable("AzureServiceBusTransport.Topology");
             if (topology != "ForwardingTopology")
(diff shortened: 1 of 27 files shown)


SeanFeldman:
@Particular/azure-maintainers this can be reviewed. **Do not** merge it yet please.

src/AcceptanceTests/Publishing/When_using_a_signle_bundle.cs
```diff
@@ -32,7 +32,9 @@ public Publisher()
                 EndpointSetup<DefaultServer>(config =>
                 {
                     var transport = config.UseTransport<AzureServiceBusTransport>();
+#pragma warning disable 618
```
danielmarbach: Would it make sense to assign the topology to var and then only have `NumberOfEntitiesInBundle` in pragma? Since `UseForwardingTopology` is not obsoleted

src/NServiceBus.AzureServiceBus.sln.DotSettings
```diff
@@ -1,6 +1,6 @@
 ﻿<wpf:ResourceDictionary xml:space="preserve" xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml" xmlns:s="clr-namespace:System;assembly=mscorlib" xmlns:ss="urn:shemas-jetbrains-com:settings-storage-xaml" xmlns:wpf="http://schemas.microsoft.com/winfx/2006/xaml/presentation">
 
-	<s:String x:Key="/Default/CodeInspection/CSharpLanguageProject/LanguageLevel/@EntryValue">CSharp60</s:String>
+	<s:String x:Key="/Default/CodeInspection/CSharpLanguageProject/LanguageLevel/@EntryValue">CSharp70</s:String>
```
danielmarbach: Really on the fence why we need to bump to CSharp 7 just to be able to use local functions in an acceptance test and we to this in a patch. What are the risks of this change?

src/AcceptanceTests/Publishing/When_using_a_signle_bundle.cs
```diff
@@ -44,7 +46,9 @@ public Subscriber()
...
```
danielmarbach: Would it make sense to assign the topology to var and then only have `NumberOfEntitiesInBundle` in pragma? Since `UseForwardingTopology` is not obsoleted

src/Tests/Topology/MetaModel/When_namespace_bundle_configurations_is_used.cs
```diff
@@ -0,0 +1,46 @@
...
+        public void SetUp()
+        {
+            namespaceBundleConfigurations = new NamespaceBundleConfigurations();
+        }
+
+        [Test]
+        [TestCase("alias", 3)]
+        [TestCase("ALIAS", 3)]
+        [TestCase("Alias", 3)]
+        public void Should_throw_if_namespace_alias_is_readded_with_small_number_of_topics_in_bundle(string name, int numberOfTopicsInBundle)
```
danielmarbach: `readded`?

src/Tests/Topology/MetaModel/When_namespace_bundle_configurations_is_used.cs
```diff
@@ -0,0 +1,46 @@
...
+        {
+            namespaceBundleConfigurations.Add(name, 2);
+            Assert.Throws<Exception>(() => namespaceBundleConfigurations.Add(name, numberOfTopicsInBundle));
+        }
+
+        [Test]
+        public void Should_get_number_of_bundles_for_a_registered_alias()
+        {
+            namespaceBundleConfigurations.Add("alias", 5);
+            var value = namespaceBundleConfigurations.GetNumberOfTopicInBundle("AlIaS");
```
danielmarbach: So this is saying that casing doesn't matter?

src/Transport/Config/AzureServiceBusTransportInfrastructure.cs
```diff
@@ -55,6 +55,11 @@ public override string ToTransportAddress(LogicalAddress logicalAddress)
 
         public override TransportReceiveInfrastructure ConfigureReceiveInfrastructure()
         {
+            if (topology is ForwardingTopology t)
```
danielmarbach: This seems to me a clear indication of a responsibility problem. 

If I understand correctly the following happens here

1. FindOutHowManyTopicExistsForBundle is called in a factory that is synchronous and only supped to return the receive infrastructure factory.
1. The method is called find but actually does insert the number of topics into the newly introduced `NamespaceBundleConfigurations`
1. The bundle check is called check and is asynchronous but is called synchronous and does not really check but fill the bundle configurations
1. The bundle configurations is then used only by the ForwardingTopologySectionManager in order to fill the topics by reading again from `settings.Get<NamespaceBundleConfigurations>(WellKnownConfigurationKeys.Topology.Bundling.NamespaceBundleConfigurations);`

For me this sounds like an unnecessary complex scattering of responsibilities which arguably belongs into the section manager.

SeanFeldman:
After discussing comments from @danielmarbach there's one item that we need to decide on - the fix won't work if there's no mange rights to the namespace as it requires to scan entities.
Options at this point are:
1. Move with this fix and make it clear that w/o manage rules it's user responsibility to set up topics/subscriptions correctly or else messages could be lost.
1. Deprecate bundle configuration API, keep the default as-is (2 topics) and anyone who was using it to contact us for assistance.

@Particular/azure-maintainers I'm putting this PR on hold until we decide what would be a better option to move with.

SeanFeldman:
There are a few paths:

1. **Installers turned off, manage rights not needed** - we don't care, user responsibility to set up the topology. Runtime exception will take place if topology is not created properly.
1. **Installers turned on, no manage rights** - we'll throw an exception about lack of manage rights at startup.
1. **Installers turned on, manage rights exist** - this PR will handle misconfiguration between publishers and subscribers.

Which would mean that Manage Rights PR #547 would need to go in first.

@danielmarbach could you repeat what we've discussed and why do you think it won't work?

src/AcceptanceTests/Issues/Issue_524.cs
```diff
@@ -0,0 +1,114 @@
...
+            var topology = EnvironmentHelper.GetEnvironmentVariable("AzureServiceBusTransport.Topology");
+            if (topology != "ForwardingTopology")
+            {
+                Assert.Inconclusive("The test is designed for ForwardingTopology only.");
+            }
+
+            await MimicAnExistingEnvironmentWithAlreadyMisconfiguredBundlesPreCreated();
+
+            await Scenario.Define<Context>()
+                .WithEndpoint<Publisher>(builder => builder.When(ctx => ctx.SubscribedToEvent, async session =>
```
SeanFeldman: If it doesn't matter, why the ask? 😛 changed.

src/AcceptanceTests/Issues/Issue_524.cs
```diff
@@ -0,0 +1,114 @@
...
+    using System.Threading.Tasks;
+    using AcceptanceTesting;
+    using Microsoft.ServiceBus;
+    using Microsoft.ServiceBus.Messaging;
+    using NServiceBus.AcceptanceTests;
+    using NServiceBus.AcceptanceTests.EndpointTemplates;
+    using NServiceBus.AcceptanceTests.ScenarioDescriptors;
+    using NUnit.Framework;
+
+    public class Issue_524 : NServiceBusAcceptanceTest
```
SeanFeldman: Done

src/AcceptanceTests/Publishing/When_using_a_signle_bundle.cs
SeanFeldman: 👍

src/Tests/Topology/MetaModel/When_namespace_bundle_configurations_is_used.cs
SeanFeldman: Yes, re-added. I've changed the name.

src/Tests/Topology/MetaModel/When_namespace_bundle_configurations_is_used.cs
SeanFeldman: Yep

src/Transport/Config/AzureServiceBusTransportInfrastructure.cs
SeanFeldman: To address this one, see my comment here https://github.com/Particular/NServiceBus.AzureServiceBus/pull/533#issuecomment-300614334

SeanFeldman:
Closing this PR as it has a dependency on #547 to be completed first. Work will be tracked under `issue-524-micsonfigured-bundles-can-cause-message-loss` branch.

Issue #532, opened 2017-05-01
Duplicates with SendsAtomicWithReceive transport transaction mode


SeanFeldman:
Related to https://github.com/Particular/NServiceBus.AzureServiceBus/issues/530

## Who's affected

Anyone using Azure Service Bus transport with `SendsAtomicWithReceive` transaction guarantee

## Symptoms 

Incoming message is processed, **not** completed (a warning about message lost lock is logged), but the outgoing messages are dispatched.
As a result of that, incoming message is re-processed, causing outgoing messages to be sent again, generating duplicates.

## Analysis

The incoming message completion and outgoing batched operations are wrapped in a transaction. Transaction should be completed when the incoming messages is completed successfully _and_ outgoing messages are dispatched.

When an [incoming message is completed](https://github.com/Particular/NServiceBus.AzureServiceBus/blob/6597458e4050c05d19eabfe482c35f2b112058f7/src/Transport/Receiving/MessageReceiverNotifier.cs#L227), there's a slight chance that the lock will be lost. If that's the case, the wrapping transaction should not be completed. Currently, the transaction is completed, causing outgoing messages to be dispatched irregardless of the incoming message completion status.

## Affected versions

The following [supported versions](https://docs.particular.net/nservicebus/upgrades/all-versions#transport-packages-nservicebus-azure-transports-windowsazureservicebus) are affected

- [x] `7.1`
- [x] `7.2`
- [ ] `develop`

## Original issue

https://github.com/Particular/NServiceBus.AzureServiceBus/issues/530

daerath:
@SeanFeldman Sweet.  Great catch!

SeanFeldman:
fixed via https://github.com/Particular/NServiceBus.AzureServiceBus/commit/609eb483cc2a9f229ccb6fb974e7043ae5a6c55b

Issue #530, opened 2017-04-26
Duplication of Timeout Messages


SeanFeldman:
Original issue raised in GG: https://groups.google.com/forum/#!topic/particularsoftware/gpDXlWqwWMg

Description: 
NSB 6.x / ASB 7.x

> We are upgrading from NSB 5 to 6 and I'm at the part where I'm upgrading code that handles deduplication of timeout messages.  We previously tried Outbox but ran into performance issues due to the volume of traffic (north of 10 million a day).  We removed Outbox and leveraged Azure's deduplication capabilities and a Feature that would format messages to enable deduplication.  That works, but with the upgrade we're looking around at some other options to streamline things.

> We have one service that creates and receives timeouts.  It gets one, kicks off some work, and puts another one on the bus with a future delivery date.  Each "job" which maps to a row in the appropriate saga table has a single timeout, so if we have 200,000 rows in the saga table, then we'd have 200,000 timeouts.  There is a separate problem where we have noticed that Timeouts can be lost (the working theory is that if the Windows Azure Worker Host crashes we will lose timeouts) and I'm also hoping that SendsAtomicWithReceive will prevent this issue, but I haven't tested that part yet.

> With NSB 6 I wanted to try and leverage TransportTransactionMode.SendsAtomicWithReceive to see if that would be an acceptable solution.  My first test of 3.6 million messages resulted in 41 duplicate timeout messages.  That was with azure deduplication turned off on that queue and with our custom feature disabled.

> I was also hoping that Batched Messaging (https://docs.particular.net/nservicebus/upgrades/5to6/messaging) would address this issue.

> So, I guess my question is, should SendsAtomicWithReceive have prevented the creation of duplicate timeouts, or is the only way to truly prevent this through NSB to use Outbox, a custom feature, or to not use timeouts at all for this purpose and instead use a webjob or something to trigger this activity?

>The code to enable SendsAtomicWithReceive is:

> transport.TransportType(Microsoft.ServiceBus.Messaging.TransportType.NetMessaging);
> transport.Transactions(TransportTransactionMode.SendsAtomicWithReceive);

SeanFeldman:
Distilled version to repro the issue:

Endpoint gets 5K messages. For each message, a short (1 minute) timeout is scheduled. When timeouts are processed, a replacement timeout message is issued. Over time, number of timeouts starts to grow (duplicates).

[...]

PR #537, opened 2017-05-03
Hotfix 7.1.6

yvesgoeleven:
Connected to #532 
Backport of `hotfix-7.2.1`
Related to #530 

## Who's affected

Anyone using Azure Service Bus transport with `SendsAtomicWithReceive` transaction guarantee

## Symptoms 

Incoming message is processed, **not** completed (a warning about message lost lock is logged), but the outgoing messages are dispatched.
As a result of that, incoming message is re-processed, causing outgoing messages to be sent again, generating duplicates.


SeanFeldman:
@yvesgoeleven this is green now.
Do **not** squash commit (default for the green button in GH 😃). Instead use "merge commit".

Issue #542, opened 2017-05-05
Message can be duplicated when incoming message processing is aborted


SeanFeldman:
## Who's affected

Anyone using Azure Service Bus transport with `SendsAtomicWithReceive` transaction guarantee and implements a custom behavior `Behavior<ITransportReceiveContext>` to cancel incoming message processing by invoking `context.AbortReceiveOperation();`.

## Symptoms

Custom behavior requests incoming message processing to be aborted.  Incoming message is successfully aborted, but the outgoing messages are still dispatched. 

## Original description

Edge case: when `MessageReceiveNotifier` handles completion, it also checks if [cancellation was requested](https://github.com/Particular/NServiceBus.AzureServiceBus/blob/master/src/Transport/Receiving/MessageReceiverNotifier.cs#L267). If so, it will abandon message in flights and complete the scope, rather than abandon and abort the scope. This will cause outgoing messages to be dispatched.

**TODO**
- [x] Port to `develop` --> PR #554
- Back-port to the [supported versions](https://docs.particular.net/nservicebus/upgrades/supported-versions#transport-packages-nservicebus-azure-transports-windowsazureservicebus)
  - [x] `support-7.2` --> PR #552 
  - [x] `support-7.1` --> PR #553

SeanFeldman:
Closed via https://github.com/Particular/NServiceBus.AzureServiceBus/commit/0e2b2e5ec3ef6ead8c80f8057721d0f634fbdb72

PR #544, opened 2017-05-09
[WIP] Message is completed instead of abandoned when incoming message processing is aborted

SeanFeldman:
Connects to #542
~~Depends on #533~~

This is an edge case when an incoming message is abandoned as a result of cancellation that was triggered via `CancellationToken` passed into Core. Currently, this is only possible using [custom behavior](https://github.com/Particular/NServiceBus.AzureServiceBus/compare/hotfix-7.2.2...issue-542-abandoning-on-shutdown?expand=1#diff-901eac66d7f6d8e53becc18c670d1f92R77) since Core doesn't use this cancellation token. Would still want to make sure we're covered and that messages are not abandoned and scope is completed.

Being an edge case, I'm not sure it's worth back-porting this.

@Particular/azure-maintainers thoughts?

### DIFF
--- a/src/AcceptanceTests/Receiving/When_receive_operation_is_aborted.cs
+++ b/src/AcceptanceTests/Receiving/When_receive_operation_is_aborted.cs
@@ -1,0 +1,115 @@
+namespace NServiceBus.Azure.Transports.WindowsAzureServiceBus.AcceptanceTests.Routing
+{
+    using System;
+    using System.Threading.Tasks;
+    using AcceptanceTesting;
+    using NServiceBus.AcceptanceTests;
+    using NServiceBus.AcceptanceTests.EndpointTemplates;
+    using NUnit.Framework;
+    using AcceptanceTesting.Customization;
+    using Pipeline;
+
+    public class When_receive_operation_is_aborted : NServiceBusAcceptanceTest
+    {
+        [Test]
+        public async Task Should_not_dispatch_outgoing_message()
+        {
+            var delay = Task.Delay(TimeSpan.FromSeconds(30));
+
+            var context = await Scenario.Define<Context>()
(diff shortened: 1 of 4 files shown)


danielmarbach:
Since calling Abort in a custom behavior is something that customer can do and doing it will lead to incorrect business behaviour I think this is worth backporting.

Like I expressed before this should not depend upon 7.2.2 and can go out as a 7.2.3 or if we think this is ready before 7.2.2 we can pull out the changes on the hotfix branch and pull this in instead and release it. I raised a responsibility discussion on hotfix 7.2.2 which needs to be addressed first.

SeanFeldman:
With the complication in PR #533, this could go as `7.2.2` or wait until #533 is addressed.
👍 on back-porting (updated the issue #542)

SeanFeldman:
@Particular/azure-maintainers ready for review, but **do NOT** merge.

src/AcceptanceTests/Receiving/When_receive_operation_is_aborted.cs
```diff
@@ -0,0 +1,115 @@
...
+
+            public class InitialMessageHandler : IHandleMessages<InitialMessage>
+            {
+                public Context Context { get; set; }
+
+                public async Task Handle(InitialMessage initialMessage, IMessageHandlerContext context)
+                {
+                    if (Context.TimesSenderHandlerInvoked == 0)
+                    {
+                        await context.Send(new DispatchedMessage { Id = Context.TestRunId });
```
danielmarbach: `.ConfigureAwait(false)` missing, only pointing it out because you added it to the behavior implementation below

SeanFeldman: Will remove from the behavior. In our tests we don't use `.ConfigureAwait()` so I shouldn't have used it here either. Thanks.

SeanFeldman:
I will have a PR that is specifically targets this change w/o anything else. Closing for now.

Issue #545, opened 2017-05-10
Disable manage rights check when installer are not configured to run


danielmarbach:
Carved out of #528

## Who's affected

Customers using connection strings with no manage rights

## Symptoms

When using a connection string with no manage rights and disabled installers the endpoint will refuse to start. Manage rights are required to operate an endpoint

## Solution

When installers are not enabled the transport should assume everything is configured properly and just run.

SeanFeldman:
Should be backported

PR #547, opened 2017-05-10
Manage rights

### LINKED ISSUE #546: Disable topology creation for subscriptions when no manage rights provided
Carved out of #528

## Who's affected

Customers using connection strings with no manage rights

## Symptoms

When an endpoint runs with no manage rights in the connection string and a subscription arrives the topology creation will try to create the topology section. The creation fails without manage rights. 

## Solution

Disable topology creation when no manage rights are given and assume everything is set up correctly

danielmarbach:
Closes #546 and #545 

No acceptance tests provided, see https://github.com/Particular/NServiceBus.AzureServiceBus/pull/547#issuecomment-301494942 and the following complexity

* Have an additional environment variable for a connection string on an existing namespace without manage rights
* Add restricted token to all namespaces on CI environment 
* All ASB maintainers would need to configure that new environment variable


yvesgoeleven:
@danielmarbach it is possible to generate access tokens with limited rights using the SharedAccessSignatureTokenProvider but it will require some setup. https://docs.microsoft.com/en-us/azure/service-bus-messaging/service-bus-sas

SeanFeldman:
Creating a **Policy** on entire namespace to effectively operate on all entities would require Subscription ID and a management certificate. Only Policies on individual entities do not require that.

We could generate an _identical_ policy name with the same SAS token for all the entities involved and use that in the connection string that has no management access... just a lot of management work.

src/Transport/Topology/TopologyCreator.cs
```diff
@@ -1,21 +1,41 @@
...
+                var namespacesWithoutManageRights = await ManageRightsCheck.Run(namespaces, container.Resolve<ReadOnlySettings>())
+                    .ConfigureAwait(false);
+                namespacesWithoutManageRightsJoined = string.Join(", ", namespacesWithoutManageRights.Select(alias => $"`{alias}`"));
+                return namespacesWithoutManageRights.Count == 0;
+            });
         }
 
         public async Task Create(TopologySection topology)
         {
+            if (!await hasManageRights)
```
SeanFeldman: Still requires `.ConfigureAwait(false)`, no?

src/Transport/Topology/TopologyCreator.cs
```diff
@@ -52,6 +72,14 @@ public TopologyCreator(ITransportPartsContainer container, IManageNamespaceManag
             }
         }
 
+        public async Task AssertManagedRights()
+        {
+            if (!await hasManageRights)
```
SeanFeldman: Still requires `.ConfigureAwait(false)`, no?

src/Transport/Utils/AsyncLazy.cs
```diff
@@ -0,0 +1,21 @@
+﻿namespace NServiceBus.AzureServiceBus.Utils
+{
+    using System;
+    using System.Runtime.CompilerServices;
+    using System.Threading.Tasks;
+
+    class AsyncLazy<T> : Lazy<Task<T>>
```
SeanFeldman: Perhaps `LazyAsync`? since it's an extension of Lazy to support task based values as well.

src/Transport/Utils/AsyncLazy.cs
```diff
@@ -0,0 +1,21 @@
...
+    {
+        public AsyncLazy(Func<T> valueFactory) :
```
SeanFeldman: This overload is to support value based option. Why for value based code not to use Lazy?

src/Transport/Topology/ICreateTopology.cs
```diff
@@ -14,8 +14,10 @@ public interface ICreateTopology
     }
 
     // TODO: Move into internalized ICreateTopology in v8
-    interface ITearDownTopology
+    interface ICreateTopologyInternal : ICreateTopology
```
yvesgoeleven: Can't we just keep the name ICreateTopology?

src/Transport/Topology/TopologyCreator.cs
```diff
@@ -1,21 +1,41 @@
...
+                namespacesWithoutManageRightsJoined = string.Join(", ", namespacesWithoutManageRights.Select(alias => $"`{alias}`"));
+                return namespacesWithoutManageRights.Count == 0;
+            });
         }
 
         public async Task Create(TopologySection topology)
         {
+            if (!await hasManageRights)
+            {
+                Logger.Info($"Configured to create topology, but have no manage rights for the following namespace(s): {namespacesWithoutManageRightsJoined}. Execution will continue and assume the topology is already created.");
```
yvesgoeleven: Wasn't this supposed to throw? If create is called and no manage rights are given

src/Transport/Utils/AsyncLazy.cs
danielmarbach: Happy to remove this overload

src/Transport/Utils/AsyncLazy.cs
danielmarbach: It is an async Lazy, not sure the proposed name change expresses that better

src/Transport/Topology/TopologyCreator.cs
danielmarbach: Yes

src/Transport/Topology/ICreateTopology.cs
danielmarbach: My problem was the following:

We had `ITeardownTopology` which was introduced for internal usage until we can internalise the interfaces. But having an AssertManagedRights doesn't really make sense on `ITeardownTopology`. That's why I went with introducing this internal interface (should also exist in develop) that now contains the methods defined in the public base interface as well as the assert method.

src/Transport/Topology/TopologyCreator.cs
danielmarbach: Nope for that we have the AssertManagedRights method. The create will just silently bypass if the rights are not given

src/Transport/Topology/TopologyCreator.cs
danielmarbach: Scratch the comment. This uses a custom awaiter which doesn't require ConfigureAwait(false)

danielmarbach: Do you want me to remove the custom awaiter and replace with more verbose

```!await hasManageRights.Value.ConfigureAwait(false)```?

Comment (SeanFeldman) [inline src/Transport/Topology/TopologyCreator.cs]: Thank you for explanation. If it's not even possible, then not needed.

**Comment by SeanFeldman** [comment]
No need to add a test. ASB is very solid on **not** executing any code if manage claim is missing. Setting that up would be too much of a hassle with very little return. My vote to skip it.

**Comment by danielmarbach** [comment]
Changed please re-review

File: src/Tests/API/APIApprovals.ApproveAzureServiceBusTransport.approved.txt [hunk 7]
```diff
@@ -334,7 +334,7 @@
         public System.Func<NServiceBus.Transport.ICreateQueues> GetQueueCreatorFactory() { }
         public System.Func<NServiceBus.Transport.IManageSubscriptions> GetSubscriptionManagerFactory() { }
         public void Initialize(NServiceBus.Settings.SettingsHolder settings) { }
-        public async System.Threading.Tasks.Task<NServiceBus.Transport.StartupCheckResult> RunPreStartupChecks() { }
+        public System.Threading.Tasks.Task<NServiceBus.Transport.StartupCheckResult> RunPreStartupChecks() { }
```
SeanFeldman: I thought API Approver should not differentiate between `async Task<T>` and `Task<T>`.

src/Transport/Topology/ICreateTopology.cs
SeanFeldman: I'm ok with this. Will be more consistent with the `develop` indeed.

SeanFeldman:
@danielmarbach I've added comment on the issues, but probably worth mentioning here as well. This fix should be backported to 7.2 (not sure about 7.1).

src/Tests/API/APIApprovals.ApproveAzureServiceBusTransport.approved.txt
danielmarbach: I think we are using an older version. New version fixes this

danielmarbach:
Rebased and merged into bcb781651a55783132688d68c3a69d2c4868c45f

danielmarbach:
Talked to @yvesgoeleven and we decided to not backport it to 7.1. Existing projects seem to use manage rights since nobody has complained about it and if a new project gets started which wants to use a connection string without manage rights the assumption is that it will use latest 7.2.x

PR #548, opened 2017-05-15
Remove INamespacePartitioningStrategy, ICompositionStrategy, ISanitizationStrategy and IHandleOversizedBrokeredMessages from the container

danielmarbach:


src/Tests/API/APIApprovals.ApproveAzureServiceBusTransport.approved.txt
```diff
@@ -65,8 +65,7 @@ namespace NServiceBus
         public NServiceBus.AzureServiceBusMessageSenderSettings BackOffTimeOnThrottle(System.TimeSpan backoffTime) { }
         public NServiceBus.AzureServiceBusMessageSenderSettings MaximuMessageSizeInKilobytes(int sizeInKilobytes) { }
         public NServiceBus.AzureServiceBusMessageSenderSettings MessageSizePaddingPercentage(int percentage) { }
-        public NServiceBus.AzureServiceBusMessageSenderSettings OversizedBrokeredMessageHandler<T>(T instance)
```
danielmarbach: Discussed this with @yvesgoelevan and we figured it is OK to do this in a major. There is no way we can deprecate this API properly that's why it was directly replaced

SeanFeldman:
It fails on flaky tests (deferral), not the changes in the PR.

src/Transport/Topology/Topologies/TypeExtensions.cs
```diff
@@ -0,0 +1,29 @@
...
+{
+    using System;
+    using System.Linq;
+    using System.Reflection;
+    using Settings;
+
+    static class TypeExtensions
+    {
+        // potentially we want to remove this when we simplify hooking customer owned implementations into the topology
+        public static TReturnedType CreateInstance<TReturnedType>(this Type typeToResolve, ReadOnlySettings settings)
```
danielmarbach: This keeps backwards compatibility:

Following assumptions. Right now with the transport parts container it is possible to:

- Resolve dependencies that are in the parts container (all of them are internal)
- Resolve ReadonlySettings
- Have empty default ctor

From a customer perspective it is only possible to do the following:

- have a ctor on their implementations which required ReadonlySettings or
- have an empty parameterless ctor

Therefore this extension methods keeps this by using either default ctor and calling or assume ctor that has a dependency to readonly settings. In all other cases it blows up

danielmarbach:
Rebased

SeanFeldman:
@yvesgoeleven is marked as a reviewer as well. Randomly failing non-related tests are **not** affected by the changes in this PR.

danielmarbach:
@seanfeldman yves and I did this together. Good to go

SeanFeldman:
Merged.

PR #551, opened 2017-05-15
Issue 524 micsonfigured bundles can cause message loss

SeanFeldman:
Connects to #524 

With #547 (manage rights check) merge, this PR addresses the issue with bundles configuration.

The code has been changed (as of commit https://github.com/Particular/NServiceBus.AzureServiceBus/pull/551/commits/f27aa8df2171f1131c63febb60030ffc9d0e95fd) to eliminate any knowledge about namespace bundles configuration from ASB transport infrastructure, ForwardingTopology, and configuration settings. The ForwardingTopologySectionManager is the one that invokes the check directly and uses it's results to perform calculations.


SeanFeldman:

src/Transport/Topology/NumberOfTopicsInBundleCheck.cs
```diff
@@ -0,0 +1,37 @@
...
+                var namespaceManager = manageNamespaceManagerLifeCycle.Get(namespaceConfiguration.Alias);
+                var namespaceManagerThatCanQueryAndFilterTopics = namespaceManager as NamespaceManagerAdapter;
+
+                // if user has provided an implementation of INamespaceManager, skip the checks all together
+                if (namespaceManagerThatCanQueryAndFilterTopics == null)
+                {
+                    break;
+                }
+
+                var numberOfTopics = 1;
```
danielmarbach: Same default value used as well here https://github.com/Particular/NServiceBus.AzureServiceBus/pull/551/files#diff-c350c6bebf11eb7d301918634d454b42R31

src/Transport/Topology/Topologies/ForwardingTopologySectionManager.cs
```diff
@@ -16,12 +17,21 @@ class ForwardingTopologySectionManager : ITopologySectionManager
...
         {
             this.settings = settings;
             this.container = container;
+
+            namespaceBundleConfigurations = new Lazy<NamespaceBundleConfigurations>(() =>
+            {
+                var manageNamespaceManagerLifeCycle = container.Resolve<IManageNamespaceManagerLifeCycle>();
+                var namespaceConfigurations = settings.Get<NamespaceConfigurations>(WellKnownConfigurationKeys.Topology.Addressing.Namespaces);
+                var bundlePrefix = settings.Get<string>(WellKnownConfigurationKeys.Topology.Bundling.BundlePrefix);
+                var bundleConfigurations = NumberOfTopicsInBundleCheck.Run(manageNamespaceManagerLifeCycle, namespaceConfigurations, bundlePrefix).GetAwaiter().GetResult();
```
danielmarbach: Discussed this with @SeanFeldman and we said that on develop the section manager has to become async and this code probably will use `AsyncLazy` as well as not use the container

src/Transport/Topology/Topologies/ForwardingTopologySectionManager.cs
```diff
@@ -102,28 +112,28 @@ public TopologySection DeterminePublishDestination(Type eventType)
...
                     Namespaces = namespaces
                 };
             });
         }
 
-        IEnumerable<EntityInfo> SelectSingleRandomTopicFromBundle(List<EntityInfo> entityInfos)
+        IEnumerable<EntityInfo> SelectFirstTopicFromBundle(List<EntityInfo> entityInfos)
         {
-            var index = randomGenerator.Next(0, entityInfos.Count);
+            const int index = 1;
```
danielmarbach: Why not index = 0?

src/Transport/Topology/Topologies/ForwardingTopologySectionManager.cs
```diff
@@ -102,28 +112,28 @@ public TopologySection DeterminePublishDestination(Type eventType)
...
             var selected = entityInfos[index];
 
-            return entityInfos.Where(i => i.Path == selected.Path);
+            yield return entityInfos.First(i => i.Path == selected.Path);
```
danielmarbach: Isn't this overly complex? We already have the first entry selected by using the list indexer. Can't we just return that? Why is the additional First with the condition for the selection needed?

src/Transport/Topology/Topologies/ForwardingTopologySectionManager.cs
```diff
@@ -273,16 +283,21 @@ void BuildTopicBundles(RuntimeNamespaceInfo[] namespaces, AddressingLogic addres
             var numberOfEntitiesInBundle = settings.Get<int>(WellKnownConfigurationKeys.Topology.Bundling.NumberOfEntitiesInBundle);
             var bundlePrefix = settings.Get<string>(WellKnownConfigurationKeys.Topology.Bundling.BundlePrefix);
 
-            for (var i = 1; i <= numberOfEntitiesInBundle; i++)
+            foreach (var @namespace in namespaces)
             {
-                topics.AddRange(namespaces.Select(n => new EntityInfo
+                var numberOfTopicsFound = namespaceBundleConfigurations.Value.GetNumberOfTopicInBundle(@namespace.Alias);
```
danielmarbach: So the implications here are that the first materialization of `.Value` will take time and after the first call the result will be cached.

src/Transport/Topology/NumberOfTopicsInBundleCheck.cs
SeanFeldman: Not sure what's the question about. Correct, if we fail to find a namespace with number of topics, default will be 1.

src/Transport/Topology/Topologies/ForwardingTopologySectionManager.cs
SeanFeldman: Should be index 0 as we want to route everything through topic called "bundle-1".

src/Transport/Topology/Topologies/ForwardingTopologySectionManager.cs
SeanFeldman: Unnecessary. The first item (with the path `bundle-1`) will be enough. I'll update the code.

src/Transport/Topology/Topologies/ForwardingTopologySectionManager.cs
SeanFeldman: Correct.

SeanFeldman:
@danielmarbach I've updated single topic selection. The rest not sure if you wanted anything to be actioned or those were general comments.

src/Transport/Topology/NumberOfTopicsInBundleCheck.cs
danielmarbach: I was merely pointing out here that we have some redundancy. I can live with that redundancy though.

PR #554, opened 2017-05-15
Porting fixes from #542 to develop

SeanFeldman:
Connects to #542 (Message is completed instead of abandoned when incoming message processing is aborted)

This fix should ensure that the edge case is covered and there's no chance we'd generate duplicates when we shouldn't.

@Particular/azure-maintainers should be merged _after_ #553 (`hotfix-7.1.7`) and #552 (`hotfix-7.2.2`) are merged.


danielmarbach:
@SeanFeldman why after? Since it is develop this can be merged right away or not?

SeanFeldman:
`develop` is red because of the flaky tests related to the delayed delivery tests.

PR #555, opened 2017-05-17
Hotfix 7.2.3

### LINKED ISSUE #545: Disable manage rights check when installer are not configured to run

danielmarbach:
Closes #545 and #546 

Announcement https://github.com/Particular/Announcements/pull/139


danielmarbach:
@yvesgoeleven @SeanFeldman you already reviewed it but please review again

src/Transport/Seam/TransportResourcesCreator.cs
```diff
@@ -17,7 +17,16 @@ public TransportResourcesCreator(ICreateTopology topologyCreator, ITopologySecti
 
         public async Task CreateQueueIfNecessary(QueueBindings queueBindings, string identity)
         {
-            if (resourcesCreated) return;
+            if (resourcesCreated)
+            {
+                return;
```
SeanFeldman: Question: would log file contain any sort of information why we're not creating infrastructure? I.e. if someone is looking at the log to determine that topology is not created _because_ installers are disabled.

danielmarbach:
both `TransportResourcesCreator `and `SubscriptionManager` call into the topology creator. The only difference is that the transport resource creator does an assert first before it tries to create the topology.

src/Transport/Seam/TransportResourcesCreator.cs
danielmarbach: Create queues only gets called when the installers are enabled. The assumption of the code is that although the implementation can be called multiple times the queue bindings will contain all the bindings to create and therefore we only create the topology once. I don't think we need to change this as part of the hotfix.

SeanFeldman:
@danielmarbach do you need a hand with releasing this one?

danielmarbach:
@SeanFeldman I can manage it. I was waiting on more feedback on the announcement but I think we are good to go

danielmarbach:
Waiting for 7.1.8 so that the announcement can be combined

PR #569, opened 2017-06-02
Timeout exception handled as transient (#568)

janpieterz:
Looked to prove this using a unit test but ended up being pretty complicated to bootstrap a test, the normal timeout doesn't seem to be tested either in the `MessageReceiverNotifier`.

Checked other usages of MessagingException. Almost all cases handled TimeoutException. Ones that didn't:

https://github.com/Particular/NServiceBus.AzureServiceBus/blob/develop/src/Transport/Creation/AzureServiceBusSubscriptionCreator.cs#L117
https://github.com/Particular/NServiceBus.AzureServiceBus/blob/develop/src/Transport/Creation/AzureServiceBusForwardingSubscriptionCreator.cs#L160

Since both are delete and it's handled specifically in all other modification scenario's where MessagingException is handled, I assumed this was intentional. If not, let me know and I'll handle it there too.

If anything needs to change, please let me know.

### DIFF
--- a/src/Transport/Receiving/MessageReceiverNotifier.cs
+++ b/src/Transport/Receiving/MessageReceiverNotifier.cs
@@ -154,9 +154,14 @@ namespace NServiceBus.Transport.AzureServiceBus
 
                 //- It is raised when the underlying connection closes because of our close operation
                 var messagingException = exceptionReceivedEventArgs.Exception as MessagingException;
+                var timeoutException = exceptionReceivedEventArgs.Exception as TimeoutException;
                 if (messagingException != null && messagingException.IsTransient)
                 {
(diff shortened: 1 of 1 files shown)


yvesgoeleven:
@janpieterz imo simply handling all TimeoutExceptions is way to drastic. They often indicate problems with the namespace. As the renew is not so important we can catch and log exceptions coming from that angle, but if they occur consistently in the dispatch path (send, complete/abort), then we need to trigger the circuit breaker

yvesgoeleven:
I see a similar discussions started on the issue already https://github.com/Particular/NServiceBus.AzureServiceBus/issues/568

SeanFeldman:
@janpieterz thank you for this PR. We have a replacement PR that will take care of this issue (#568). Closing this PR in favor of #583. Fix will be reflected in `hotfix-7.2.6` that will be released shortly.
Thank you again.

PR #570, opened 2017-06-02
Porting hotfix-7.2.4 to develop - Misconfigured bundles can cause message loss (#551)

SeanFeldman:
Connected to #524 

- Port of `hotfix-7.2.4` to develop
- Enabling C# 7


src/Transport/Topology/Topologies/ForwardingTopologySectionManager.cs
```diff
@@ -89,30 +98,33 @@ public TopologySectionInternal DetermineResourcesToCreate(QueueBindings queueBin
             };
         }
 
-        public TopologySectionInternal DeterminePublishDestination(Type eventType)
+        public async Task<TopologySectionInternal> DeterminePublishDestination(Type eventType)
         {
-            return publishDestinations.GetOrAdd(eventType, t =>
+            TopologySectionInternal destination;
```
SeanFeldman: @danielmarbach I couldn't use `ConcurrentDictionary.GetOrAdd`. Can you poke any holes in this? Thanks.

SeanFeldman:
@Particular/azure-maintainers please review/merge

src/Transport/Connectivity/INamespaceManager.cs
```diff
@@ -31,6 +31,8 @@ interface INamespaceManagerInternal
 
         Task<IEnumerable<RuleDescription>> GetRules(SubscriptionDescription subscriptionDescription);
         Task<SubscriptionDescription> CreateSubscription(SubscriptionDescription subscriptionDescription, RuleDescription ruleDescription);
+        // TODO: enable back on the interface once internalized in v8
```
danielmarbach: Is this comment still valid since we are now on develop?

src/Transport/Topology/NumberOfTopicsInBundleCheck.cs
```diff
@@ -0,0 +1,37 @@
...
+    static class NumberOfTopicsInBundleCheck
+    {
+        public static async Task<NamespaceBundleConfigurations> Run(IManageNamespaceManagerLifeCycleInternal manageNamespaceManagerLifeCycle, NamespaceConfigurations namespaceConfigurations, string bundlePrefix)
+        {
+            var namespaceBundleConfigurations = new NamespaceBundleConfigurations();
+
+            foreach (var namespaceConfiguration in namespaceConfigurations)
+            {
+                var namespaceManager = manageNamespaceManagerLifeCycle.Get(namespaceConfiguration.Alias);
+                var namespaceManagerThatCanQueryAndFilterTopics = namespaceManager as NamespaceManagerAdapterInternal;
```
danielmarbach: On develop we can directly acquire the implementation and there is not cast needed correct?

src/Transport/Topology/NumberOfTopicsInBundleCheck.cs
```diff
@@ -0,0 +1,37 @@
...
+
+                // if user has provided an implementation of INamespaceManager, skip the checks all together
+                if (namespaceManagerThatCanQueryAndFilterTopics == null)
```
danielmarbach: This check also not needed on develop?

src/Transport/Topology/Topologies/ForwardingTopologySectionManager.cs
```diff
@@ -89,30 +98,32 @@ public TopologySectionInternal DetermineResourcesToCreate(QueueBindings queueBin
             };
         }
 
-        public TopologySectionInternal DeterminePublishDestination(Type eventType)
+        public async Task<TopologySectionInternal> DeterminePublishDestination(Type eventType)
```
danielmarbach: On the fence about this. Maybe we need to brainstorm here about the consequences of this change. GetOrAdd could be executed multiple times but only one result would be added. The call offerred some "atomicity" but this new version splits it into multiple calls. It is possible to combine Lazy or AsyncLazy with the dictionary if we do it right. The question is whether it is necessary.

@Scooletz thoughts?

src/Transport/Sending/IBatcher.cs
```diff
@@ -1,10 +1,11 @@
 namespace NServiceBus.Transport.AzureServiceBus
 {
     using System.Collections.Generic;
+    using System.Threading.Tasks;
     using Transport;
 
     interface IBatcherInternal
     {
-        IList<BatchInternal> ToBatches(TransportOperations operations);
+        Task<List<BatchInternal>> ToBatches(TransportOperations operations);
```
yvesgoeleven: Why is this interface returning a Task? I would not expect any IO to happen inside a conversion function

src/Transport/Connectivity/INamespaceManager.cs
SeanFeldman: 👍

src/Transport/Topology/NumberOfTopicsInBundleCheck.cs
SeanFeldman: This is exactly why I had that TODO comment [here](https://github.com/Particular/NServiceBus.AzureServiceBus/pull/570#discussion_r120271310) 😃

src/Transport/Topology/NumberOfTopicsInBundleCheck.cs
SeanFeldman: Correct. Not needed. Same as above.

src/Transport/Sending/IBatcher.cs
SeanFeldman: You're correct, but due to the fact that `topologySectionManager.DeterminePublishDestination()` now needs to access namespace to detect topics in bundle when using forwarding topology, there is IO involved. And async has to be propagated up the stack.

src/Transport/Topology/Topologies/ForwardingTopologySectionManager.cs
SeanFeldman: @danielmarbach I'm not convinced it's needed.
If you still think that we should, let discuss discuss on a call.

SeanFeldman:
@Particular/azure-maintainers I also wonder if for develop we should switch from the default 2 topics in a bundle to a single topic. Currently, it's [set to 2](https://github.com/Particular/NServiceBus.AzureServiceBus/blob/b500c13e4d32c411cd0cf69fdb66615a12fb947d/src/Transport/Topology/Topologies/ForwardingTopology.cs#L50) by default, and we use the max between the default value and the actual/found number of topics. 

I think we should go with the default of 1 and we find more than 1, we'll subscribe to whatever is found (this PR provided the functionality).

danielmarbach:
> I think we should go with the default of 1 and we find more than 1, we'll subscribe to whatever is found (this PR provided the functionality).

Makes sense to me.

src/Transport/Sending/IBatcher.cs
yvesgoeleven: @SeanFeldman I don't like this, can we split the IO away from the conversion please, do it up front and pass in the results instead?

yvesgoeleven: there is a new service that is supposed to help make every services in azure able to "react" to events. similar to how Azure Functions hooks on lots of things

SeanFeldman: Had to do it this way as it has to run after we've gone through installers and manage right checks.
I'll check if there's an option.

SeanFeldman: @yvesgoeleven I don't see how we'd call it before.
Fist, only if installers are enabled, we will need to check if manage rights check was executed.
Manage rights check is performed by queue creator and then section manager has to provide section for topology creator to operate on. So we don't have a place where we'd do it up-front.

- TransportResourcesCreator : ICreateQueues
  - AssertManageRights()
  - DetermineResourcesToCreate(queueBindings)
       _if no topics found in section manager, call_ BuildTopicBundles()

Also, `BuildTopicBundles()` can be invoked from 2 additional places:

From the subscription manager

- SubscriptionManager : IManageSubscriptions
  - DetermineResourcesToSubscribeTo
    - BuildSubscriptionHierarchy(eventType)
           _if no topics found in section manager, call_ BuildTopicBundles()

and from the batcher when messages are dispatched

- DeterminePublishDestination(eventType)
  - Batcher.AddMulticastOperationBatches()
      _if no topics found in section manager, call_ BuildTopicBundles()

In all 3 cases section manager has to have the information about _actual_ number of topics.

If you have a suggestion how to tackle it, by all means. I don't have a better idea.

yvesgoeleven: But dispatch happens after the other 2, so you can just remember the result somewhere and reuse it during dispatch, we can't do the IO again during every dispatch operation.

danielmarbach: @yvesgoeleven that's what the async local does there. Agree it would be better to not return a task here. I think we have two options:

- Address it now (I'm very limited with my availability so @yvesgoeleven would need to jump in)
- Merge it with a TODO and address it in the larger PDev refactorings (we basically build dept now)

I'm OK with both approaches. If you need/want my help it would have to wait until my return (not implying I'm required here just offering my help)

SeanFeldman: I think there's some misunderstanding here @yvesgoeleven.
The IO is **not** performed on every dispatch. The result is [cached and re-used](https://github.com/Particular/NServiceBus.AzureServiceBus/blob/d4f6f967bc665d9330e55b753228c369e1d0877a/src/Transport/Topology/Topologies/ForwardingTopologySectionManager.cs#L37).

SeanFeldman: When the cached check is used

```c#
await namespaceBundleConfigurations.Value.ConfigureAwait(false);
```

the task will be cached and we won't perform any IO.

yvesgoeleven: I suggest we do an explicit .GetAwaiter().GetResult() here, with a comment that we are certain that the value is cached, this way the Task can be removed from the interface and it is clear what the assumption is.

yvesgoeleven: Or, second option is to load the destinations per message types up front and use a dictionary of those things during the conversion.

SeanFeldman: The original implementation (`hotfix-7.2.4` did exactly that while using `Lazy` instead of `AsyncLazy`).
Downside was a single blocking invocation. I can revert changes to the API and prevent Task propagation so that `IBatcher` and `ITopologySectionManagerInternal` do not bleed `Task` because of the `BuildTopicBundles` method.

SeanFeldman: Rather than doing
```c#
namespaceBundleConfigurations.Value.GetAwaiter().GetResult();
```

I'd suggest to perform a blocking call in the Lazy block, replacing `AsynLazy` with `Lazy`.

SeanFeldman:
@yvesgoeleven @danielmarbach to move forward with this, this is what I'm applying:
1. Revert the async changes (Task based API for `IBatcher` and `ITopologySectionManagerInternal`). APIs that is not IO based should not have `Task` in their signature, agreed.
1. `ForwardingTopologySectionManager` calls `NumberOfTopicsInBundleCheck.Run` once only in a blocking manner. The call is cached.
1. Left a TODO comment to review this later. We have other significant things to take care of rather than looking into a single blocking call that is cached. The async purists will forgive me for this sin 😃 
1. Additionally, rebased on top of new `develop` csproj format.

Please review this PR one more time with all the changes and updates.

yvesgoeleven:
👍 for me

PR #583, opened 2017-06-27
Hotfix 7.2.6

### LINKED ISSUE #558: OutgoingBatchRouter doesn't await for the over-sized message handler
## Who's affected

- Anyone overriding default [over-sized message handler](https://docs.particular.net/transports/azure-service-bus/oversized-sends).

## Symptoms

Handler invocation is not awaited, which could lead to undesired results.

## Original issue

The returned [task is not awaited](https://github.com/Particular/NServiceBus.AzureServiceBus/blob/3b2bd689fc07a8b93536f4a0182a9dd4fb36fe52/src/Transport/Sending/OutgoingBatchRouter.cs#L229).

Also update to latest Particular code analysis package.

**Affected versions**
- `develop`
- `support-7.2`

SeanFeldman:
Fixes #558 and #568 bugs.

### DIFF
--- a/src/Transport/Receiving/MessageReceiverNotifier.cs
+++ b/src/Transport/Receiving/MessageReceiverNotifier.cs
@@ -13,7 +13,7 @@ namespace NServiceBus.Transport.AzureServiceBus
     using NServiceBus.AzureServiceBus.Utils;
     using Settings;
 
-    partial class MessageReceiverNotifier : INotifyIncomingMessages
+    class MessageReceiverNotifier : INotifyIncomingMessages
     {
         public MessageReceiverNotifier(IManageMessageReceiverLifeCycle clientEntities, IConvertBrokeredMessagesToIncomingMessages brokeredMessageConverter, ReadOnlySettings settings)
         {
@@ -186,7 +186,7 @@ namespace NServiceBus.Transport.AzureServiceBus
                 var invocations = (ConcurrentDictionary<Task, Task>) state;
                 Task toBeRemoved;
                 invocations.TryRemove(t, out toBeRemoved);
-            }, pipelineInvocations, TaskContinuationOptions.ExecuteSynchronously);
+            }, pipelineInvocations, TaskContinuationOptions.ExecuteSynchronously).Ignore();
             return processTask;
         }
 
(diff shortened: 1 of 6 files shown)


SeanFeldman:
I still have second thoughts on how we've agreed to handle the check. See my comment [here](https://github.com/Particular/NServiceBus.AzureServiceBus/issues/568#issuecomment-311514606).

src/Transport/Receiving/MessageReceiverNotifier.cs
```diff
@@ -187,7 +186,7 @@ Task ReceiveMessage(IMessageReceiver internalReceiver, BrokeredMessage message,
                 var invocations = (ConcurrentDictionary<Task, Task>) state;
                 Task toBeRemoved;
                 invocations.TryRemove(t, out toBeRemoved);
-            }, pipelineInvocations, TaskContinuationOptions.ExecuteSynchronously);
+            }, pipelineInvocations, TaskContinuationOptions.ExecuteSynchronously).Ignore();
```
danielmarbach: How about putting `Ignore` on a new line to make it better visible?

src/Transport/Receiving/MessageReceiverNotifier.cs
```diff
@@ -152,11 +153,9 @@ public void Start()
                     return;
                 }
 
-                //- It is raised when the underlying connection closes because of our close operation
-                var messagingException = exceptionReceivedEventArgs.Exception as MessagingException;
-                if (messagingException != null && messagingException.IsTransient)
+                if (exceptionReceivedEventArgs.Exception.IsTransientException(exceptionReceivedEventArgs.Action))
```
danielmarbach: Not hang up on this but I would have gone for an extension method that is applied to `exceptionReceivedEventArgs` instead of the current proposal because it makes the code less verbose

src/Transport/Receiving/MessageReceiverNotifier.cs
SeanFeldman: Frankly, I want to scrape it and do the [right thing](https://github.com/Particular/NServiceBus.AzureServiceBus/issues/568#issuecomment-311514606). 

No pressure @yvesgoeleven 😃

yvesgoeleven: @SeanFeldman I feel none :p

src/Transport/Receiving/MessageReceiverNotifier.cs
SeanFeldman: With the change to how we'll evaluate if exception is transient or not, it will be less verbose.
I'm not sure `exceptionReceivedEventArgs.IsTransient()` would read right as it's the exception that should be checked for being transient, not the arguments.

SeanFeldman:
@yvesgoeleven @danielmarbach I've changed the implementation. 2nd round 🙂

PR #598, opened 2017-08-17
Removed last usage of the internal container

### LINKED ISSUE #448: Replace the internal container with newing up instances directly where needed
@SeanFeldman and I made a spike to show that it is possible to replace the internal container inside the topology. The approach would be following

- The topology is the composition root of the infrastructure required to run it
- All singleton scoped objects are held as fields inside the topology
- On demand created objects with transient scope are created with `new `when needed and held by the instance that needs the transient objects
- All container resolve calls happening during runtime are replaced with constructor injection or explicit method injection (method accepting parameters which define what dependencies an instance needs).

danielmarbach:
Closes #448


src/Transport/Topology/Topologies/EndpointOrientedTopology.cs
```diff
@@ -24,23 +24,17 @@ class EndpointOrientedTopologyInternal : ITopologyInternal
...
-
-        internal EndpointOrientedTopologyInternal(ITransportPartsContainerInternal container)
-        {
-            this.container = container;
-        }
-
         public bool HasNativePubSubSupport => true;
         public bool HasSupportForCentralizedPubSub => true;
         public TopologySettings Settings { get; } = new TopologySettings();
+        public ITopologySectionManagerInternal TopologySectionManager => topologySectionManager;
```
yvesgoeleven: can't this be internal? It's only exposed for testing purposes

src/Transport/Topology/Topologies/EndpointOrientedTopology.cs
```diff
@@ -24,23 +24,17 @@ class EndpointOrientedTopologyInternal : ITopologyInternal
...
+        public IOperateTopologyInternal Operator => topologyOperator;
```
yvesgoeleven: can't this be internal? It's only exposed for testing purposes

src/Transport/Topology/Topologies/ForwardingTopology.cs
```diff
@@ -31,17 +30,11 @@ class ForwardingTopologyInternal : ITopologyInternal
...
-        internal ForwardingTopologyInternal(ITransportPartsContainerInternal container)
-        {
-            this.container = container;
-        }
-
         public bool HasNativePubSubSupport => true;
         public bool HasSupportForCentralizedPubSub => true;
         public TopologySettings Settings { get; } = new TopologySettings();
-
+        public ITopologySectionManagerInternal TopologySectionManager => topologySectionManager;
```
yvesgoeleven: can't this be internal? It's only exposed for testing purposes

src/Transport/Topology/Topologies/ForwardingTopology.cs
```diff
@@ -31,17 +30,11 @@ class ForwardingTopologyInternal : ITopologyInternal
...
+        public IOperateTopologyInternal Operator => topologyOperator;
```
yvesgoeleven: can't this be internal? It's only exposed for testing purposes

src/Transport/Topology/Topologies/EndpointOrientedTopology.cs
yvesgoeleven: class itself is internal

src/Transport/Topology/Topologies/EndpointOrientedTopology.cs
yvesgoeleven: class itself is internal

src/Transport/Topology/Topologies/ForwardingTopology.cs
yvesgoeleven: class itself is internal

src/Transport/Topology/Topologies/ForwardingTopology.cs
yvesgoeleven: class itself is internal

yvesgoeleven:
Build server says a few inspections need to be fixed, but besides that, good to go for me

danielmarbach:
@yvesgoeleven green now

PR #601, opened 2017-08-18
Configure strategy instance and explicitely initialize strategy with settings

danielmarbach:
Came to the conclusion that customers might rely on the possibility to access `ReadOnlySettings`. So instead of wrapping the settings per strategy in a custom object (that would need to be extendable for custom settings), the strategies will have a dedicated initialize phase where the settings are passed in. That allows caching the relevant settings so that runtime settings access is no longer needed. 

Furthermore, this PR removes the type based settings and allows passing in the instance of the strategy directly to simplify custom dependencies passing into the ctor.


danielmarbach:
One transport test failed. Not sure if it is related to this change though

danielmarbach:
@Particular/azure-maintainers build is green

danielmarbach:
I just realized there is a way more elegant and simpler approach. Will update this PR

danielmarbach:
While starting the documentation of this I realized how awkward the configuration code can look. You create a new instance of the strategy and then go over the settings extensions to configure it.

```
var strategy = sanitization.UseStrategy(new ValidateAndHashIfNeeded());
        strategy.QueuePathSanitization(queuePath => "sanitized queuePath");
        strategy.TopicPathSanitization(topicPath => "sanitized topicPath");
        strategy.SubscriptionNameSanitization(
            subscriptionNameSanitizer: subscriptionName =>
            {
                return "sanitized subscriptionName";
            });
        strategy.RuleNameSanitization(
            ruleNameSanitizer: ruleName =>
            {
                return "sanitized ruleName";
            });
        strategy.Hash(
            hash: pathOrName =>
            {
                return "hashed pathOrName";
            });
```

After reflecting on this I came to the conclusion it is not worth forcing the changes on the users for an arguably more weird looking API. I moved all optimizations out of this PR into
https://github.com/Particular/NServiceBus.AzureServiceBus/pull/603

Closing this

Issue #282, opened 2016-07-21
Don't use ReadOnlySettings at runtime


melkio:
Taskforce: @danielmarbach @seanfeldman
## Original issue raised by @melkio

Transport shouldn't use `ReadOnlySettings` at "runtime". It should read settings and configure components passing raw values into constructor (or somewhere else).
This approach should be better for two reasons:
- avoid _boxing/unboxing_ with `ValueType`: `settings.Get<bool>(...)` invocation is casting `object` stored into dictionary to `bool`
- don't rely on a component (`ReadOnlySettings`) managed by Core, that could change behavior in the future

## Assumptions
1. The work of this taskforce is not going to be based on v7 release of the transport, but result in v7.x.
2. This is a starting issue and after retrospective, if found effective, spin up the next item.
3. The issue should solely focus on Settings and any other issues discovered during the work should be registered under ASB repo.

## Expectations
1. TF members to communicate to their maintainer groups and get them involved when needed, especially for discussions where GH issue is not the right medium to be used. 
2. Have mutual respect and understanding when dealing with code that is done the way it's done for historical reasons.

## PoA
- [X] Config
- [X] Creation already good
- [x] Connectivity https://github.com/Particular/NServiceBus.AzureServiceBus/pull/604
- [x] Receiving https://github.com/Particular/NServiceBus.AzureServiceBus/pull/605
- [x] Addressing https://github.com/Particular/NServiceBus.AzureServiceBus/pull/603
- [x] Sending #607
- [x] Topology #608

yvesgoeleven:
I don't understand. Configuration API fills up this structure only, why should we not be allowed to use it? Are we supposed to copy everything over to some other structure?

danielmarbach:
The settings API is a composition level API only. The design is the following:

After a certain point in the bootstrapping phase, the settings holder gets locked for modifications. Then the features get invoked. You get the settings holder as a segregated interface called `ReadOnlySettings`. Those settings are meant to be transferred to custom structure, bool flags, value types of any sorts. Like @melkio mentioned not following this design principle can have following consequences:

* If at some point core takes a decision to clear all settings after the feature configuration phase, your code breaks
* Using read-only settings during runtime does not make sense because
 1. The settings keys will never change after the feature configuration phase
[...]

PR #607, opened 2017-08-21
Remove ReadOnlySettings from sending runtime

SeanFeldman:
Connects to #282 

No longer relying on `ReadOnlySettings` during runtime.

PR #608, opened 2017-08-21
Remove readonly settings from topology related code

SeanFeldman:
Connects to #282 

- No longer relying on `ReadOnlySettings` during runtime for topology related code.
- Modified `MessageReceiveNotifier` not to need `ReadOnlySettings` (replaced with `MessageReceiveNotifierSettings`).

### DIFF
--- a/src/Transport/Topology/TopologyCreator.cs
+++ b/src/Transport/Topology/TopologyCreator.cs
@@ -19,10 +19,11 @@ namespace NServiceBus.Transport.AzureServiceBus
             this.queuesCreator = queuesCreator;
             this.subscriptionsCreator = subscriptionsCreator;
             this.namespaces = namespaces;
+            var namespaceConfigurations = settings.Get<NamespaceConfigurations>(WellKnownConfigurationKeys.Topology.Addressing.Namespaces);
 
             hasManageRights = new AsyncLazy<bool>(async () =>
             {
-                var namespacesWithoutManageRights = await ManageRightsCheck.Run(namespaces, settings)
+                var namespacesWithoutManageRights = await ManageRightsCheck.Run(namespaces, namespaceConfigurations)
                     .ConfigureAwait(false);
                 namespacesWithoutManageRightsJoined = string.Join(", ", namespacesWithoutManageRights.Select(alias => $"`{alias}`"));
                 return namespacesWithoutManageRights.Count == 0;
@@ -33,7 +34,7 @@ namespace NServiceBus.Transport.AzureServiceBus
         {
             if (!await hasManageRights.Value.ConfigureAwait(false))
             {
-                Logger.Info($"Configured to create topology, but have no manage rights for the following namespace(s): {namespacesWithoutManageRightsJoined}. Execution will continue and assume the topology is already created.");
+                logger.Info($"Configured to create topology, but have no manage rights for the following namespace(s): {namespacesWithoutManageRightsJoined}. Execution will continue and assume the topology is already created.");
                 return;
             }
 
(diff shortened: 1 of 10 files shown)


SeanFeldman:

src/Transport/Topology/TopologyCreator.cs
```diff
@@ -19,10 +19,11 @@ public TopologyCreator(ICreateAzureServiceBusSubscriptionsInternal subscriptions
             this.queuesCreator = queuesCreator;
             this.subscriptionsCreator = subscriptionsCreator;
             this.namespaces = namespaces;
+            var namespaceConfigurations = settings.Get<NamespaceConfigurations>(WellKnownConfigurationKeys.Topology.Addressing.Namespaces);
```
yvesgoeleven: Are we certain this constructor is called after settings are complete?

src/Transport/Topology/TopologyOperator.cs
```diff
@@ -5,15 +5,21 @@ namespace NServiceBus.Transport.AzureServiceBus
...
         {
-            readOnlySettings = settings;
             this.brokeredMessageConverter = brokeredMessageConverter;
             messageReceiverLifeCycle = clientEntities;
+
+            messageReceiverNotifierSettings = new MessageReceiverNotifierSettings(
+                settings.Get<ReceiveMode>(WellKnownConfigurationKeys.Connectivity.MessageReceivers.ReceiveMode),
+                settings.HasExplicitValue<TransportTransactionMode>() ? settings.Get<TransportTransactionMode>() : settings.SupportedTransactionMode(),
+                settings.Get<TimeSpan>(WellKnownConfigurationKeys.Connectivity.MessageReceivers.AutoRenewTimeout),
+                settings.Get<int>(WellKnownConfigurationKeys.Connectivity.NumberOfClientsPerEntity));
```
yvesgoeleven: is it guaranteed that settings are complete when this constructor is called?

SeanFeldman:
Re-based to include changes from #607

danielmarbach:
Bit on the fence if it really requires a decorator class for 4 values but also not against it

SeanFeldman:
@danielmarbach my logic is simple: having a bunch of variables around vs having an explicit concept of `MessageReceiverNotifierSettings`, I prefer an explicit concept. For example, when `TopologyOperator` passes those into `MessageReceiveNotifier`, having `messageReceiverNotifierSettings` reduces the need to remember which variables / settings are intended to be [forwarded to `MessageReceiveNotifier`](https://github.com/Particular/NServiceBus.AzureServiceBus/blob/62dc88b362b7dd3190139ac426cbbdaac69f7f75/src/Transport/Topology/TopologyOperator.cs#L18).

src/Transport/Topology/TopologyOperator.cs
SeanFeldman: I've confirmed that the settings defined during configuration time for an endpoint are the settings passed into TopologyOperator and MessageRecieveNotifier later @yvesgoeleven. Here's the chain of calls:

`MessagePump.Start()` -> `TopologyOperator.Start()` -> `TopologyOperator.StartNotifiers()` -> `TopologyOperator.CreateNotifier()` -> `MessageReceiverNotifier.ctor(notifierSettings)`

`MessagePump.Init()` -> `MessagePump.DetermineTopologyOperator()` -> `TopologyOperator.ctor(settings)`

Settings passed into `MessagePump` ctor are already `ReadOnlySettings`, indicating that configuration stage is finished.

I.e. an endpoint will not be running and using settings until  endpoint configuration is finalized and endpoint is started (message pump is initialized and started).

src/Transport/Topology/TopologyCreator.cs
SeanFeldman: Same answer as [here](https://github.com/Particular/NServiceBus.AzureServiceBus/pull/608#discussion_r135875073) will apply. Configuration object is finalized **prior** to endpoint running. Meaning message pump (which kicks off topology operator) will not be invoked before user-provided configuration  is captured.

SeanFeldman:
@yvesgoeleven I've verified the two items.
@Particular/azure-maintainers are we good to merge this?

SeanFeldman:
Once this one is merged, #282 can be closed as completed.

SeanFeldman:
@yvesgoeleven since you had some concerns, could you please chime in?

Ready to be merged.

PR #615, opened 2017-09-18
Remove default serializer

boblangley:
Removes the default serializer from the transport and requires the user to set a serializer.


src/Transport/Utils/Guard.cs
```diff
@@ -35,5 +37,13 @@ public static void AgainstNegative(string argumentName, TimeSpan value)
                 throw new ArgumentOutOfRangeException(argumentName);
             }
         }
+
+        public static void AgainstUnsetSerializerSetting(SettingsHolder settings)
+        {
+            if (!settings.TryGet<Tuple<SerializationDefinition, SettingsHolder>>(WellKnownConfigurationKeys.Core.MainSerializerSettingsKey, out var _))
+            {
+                throw new Exception("Due to message size restrictions setting a serializer is mandatory. Use 'endpointConfiguration.UseSerialization<T>();' to select a serializer. If you are upgrading, install the `NServiceBus.Newtonsoft.Json` NuGet package and consult the upgrade guide for further information.");
```
SeanFeldman: > Due to message size restrictions setting a serializer is mandatory

Is this accurate? We're not forcing seializer selection because of the message size. We're going away from internally defaulting to JSON.

danielmarbach: According to @yvesgoeleven message size was the primary reason why we went with Json by default. So yes I think it is accurate but maybe it is too detailed?

SeanFeldman: That's the reason we chose JSON serializer. But that's not the reason why a serializer is mandatory, right? 🙂 

I.e. what I'm saying is that we're making this setting mandatory. Why are we doing so? We also instruct people to use NServiceBus.Newtonsoft.Json if they are upgrading. 
As a new custoemr, since I don't upgrade, I get to choose any serializer?

SeanFeldman: To clarify, I suggest to remove the "Due to message size restrictions".

boblangley: If we remove it we will also need to remove the same message from ASQ. No objections here.

SeanFeldman: 👍 don't forget to add this change to the ASB's upgrade guide. I've covered the ASQ 🙂

boblangley:
Doco PR for this change https://github.com/Particular/docs.particular.net/pull/3240

SeanFeldman:
Merged.

PR #618, opened 2017-09-22
New core ATT requires publisher registration

SeanFeldman:
@yvesgoeleven 
This will fix the `When_publishing_with_overridden_local_address` failing test on `EndpointOrientedTopology`.

### DIFF
--- a/src/AcceptanceTests/ConfigureEndpointAzureServiceBusTransport.cs
+++ b/src/AcceptanceTests/ConfigureEndpointAzureServiceBusTransport.cs
@@ -2,6 +2,7 @@
 using NServiceBus;
 using NServiceBus.AcceptanceTesting.Support;
 using NServiceBus.AcceptanceTests.DelayedDelivery;
+using NServiceBus.AcceptanceTests.Routing;
 using TesingConventions = NServiceBus.AcceptanceTesting.Customization;
 using NServiceBus.AcceptanceTests.Routing.NativePublishSubscribe;
 using NServiceBus.AcceptanceTests.Sagas;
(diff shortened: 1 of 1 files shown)


src/AcceptanceTests/ConfigureEndpointAzureServiceBusTransport.cs
```diff
@@ -82,6 +83,8 @@ public Task Configure(string endpointName, EndpointConfiguration configuration,
             endpointOrientedTopology.RegisterPublisher(typeof(When_publishing_from_sendonly.MyEvent), TesingConventions.Conventions.EndpointNamingConvention(typeof(When_publishing_from_sendonly.SendOnlyPublisher)));
             endpointOrientedTopology.RegisterPublisher(typeof(When_subscribing_to_a_base_event.IBaseEvent), TesingConventions.Conventions.EndpointNamingConvention(typeof(When_subscribing_to_a_base_event.Publisher)));
             endpointOrientedTopology.RegisterPublisher(typeof(When_subscribing_to_a_derived_event.SpecificEvent), TesingConventions.Conventions.EndpointNamingConvention(typeof(When_subscribing_to_a_derived_event.Publisher)));
+
+            endpointOrientedTopology.RegisterPublisher(typeof(When_publishing_with_overridden_local_address.MyEvent), TesingConventions.Conventions.EndpointNamingConvention(typeof(When_publishing_with_overridden_local_address.Publisher)));
```
danielmarbach: So nothing we can or should address in the core?

SeanFeldman: Not really. Core has changed it's API back in v6 and sealed publishers info. Therefore whenever a pub/sub ATT is added to the suite, we have to register publisher for the EndpointOriented topology.

danielmarbach: Can't remember why some of them don't use the metadata https://github.com/Particular/NServiceBus.AzureServiceBus/blob/8f9aadf9f5f771734b53c840a35f906c948881e9/src/AcceptanceTests/ConfigureEndpointAzureServiceBusTransport.cs#L48

PR #620, opened 2017-09-22
Using core LocalAddress abstraction

### LINKED ISSUE #619: Support for overridden local address is broken
The test When_publishing_with_overridden_local_address, overrides the local address 

`builder.OverrideLocalAddress("myinputqueue");`

it creates the following topology

image

Note how the transport creates 2 input queues

one for the endpoint name and one for the overriden local address
the subscription has a forwarding rule set up  to route to the endpoint named one, and not the overridden local address

SeanFeldman:
Connects to #619 
Fixes #619

When an endpoint name is overridden, endpoint's original input queue shouldn't be used.
Instead, the override name should.

For example, with 2 endpoints, `Publisher` and `Subscriber` where `Subscriber` overrides endpoint name to be `myInputQueue` and subscribes to the publisher's event `MyEvent`. The following _should_ happen:

1.  Subscriber endpoint to create `myInputQueue` queue as its input queue.
1. `MyEvent` events published by Publisher to be received by Subscriber via `myInputQueue`.
1. `Subscriber` queue to **not** exist

Subscriber's endpoint name should be available via `settings.LocalAddress`. Problem with this extension method is that the transport upon Initialization constructs topology that _caches_ endpoint name that is **not finalized by the core yet**.  Therefore, endpoint name should not be cached, but instead `settings.LocalAddress` be used when 
- `SubscriptionManager` needs to create a subscription  ==> `ITopologyInternal.GetSubscriptionManagerFactory()`
- or `TransportResourcesCreator` needs to create an input queue ==> `ITopologyInternal.GetQueueCreatorFactory()`

### DIFF
--- a/src/Transport/Sending/Batcher.cs
+++ b/src/Transport/Sending/Batcher.cs
@@ -1,5 +1,6 @@
 namespace NServiceBus.Transport.AzureServiceBus
 {
+    using System;
     using System.Collections.Generic;
     using System.Linq;
     using Settings;
@@ -10,6 +11,7 @@ namespace NServiceBus.Transport.AzureServiceBus
         {
             this.topologySectionManager = topologySectionManager;
             messageSizePaddingPercentage = settings.Get<int>(WellKnownConfigurationKeys.Connectivity.MessageSenders.MessageSizePaddingPercentage);
+            localAddress = new Lazy<string>(() => settings.LocalAddress());
         }
 
         public IList<BatchInternal> ToBatches(TransportOperations operations)
@@ -51,7 +53,7 @@ namespace NServiceBus.Transport.AzureServiceBus
                     batch = new BatchInternal();
                     indexedBatches[key] = batch;
 
-                    batch.Destinations = topologySectionManager.DeterminePublishDestination(multicastOperation.MessageType);
+                    batch.Destinations = topologySectionManager.DeterminePublishDestination(multicastOperation.MessageType, localAddress.Value);
                     batch.RequiredDispatchConsistency = multicastOperation.RequiredDispatchConsistency;
                 }
                 batch.Operations.Add(new BatchedOperationInternal(messageSizePaddingPercentage)
@@ -64,5 +66,6 @@ namespace NServiceBus.Transport.AzureServiceBus
 
         ITopologySectionManagerInternal topologySectionManager;
         int messageSizePaddingPercentage;
+        Lazy<string> localAddress;
     }
 }
--- a/src/Transport/Sending/Batcher.cs
+++ b/src/Transport/Sending/Batcher.cs
@@ -25,8 +25,7 @@ namespace NServiceBus.Transport.AzureServiceBus
             foreach (var unicastOperation in operations.UnicastTransportOperations)
             {
                 var key = $"unicast-{unicastOperation.Destination}-consistency-{unicastOperation.RequiredDispatchConsistency}";
-                BatchInternal batch;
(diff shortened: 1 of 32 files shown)


SeanFeldman:
With the changes in this PR we get a single queue for the subscriber, subscriber's ASB subscription is correct, and metadata is using endpoint original name rather than override.
image

SeanFeldman:
Works for Forwarding topology as well.

image

SeanFeldman:
Edge case I'm not sure ATT for is written correctly: endpoint that overrides its name and self subscribing/publishing.

SeanFeldman:
Don not merge.
TC is not showing all checks... only for the forwarding topology. Confirmation test is red.

SeanFeldman:
- I've introduced `When_publishing_and_subscribing_to_self_with_overridden_address` to verify a case for publisher and subscriber being the same endpoint. Figured that in such a case, we can't figure what's the override of the publisher. Therefore with the EndpointOrientedTopology a registered publisher should be [registered by it's override name](https://github.com/Particular/NServiceBus.AzureServiceBus/pull/620/files#diff-ba2d9535a94c34a9c45553208d6c4856R88) and not the endpoint name.
- Had to use settings at runtime for `Batcher` when determining publishing destination (topic). To ensure Daniel doesn't fly over the pond and spank me, I've [wrapped it with Lazy](https://github.com/Particular/NServiceBus.AzureServiceBus/pull/620/files#diff-5065bc18ab8716722213d3d042591dafR14).

SeanFeldman:

SeanFeldman:
Rebased on top of the latest `release-8.0`

SeanFeldman:
Build is green. If PR is approved, do not hesitate to merge.

danielmarbach:
I will rebase this

src/Transport/Sending/Batcher.cs
```diff
@@ -10,6 +11,7 @@ public Batcher(ITopologySectionManagerInternal topologySectionManager, ReadOnlyS
         {
             this.topologySectionManager = topologySectionManager;
             messageSizePaddingPercentage = settings.Get<int>(WellKnownConfigurationKeys.Connectivity.MessageSenders.MessageSizePaddingPercentage);
+            localAddress = new Lazy<string>(() => settings.LocalAddress());
```
danielmarbach: I'm ok with that although it smells like our whole initialization chain in the transport infrastructure should be overhauled. Not important enough though to hold up the PR

danielmarbach:
@SeanFeldman heads up I rebased this to get it compiling again

src/Transport/Sending/Batcher.cs
SeanFeldman: The constructor is invoked way too early and LocalAddress is not available. If there's a better way that would not require massive refactoring, I'm up for a separate PR where we could experiment.

danielmarbach:
@SeanFeldman I move the batcher instantiation to the dispatcher creation. What that no lazy is needed.

danielmarbach:
Ultimately the problems we are facing are because we are using the transport infrastructure improperly. In the initialization phase, we create the topology and then extract everything. That only works for our own settings. A better place would be to extract it during the start of the infrastructure. I will see if I can send in a PR to simplify things after this has been merged

SeanFeldman:
[That's](https://github.com/Particular/NServiceBus.AzureServiceBus/pull/620#issuecomment-332429270) exactly what it is @danielmarbach

PR #621, opened 2017-09-25
[SPIKE] Self subscription with overridden queue name 

SeanFeldman:
Connects to #620

Do **NOT** merge.


SeanFeldman:
Not needed

Issue #585, opened 2017-07-04
Send to another namespace cumbersome


yvesgoeleven:
Now that endpoint mappings is gone, ever send that occurs to an endpoint another namespace needs to pass in the namespace using `SendOptions`, there is no mapping anymore between type & namespace

Relates to https://github.com/Particular/NServiceBus.AzureServiceBus/issues/586

yvesgoeleven:
MessageEndpointMappings still exist, so can serve as workaround until we can add a new api for this

SeanFeldman:
`MessageEndpointMappings` is deprecated in NSB v7.

yvesgoeleven:
suggested API:

Additional overloads that accept a connectionstring

transport.Routing().RouteToEndpoint();

danielmarbach:
@yvesgoeleven I'm tasked to get this started. I will start with it today or tomorrow depending on my schedule.

yvesgoeleven:
@danielmarbach cool, I'll help as well

yvesgoeleven:
Been discussing this capability with @danielmarbach and want to discuss it further during our sync today. Some concerns and thoughts have been raised that will influence the design (so ignore my api suggestion above)

concerns:

We should not blend logical & physical routing, RouteToEndpoint is at the logical level and adding a connectionstring to it would blend this. Maybe better is to have a config method that allows user to specify that a logical endpoint, maps to a connectionstring outside the topo. e.g;
  `transport.NamespaceRouting().EndpointIsLocatedIn(endpointname, connectionstring or alias)`

Should this be considered a bug? or a new feature?

yvesgoeleven:
Implementation wise, I think looking up the endpoint to destination mapping belongs here (and not inside the topo)

https://github.com/Particular/NServiceBus.AzureServiceBus/blob/develop/src/Transport/Sending/Batcher.cs#L34

yvesgoeleven:
It was decided to go forward with send as described above, should be fairly easy to implement

danielmarbach:
Closed via https://github.com/Particular/NServiceBus.AzureServiceBus/pull/623

yvesgoeleven:
Maybe add the send to the example after as well?

endpointInstance.Send(destination: "sales",   message: new MyMessage());

danielmarbach:
Done

SeanFeldman:
> Maybe add the send to the example after as well?

Are you referring to a sample?!

Issue #584, opened 2017-07-04
Subscribe on other namespace is broken


yvesgoeleven:
Relates to https://github.com/Particular/NServiceBus.AzureServiceBus/issues/586

Looks like it is no longer possible to subscribe on an endpoint in a different namespace (one not part of topology), now that endpoint mappings are gone

**Original issue**

[GG post](https://groups.google.com/forum/#!topic/particularsoftware/9XK0kL5YuJc)

> I have have an azure cloud service(worker role) endpoint. This is worker role listens to 2 ASB namespaces N1 and N2. Events E1, E2, E3 are published by another component on both the namespaces N1 and N2. Is there a way to for NSB to listen to events E1 and E2 in ASB Namespace N1 and listed to Event E3 in ASB Namespace N2? Currently it is listening to events from both the namespaces.

> This was possible in NSB 4.7.6 and the control was through EndpointMappings. So this feature is no possible with NSB 6.0?

TODO:
 - [ ] Documentation update

HaripraghashSubramaniam:
May I please know how this is being prioritized at the moment?

SeanFeldman:
@HaripraghashSubramaniam we'll start investigation this week to identify at what point this feature was no longer supported.

danielmarbach:
Hi @HaripraghashSubramaniam 

It is still possible to listen to events E1 and E2 in ASB Namespace N1 and listed to Event E3 in ASB Namespace N2 by doing the following:

The subscriber endpoint which wants to listen to E1 and E2 could add a custom namespace partitioning strategy like described in [Custom ABS Namespace Partitioning](https://docs.particular.net/samples/azure/custom-partitioning-asb/) documentation page.

The subscriber can then use the strategy in the following way:

```
        var namespacePartitioning = transport.NamespacePartitioning();
        namespacePartitioning.AddNamespace("namespace1", "Endpoint=sb://namespace1");
        namespacePartitioning.AddNamespace("namespace2", "Endpoint=sb://namespace2");
        namespacePartitioning.UseStrategy<ReplicatedNamespacePartitioningStrategy>();
```

then it could declare handlers that receive events from both namespaces.

```
public class EventFromNamespace1Handler :
    IHandleMessages<EventFromNamespace1>
{
    static ILog log = LogManager.GetLogger<EventFromNamespace1Handler>();

    public Task Handle(EventFromNamespace1 message, IMessageHandlerContext context)
    {
[...]

PR #622, opened 2017-09-26
RegisteredEndpoints can be added to the namespace info

danielmarbach:
API draft to solve #585 and #584 

- `AddNamespace `on routing will return a `NamespaceInfo  `since that class is already public
- `NamespaceInfo `has been extended to have a property called `RegisteredEndpoints `which allows indicating that an endpoint belongs to the specific namespace
- Uses a HashSet with ordinal ignore case comparison to avoid having duplicate entries in the set


SeanFeldman:
@danielmarbach `RegisteredEndpoints` was added to the `NamespaceInfo` and it feels as an enabler/prerequisite for the follow-up PRs. For such PRs, it would help reviewers to understand better the changes if a little of context is provided in the general description. Thanks.

SeanFeldman:
We'll need to add that to the v8 doco.

danielmarbach:
I can handle the doco tomorrow

PR #623, opened 2017-09-26
Send over namespaces

### LINKED ISSUE #585: Send to another namespace cumbersome

danielmarbach:
Based on #622 
Closes #585 
Doco https://github.com/Particular/docs.particular.net/pull/3276

`DetermineSendDestination` is the central location where either explicit destinations with alias or connection string or endpoint names will go through. The routing has been extended to look up whether a registered endpoint is available and then use that as a destination. If no registered endpoint is available the normal partitioning strategy approach will be used. 

The assumptions are that if a destination contains `@alias` or `@connectionstring` the `inputQueueAddress` will have the properties set as usual and this takes precedence. For all other destinations without `@` the else part is responsible.

### DIFF
--- a/src/AcceptanceTests/Routing/When_sending_to_specific_namespace.cs
+++ b/src/AcceptanceTests/Routing/When_sending_to_specific_namespace.cs
@@ -26,7 +26,7 @@ namespace NServiceBus.Azure.Transports.WindowsAzureServiceBus.AcceptanceTests.Ro
                 var targetConnectionString = EnvironmentHelper.GetEnvironmentVariable("AzureServiceBus.ConnectionString.Fallback");
                 extensions.UseNamespaceAliasesInsteadOfConnectionStrings();
 
-                if (endpointName == "UsingMultipleNamespaces.EndpointInTargetNamespace")
+                if (endpointName == Conventions.EndpointNamingConvention(typeof(EndpointInTargetNamespace)))
                 {
                     extensions.NamespaceRouting().AddNamespace("source", connectionString);
                     extensions.NamespacePartitioning().AddNamespace("target", targetConnectionString);
@@ -71,7 +71,7 @@ namespace NServiceBus.Azure.Transports.WindowsAzureServiceBus.AcceptanceTests.Ro
             var ctx = new AzureServiceBusTransportConfigContext();
             ctx.Callback = (endpointName, extensions) =>
             {
-                   if (endpointName == "UsingMultipleNamespaces.EndpointInTargetNamespace")
+                   if (endpointName == Conventions.EndpointNamingConvention(typeof(EndpointInTargetNamespace)))
                 {
                     extensions.NamespaceRouting().AddNamespace("source", connectionString);
                     extensions.NamespacePartitioning().AddNamespace("target", targetConnectionString);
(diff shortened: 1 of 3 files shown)


src/AcceptanceTests/Routing/When_sending_to_specific_namespace.cs
```diff
@@ -103,6 +103,46 @@ public class When_using_multiple_namespaces : NServiceBusAcceptanceTest
...
+            var runSettings = new RunSettings();
+            runSettings.TestExecutionTimeout = TimeSpan.FromMinutes(1);
+
+            var connectionString = Environment.GetEnvironmentVariable("AzureServiceBusTransport.ConnectionString");
+            var targetConnectionString = Environment.GetEnvironmentVariable("AzureServiceBus.ConnectionString.Fallback");
+
+            var ctx = new AzureServiceBusTransportConfigContext();
+            ctx.Callback = (endpointName, extensions) =>
+            {
+                if (endpointName == Conventions.EndpointNamingConvention(typeof(EndpointInTargetNamespace)))
```
SeanFeldman: Personally I would flip if/else to emphasize that we're testing sending (routing) outside of the source namespace via `NamespaceRouting()` API. Not critical though.

src/Transport/Topology/Topologies/EndpointOrientedTopologySectionManager.cs
```diff
@@ -87,71 +87,87 @@ public TopologySectionInternal DetermineResourcesToCreate(QueueBindings queueBin
...
-                namespaces = namespacePartitioningStrategy.GetNamespaces(PartitioningIntent.Sending).ToArray();
-            }
 
-            if (namespaces == null)
-            {
-                throw new Exception($"Could not determine namespace for destination {destination}");
-            }
+                if (namespaces == null)
+                {
+                    throw new Exception($"Could not determine namespace for destination {d}");
```
SeanFeldman: This was done incorrectly in the past, but should we consider escaping the value with single quotes? Doesn't have to be this PR at all, just a question 🙂

src/Transport/Topology/Topologies/ForwardingTopologySectionManager.cs
```diff
@@ -130,14 +130,25 @@ public TopologySectionInternal DetermineSendDestination(string destination)
...
+                                new RuntimeNamespaceInfo(configured.Alias, configured.Connection, configured.Purpose)
                             };
                         }
                     }
                 }
-                else // sending to the partition
+                else
                 {
-                    namespaces = namespacePartitioningStrategy.GetNamespaces(PartitioningIntent.Sending).ToArray();
+                    var configured = namespaceConfigurations.FirstOrDefault(n => n.RegisteredEndpoints.Contains(d, StringComparer.OrdinalIgnoreCase));
```
SeanFeldman: `configuredToRouteToAnotherNamespace`? everything is configured at some point 😃

src/AcceptanceTests/Routing/When_sending_to_specific_namespace.cs
danielmarbach: I kept it aligned with the other tests. I think I'll leave it

src/Transport/Topology/Topologies/EndpointOrientedTopologySectionManager.cs
danielmarbach: Will do

src/Transport/Topology/Topologies/ForwardingTopologySectionManager.cs
danielmarbach: will change

danielmarbach:
@SeanFeldman rebased and addressed your comments

SeanFeldman:
Once build is green, can merge

danielmarbach:
Added the [doco PR](https://github.com/Particular/docs.particular.net/pull/3276) to the description

PR #624, opened 2017-09-26
Subscribe outside topology

yvesgoeleven:
This PR fixes cross namespace subscriptions in the endpoint oriented topology, https://github.com/Particular/NServiceBus.AzureServiceBus/issues/584

It builds on the new api, https://github.com/Particular/NServiceBus.AzureServiceBus/pull/622, that allows to register an endpoint in a namespace outside the topology using

```
extensions.NamespaceRouting()
        .AddNamespace("publisherNamespace", publisherConnectionString)
        .RegisteredEndpoints.Add(AcceptanceTesting.Customization.Conventions.EndpointNamingConvention(typeof(Publisher)));
```

and takes this info into account when creating subscriptions.

A test, `When_subscribing_outside_the_endpoint_oriented_topology`, was added to proof that it works

I had to change the ATT's a little bit as the configure class did overwrite whatever topology was set in the configuration callback, that's why a few other tests have slightly changed as well

**TODO**
- [ ] documentation

### DIFF
--- a/src/AcceptanceTests/Routing/When_subscribing_outside_the_endpoint_oriented_topology.cs
+++ b/src/AcceptanceTests/Routing/When_subscribing_outside_the_endpoint_oriented_topology.cs
@@ -4,12 +4,10 @@
     using System.Threading.Tasks;
     using AcceptanceTesting;
     using AcceptanceTesting.Support;
-    using AzureServiceBus;
     using NServiceBus.AcceptanceTests;
     using NServiceBus.AcceptanceTests.EndpointTemplates;
     using Features;
     using NServiceBus.AcceptanceTests.ScenarioDescriptors;
-    using Settings;
     using NUnit.Framework;
 
     public class When_subscribing_outside_the_endpoint_oriented_topology : NServiceBusAcceptanceTest
@@ -63,7 +61,7 @@
 
                     c.ReceiverSubscribedToEvents = true;
                 }))
-                .Done(ctx => ctx.SubscriberGotTheEvent || !ctx.IsEndpointOrientedTopology)
+                .Done(ctx => ctx.SubscriberGotTheEvent)
                 .Run(runSettings);
 
             Assert.That(context.SubscriberGotTheEvent, Is.True, "Should receive the event");
@@ -72,7 +70,6 @@
         public class Context : ScenarioContext
         {
             public bool SubscriberGotTheEvent { get; set; }
-            public bool IsEndpointOrientedTopology { get; set; }
             public bool ReceiverSubscribedToEvents { get; set; }
         }
 
@@ -82,41 +79,7 @@
 
             public Publisher()
             {
-                EndpointSetup<DefaultPublisher>(endpointConfiguration =>
-                {
-                    endpointConfiguration.EnableFeature<DetermineWhatTopologyIsUsed>();
-                });
-            }
-
-            class DetermineWhatTopologyIsUsed : Feature
-            {
-                protected override void Setup(FeatureConfigurationContext context)
-                {
-                    context.RegisterStartupTask(builder => new TaskToDetermineCurrentTopology(builder.Build<Context>(), builder.Build<ReadOnlySettings>()));
-                }
-            }
-
-            class TaskToDetermineCurrentTopology : FeatureStartupTask
-            {
-                Context context;
-                ReadOnlySettings settings;
-
-                public TaskToDetermineCurrentTopology(Context context, ReadOnlySettings settings)
-                {
-                    this.context = context;
-                    this.settings = settings;
-                }
-
-                protected override Task OnStart(IMessageSession session)
-                {
-                    context.IsEndpointOrientedTopology = settings.Get<string>("AzureServiceBus.AcceptanceTests.UsedTopology") == "EndpointOrientedTopology";
-                    return TaskEx.Completed;
-                }
-
-                protected override Task OnStop(IMessageSession session)
-                {
-                    return TaskEx.Completed;
-                }
+                EndpointSetup<DefaultPublisher>();
             }
         }
 
--- a/src/AcceptanceTests/Routing/When_subscribing_outside_the_endpoint_oriented_topology.cs
+++ b/src/AcceptanceTests/Routing/When_subscribing_outside_the_endpoint_oriented_topology.cs
@@ -17,6 +17,12 @@
         [Test]
         public async Task Should_receive_event()
         {
+            var topology = EnvironmentHelper.GetEnvironmentVariable("AzureServiceBusTransport.Topology");
+            if (topology == "ForwardingTopology")
+            {
+                Assert.Inconclusive("The test is designed for EndpointOrientedTopology only.");
+            }
+
             var runSettings = new RunSettings();
             runSettings.TestExecutionTimeout = TimeSpan.FromMinutes(1);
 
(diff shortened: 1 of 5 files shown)


src/Transport/Topology/Topologies/EndpointOrientedTopologySectionManager.cs
```diff
@@ -193,15 +193,30 @@ TopologySectionInternal BuildSubscriptionHierarchy(Type eventType)
...
             var topics = new List<EntityInfoInternal>();
             var subs = new List<SubscriptionInfoInternal>();
-            foreach (var topicPath in topicPaths)
+            foreach (var publisher in publishers)
             {
+                var topicPath = publisher + ".events";
                 var path = addressingLogic.Apply(topicPath, EntityType.Topic).Name;
-                topics.AddRange(namespaces.Select(ns => new EntityInfoInternal
+
+                var destinationsOutsideTopology = namespaceConfigurations.Where(c => c.RegisteredEndpoints.Contains(publisher)).ToList();
```
danielmarbach: I used `RegisteredEndpoints.Contains(d, StringComparer.OrdinalIgnoreCase)`

src/Transport/Topology/Topologies/EndpointOrientedTopologySectionManager.cs
```diff
@@ -193,15 +193,30 @@ TopologySectionInternal BuildSubscriptionHierarchy(Type eventType)
...
+                if (destinationsOutsideTopology.Any())
                 {
-                    Namespace = ns,
-                    Type = EntityType.Topic,
-                    Path = path
-                }));
+                    topics.AddRange(destinationsOutsideTopology.Select(ns => new EntityInfoInternal
```
danielmarbach: Can we add an acceptance test?

yvesgoeleven: Added a test

src/Transport/Topology/Topologies/EndpointOrientedTopologySectionManager.cs
yvesgoeleven: makes sense, I'll change it

SeanFeldman:
@yvesgoeleven could you add general description that provides the context of this PR? 
image

Description should not be left out blank and should help reviewers understand the context of the change. Thanks.

src/AcceptanceTests/Routing/When_subscribing_outside_the_endpoint_oriented_topology.cs
```diff
@@ -0,0 +1,113 @@
...
+            var runSettings = new RunSettings();
+            runSettings.TestExecutionTimeout = TimeSpan.FromMinutes(1);
+
+            var config = new AzureServiceBusTransportConfigContext();
+            config.Callback = (endpointName, extensions) =>
+            {
+                var publisherConnectionString = EnvironmentHelper.GetEnvironmentVariable("AzureServiceBusTransport.ConnectionString");
+                var subscriberConnectionString = EnvironmentHelper.GetEnvironmentVariable("AzureServiceBus.ConnectionString.Fallback");
+
+                if (endpointName == "SubscribingOutsideTheEndpointOrientedTopology.Subscriber")
```
SeanFeldman: We shouldn't use hard-coded name. Naming convention can be used
`TesingConventions.Conventions.EndpointNamingConvention(typeof(SubscribingOutsideTheEndpointOrientedTopology.Subscriber))`

src/AcceptanceTests/Routing/When_subscribing_outside_the_endpoint_oriented_topology.cs
```diff
@@ -0,0 +1,113 @@
...
+                {
+
+                    extensions.NamespacePartitioning().AddNamespace("subscriberNamespace", subscriberConnectionString);
+                    extensions.NamespaceRouting()
+                                    .AddNamespace("publisherNamespace", publisherConnectionString)
+                                    .RegisteredEndpoints.Add(AcceptanceTesting.Customization.Conventions.EndpointNamingConvention(typeof(Publisher)));
```
SeanFeldman: Would suggest to take the namespaceInfo returned by `extensions.NamespaceRouting().AddNamespace()` and assign to a variable used along with `.RegisteredEndpoints.Add()`. Fluent API while correct, harder to digest.

src/AcceptanceTests/Routing/When_subscribing_outside_the_endpoint_oriented_topology.cs
```diff
@@ -0,0 +1,113 @@
...
+                    extensions.UseEndpointOrientedTopology()
+                        .RegisterPublisher(typeof(MyEvent), AcceptanceTesting.Customization.Conventions.EndpointNamingConvention(typeof(Publisher)));
+                }
+                else
+                {
+                    extensions.NamespacePartitioning().AddNamespace("publisherNamespace", publisherConnectionString);
+                    extensions.UseEndpointOrientedTopology();
```
SeanFeldman: Is this necessary? It can't be ForwardingTopology or else we'd be out of this test by no thanks to [this check](https://github.com/Particular/NServiceBus.AzureServiceBus/pull/624/files?diff=split#diff-73b5a886317cfcbfd70a634fe7937825R21).

src/AcceptanceTests/Routing/When_subscribing_outside_the_endpoint_oriented_topology.cs
```diff
@@ -0,0 +1,113 @@
...
+                }
+            };
+
+            runSettings.Set("AzureServiceBus.AcceptanceTests.TransportConfigContext", config);
+
+            var context = await Scenario.Define<Context>()
+                // Publish event only when it was signaled that events went out
+                .WithEndpoint<Publisher>(b => b.When(c => c.ReceiverSubscribedToEvents, async bus =>
```
SeanFeldman: nitpicking: `bus` -> `endpoint` 🙂

src/Transport/Topology/Topologies/EndpointOrientedTopologySectionManager.cs
```diff
@@ -193,39 +193,81 @@ TopologySectionInternal BuildSubscriptionHierarchy(Type eventType)
...
             var topics = new List<EntityInfoInternal>();
             var subs = new List<SubscriptionInfoInternal>();
-            foreach (var topicPath in topicPaths)
+            foreach (var publisher in publishers)
             {
+                var topicPath = publisher + ".events";
                 var path = addressingLogic.Apply(topicPath, EntityType.Topic).Name;
-                topics.AddRange(namespaces.Select(ns => new EntityInfoInternal
+
+                var destinationsOutsideTopology = namespaceConfigurations.Where(c => c.RegisteredEndpoints.Contains(publisher, StringComparer.OrdinalIgnoreCase)).ToList();
```
SeanFeldman: Is supplying `StringComparer.OrdinalIgnoreCase` necessary if `RegisteredEndpoints` defined as [`HashSet<string>(StringComparer.OrdinalIgnoreCase)`](https://github.com/Particular/NServiceBus.AzureServiceBus/pull/622/files#diff-95953b1908393b31d4f75e5479243552R23)?

SeanFeldman: `destinationsOutsideTopology` - "destinations" or "subscribers" outside of topology?

SeanFeldman:
This would need to be added to the v8 doco.

danielmarbach:
About https://github.com/Particular/NServiceBus.AzureServiceBus/pull/624#discussion_r141127197 if I recall correctly it specifying it in the ctor only means that the comparison while adding is case insensitive. But that still means the value can be stored in any casing (whatever came first). Therefore I was assuming we required it for the contains.

src/Transport/Topology/Topologies/EndpointOrientedTopologySectionManager.cs
SeanFeldman: nm, comparer at construction is only used of inserts.

SeanFeldman:
Rebased on top of the latest `release-8.0`.
Can this work be wrapped up?

PR #633, opened 2017-09-28
Turn two topologies into two distinct transport infrastructures

danielmarbach:
In order to fix the send only stuff I need an asynchronous entry point to create the topics. To achieve that I started refactoring the topologies.

What do you think about this approach?


src/Transport/Config/AzureServiceBusTransport.cs
```diff
@@ -27,30 +27,24 @@ public override TransportInfrastructure Initialize(SettingsHolder settings, stri
...
-            if (!settings.TryGet(WellKnownConfigurationKeys.Topology.Selected, out string configuredTopology))
-            {
-                throw new Exception("Azure Service Bus transport requires a topology to be specified. Use `.UseForwardingTopology()` or `.UseEndpointOrientedTopology()` configuration API to specify topology to use.");
-            }
+            settings.TryGet(WellKnownConfigurationKeys.Topology.Selected, out string configuredTopology);
             switch (configuredTopology)
             {
                 case WellKnownConfigurationKeys.Topology.EndpointOrientedTopology:
-                    return new EndpointOrientedTopologyInternal();
+                    return new EndpointOrientedTransportInfrastructure(settings);
```
SeanFeldman: This is in essence 2 transports 🙂

src/Transport/Config/AzureServiceBusTransportInfrastructure.cs
```diff
@@ -7,34 +7,81 @@
...
             typeof(DiscardIfNotReceivedBefore)
         };
 
+
         public override TransportTransactionMode TransactionMode { get; }
 
-        public override OutboundRoutingPolicy OutboundRoutingPolicy => topology.GetOutboundRoutingPolicy();
+        public override OutboundRoutingPolicy OutboundRoutingPolicy { get; } = new OutboundRoutingPolicy(OutboundRoutingType.Unicast, OutboundRoutingType.Multicast, OutboundRoutingType.Unicast);
+
+        // make thread
```
SeanFeldman: What do you mean by that?

src/Transport/Config/AzureServiceBusTransportInfrastructure.cs
```diff
@@ -7,34 +7,81 @@
...
+        void Initialize()
+        {
+            if (initialized)
```
SeanFeldman: Why would `Initialize()` be invoked more than once?

danielmarbach:
@SeanFeldman I think the timing issues should be resolved now. Apart from the still commented out code, this should be good for a first pass over.

SeanFeldman:
This does feel dirty 😉 
`When_publishing_from_sendonly.Should_be_delivered_to_all_subscribers` is failing.
Let's discuss this tomorrow.

src/Transport/Topology/Topologies/ForwardingTransportInfrastructure.cs
```diff
@@ -0,0 +1,47 @@
...
+                if (Interlocked.Exchange(ref initializeSignaled, 1) != 0)
+                {
+                    return;
+                }
+
+                var bundleConfigurations = await NumberOfTopicsInBundleCheck.Run(namespaceManager, namespaceConfigurations, bundlePrefix)
+                    .ConfigureAwait(false);
+
+                topologySectionManager.BundleConfigurations = bundleConfigurations;
+            };
```
yvesgoeleven: Why not just pass namespaceManager and namespaceConfigurations in the constructor and have this lambda initialized inside the section manager?

src/Tests/Seam/When_receiving_messages.cs
```diff
@@ -1,95 +1,95 @@
...
-            await pump.Stop();
-        }
-    }
-}
\ No newline at end of file
+//            // cleanup
+//            await pump.Stop();
+//        }
+//    }
+//}
```
yvesgoeleven: Why is this test commented out? Do you need to redo it? or does it need to be deleted?

yvesgoeleven:
I like it, lots of redundant code gone. Just a bit hesitant to take it in just before RC...

src/Transport/Topology/Topologies/ForwardingTransportInfrastructure.cs
danielmarbach: The section manager is not async. Previously we wrapped this call in a Lazy and executed it with GetAwaiter().GetResult(). In the past we had a discussion that the section manager could become async but if I recall correctly you rightfully said that building the topology should be possible local only without information from ASB itself. Therefore this bundle check does not belong into the manager.

src/Tests/Seam/When_receiving_messages.cs
danielmarbach: I need to make this compiling again but wanted to first hear your feedback before I proceed

src/Transport/Topology/Topologies/ForwardingTransportInfrastructure.cs
yvesgoeleven: Ah yes... add this comment in the code instead of //utter hack... so I wont forget next time I see it ;)

danielmarbach:
Updated and addressed comments

danielmarbach:
Forwarding Topology tests starting failing. Although locally they are working

image

image

danielmarbach:
The issue was caused by me returning the wrong subscription creator https://github.com/Particular/NServiceBus.AzureServiceBus/pull/633/commits/9763c14cac4c592b047ba8b8a4381dbf3a135b2d#diff-2e2967b39a0a225e68a8c464586e7de5L47

SeanFeldman:
@danielmarbach @boblangley I took the liberty to fix the failing bundles test in commit https://github.com/Particular/NServiceBus.AzureServiceBus/pull/633/commits/9fc815a447610335383aace857de935968957755

yvesgoeleven:
Send only issue will be covered by different PR

PR #635, opened 2017-10-02
Cleanup test to remove all queues and topics

### LINKED ISSUE #634: Perform namespace cleanup after ATTs executed
Add a single explicit test that can be triggered from TC build configuration step to with `dotnet test` at the end of ATTs steps to remove all ASB entities.

SeanFeldman:
Connects to #634 
Fixes #634 

Adding a single `explicit` test that should be triggered from TC build configuration step at the **after** ATTs step is done, to remove all queues and topics.

### ATT project

An explicit test called `DeleteEntities()` that asynchronously deletes all queues and topics on the main namespace with up to 5 retries per each entity.

### TC

image

1 - Additional (last) build step for EndpointOriented and Forwarding topology.
2 - Step is set to always execute .
3 - Cleaner is explicitly executed by selecting it using a filter ` --where "name==DeleteEntities && cat==Cleanup"`
4 - Assembly for Cleaner test is always `AcceptanceTests.dll`

### DIFF
--- a/src/AcceptanceTests/Infrastructure/Cleaner.cs
+++ b/src/AcceptanceTests/Infrastructure/Cleaner.cs
@@ -4,9 +4,9 @@
     using System.Linq;
     using System.Threading.Tasks;
     using Microsoft.ServiceBus;
-    using Microsoft.ServiceBus.Messaging;
     using NServiceBus.AcceptanceTests;
     using NUnit.Framework;
+    using Utils;
 
     public class Cleaner : NServiceBusAcceptanceTest
     {
@@ -37,9 +37,9 @@
                     .ConfigureAwait(false);
             }
             catch (Exception exception)
-                when (usedRetryAttempts < maxRetryAttempts && (exception is TimeoutException || exception is MessagingCommunicationException || exception is ServerBusyException))
+                when (usedRetryAttempts < maxRetryAttempts && (exception is TimeoutException || exception.IsTransientException()))
             {
-                TestContext.WriteLine($"Attempt to delete '{entityPath}' has failed. Trying attempt {usedRetryAttempts + 2}/{maxRetryAttempts} in 5 seconds.");
+                TestContext.WriteLine($"Attempt to delete '{entityPath}' has failed. Retrying attempt {usedRetryAttempts + 1}/{maxRetryAttempts} in 5 seconds.");
                 await Task.Delay(5000)
                     .ConfigureAwait(false);
                 await TryWithRetries(entityPath, task, maxRetryAttempts, usedRetryAttempts + 1)
--- a/src/AcceptanceTests/Infrastructure/Cleaner.cs
+++ b/src/AcceptanceTests/Infrastructure/Cleaner.cs
@@ -23,8 +23,8 @@
                 .ConfigureAwait(false);
 
             const int maxRetryAttempts = 5;
-            var deleteQueues = queryQueues.Result.Select(queueDescription => TryWithRetries(queueDescription.Path, namespaceManager.DeleteQueueAsync(queueDescription.Path), maxRetryAttempts));
-            var deleteTopics = queryTopics.Result.Select(topicDescription => TryWithRetries(topicDescription.Path, namespaceManager.DeleteTopicAsync(topicDescription.Path), maxRetryAttempts));
+            var deleteQueues = (await queryQueues).Select(queueDescription => TryWithRetries(queueDescription.Path, namespaceManager.DeleteQueueAsync(queueDescription.Path), maxRetryAttempts));
+            var deleteTopics = (await queryTopics).Select(topicDescription => TryWithRetries(topicDescription.Path, namespaceManager.DeleteTopicAsync(topicDescription.Path), maxRetryAttempts));
 
             await Task.WhenAll(deleteQueues.Concat(deleteTopics))
                 .ConfigureAwait(false);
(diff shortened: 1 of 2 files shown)


src/AcceptanceTests/Infrastructure/Cleaner.cs
```diff
@@ -0,0 +1,56 @@
...
+
+            var queryQueues = namespaceManager.GetQueuesAsync();
+            var queryTopics = namespaceManager.GetTopicsAsync();
+            await Task.WhenAll(queryQueues, queryTopics)
+                .ConfigureAwait(false);
+
+            var deleteQueues = queryQueues.Result.Select(queueDescription => namespaceManager.DeleteQueueAsync(queueDescription.Path));
+            var deleteTopics = queryTopics.Result.Select(topicDescription => namespaceManager.DeleteTopicAsync(topicDescription.Path));
+
+            await TryWithRetries(Task.WhenAll(deleteQueues.Concat(deleteTopics)), 5)
```
danielmarbach: Wouldn't we want to retry and individual queue or topic deletion operation instead of the whole thing?

src/AcceptanceTests/Infrastructure/Cleaner.cs
```diff
@@ -0,0 +1,56 @@
...
+        {
+            try
+            {
+                await task
+                    .ConfigureAwait(false);
+            }
+            catch (Exception exception)
+                when (usedRetryAttempts < maxRetryAttempts && (exception is TimeoutException || exception is MessagingCommunicationException || exception is ServerBusyException))
+            {
+                logger.InfoFormat("Attempt to delete all queues and topics has failed. Trying attempt {0}/{1} in 5 seconds.", usedRetryAttempts + 2, maxRetryAttempts);
```
danielmarbach: Why is it here `usedRetryAttempts + 2`, below `usedRetryAttempts + 1` and further down `usedRetryAttempts` that makes it hard to read and understand

src/AcceptanceTests/Infrastructure/Cleaner.cs
```diff
@@ -0,0 +1,56 @@
...
+                .ConfigureAwait(false);
+
+            var deleteQueues = queryQueues.Result.Select(queueDescription => namespaceManager.DeleteQueueAsync(queueDescription.Path));
+            var deleteTopics = queryTopics.Result.Select(topicDescription => namespaceManager.DeleteTopicAsync(topicDescription.Path));
+
+            await TryWithRetries(Task.WhenAll(deleteQueues.Concat(deleteTopics)), 5)
+                .ConfigureAwait(false);
+        }
+
+        static async Task TryWithRetries(Task task, int maxRetryAttempts, int usedRetryAttempts = 0)
```
danielmarbach: We have this helper method https://github.com/Particular/NServiceBus.AzureServiceBus/blob/db4a1d9ca44f59e6d95f47368e3716826abf4d14/src/Transport/Connectivity/TaskWithRetryExtensions.cs#L10 which might come in handy?

src/AcceptanceTests/Infrastructure/Cleaner.cs
danielmarbach: Doesn't the namespace manager have a retry policy built in that we could leverage?

SeanFeldman: I could prune everything that is completed, leaving failed and incomplete to be retried. But I wouldn't necessarily retry individually. We still want to execute in bulk.

SeanFeldman: It's there, but when it comes to exceptions propagated to the client, it won't help us. Throttling or timeouts for example. That's why we do our operations with retries as well.

src/AcceptanceTests/Infrastructure/Cleaner.cs
SeanFeldman: It's on `IMessageSender` which I don't need.

src/AcceptanceTests/Infrastructure/Cleaner.cs
SeanFeldman: When it's a first attempt, `usedRetryAttempts`  will be zero.
Writing "attempt 1/5" is meaningless. Next attempt is a second attempt, so it's "attempt 2/5 in 5 seconds". When we recursively invoke this method, only then `usedRetryAttempts` is increased and becomes 1. So you need to add 2 to get the message correct.

src/AcceptanceTests/Infrastructure/Cleaner.cs
danielmarbach: I should have been more clear. I meant have you looked at how this method is implemented and reused what you needed?

src/AcceptanceTests/Infrastructure/Cleaner.cs
danielmarbach: Ok makes sense

danielmarbach: Yeah but you would exactly get that. Everything that is successfully completed would be, everything that is not yet will retry or fail eventually. Things are still concurrent.

src/AcceptanceTests/Infrastructure/Cleaner.cs
SeanFeldman: I've used a similar pattern indeed.

SeanFeldman:
I was thinking to add fallback namespace cleanup as well, but it's just not worth the effort. Vast majority of the tests and usage is with the main connection string.

PR #639, opened 2017-10-03
Clear separation of queues and topic creation

### LINKED ISSUE #625: Send only endpoints do not create their publishing infrastructure
When_publishing_from_sendonly.Should_be_delivered_to_all_subscribers fails because it's topic is never created

danielmarbach:
Closes #625

Fixes SendOnly endpoint publish

Assuming manage rights:
- Queue creation happens during ressource creation phase
- Topic creation happens when the transport is started

### DIFF
--- a/src/Transport/Config/AzureServiceBusTransportInfrastructure.cs
+++ b/src/Transport/Config/AzureServiceBusTransportInfrastructure.cs
@@ -72,10 +72,13 @@
             topologyOperator = new TopologyOperator(messageReceiverLifeCycleManager, new BrokeredMessagesToIncomingMessagesConverter(Settings, new DefaultConnectionStringToNamespaceAliasMapper(Settings)), Settings);
         }
 
-        public override Task Start()
+        public override async Task Start()
         {
             InitializeIfNecessary();
-            return topologyManager.Initialize();
+
+            await topologyManager.Initialize().ConfigureAwait(false);
+            var topicsToCreate = topologyManager.DetermineTopicsToCreate(Settings.LocalAddress());
+            await topologyCreator.Create(topicsToCreate).ConfigureAwait(false);
         }
 
         public override Task Stop()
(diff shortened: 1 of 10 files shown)


src/Transport/Config/AzureServiceBusTransportInfrastructure.cs
```diff
@@ -72,10 +72,13 @@ void InitializeIfNecessary()
...
         }
 
-        public override Task Start()
+        public override async Task Start()
         {
             InitializeIfNecessary();
-            return topologyManager.Initialize();
+
+            await topologyManager.Initialize().ConfigureAwait(false);
+            var resourcesToCreate = topologyManager.DetermineTopicsToCreate(Settings.LocalAddress());
```
SeanFeldman: Nitpicking: topics are resources, but here the variable should probably be called `topicsToCreate`

src/Transport/Seam/TransportResourcesCreator.cs
```diff
@@ -21,7 +21,7 @@ public TransportResourcesCreator(TopologyCreator topologyCreator, ITopologySecti
             await topologyCreator.AssertManagedRights().ConfigureAwait(false);
 
             await sections.Initialize().ConfigureAwait(false);
-            var receiveResources = sections.DetermineResourcesToCreate(queueBindings, localAddress);
+            var receiveResources = sections.DetermineQueuesToCreate(queueBindings, localAddress);
```
SeanFeldman: same as above

src/Transport/Config/AzureServiceBusTransportInfrastructure.cs
danielmarbach: The method returns `TopologySectionInternal` so from that perspective the generic version is `resourcesToCreate` or `sectionToCreate` but I'm guessing being explicit does not hurt. I'll change it

danielmarbach:
@SeanFeldman pushed an update. Will merge once green

PR #640, opened 2017-10-04
Using addressingLogic to translate transport addresses

### LINKED ISSUE #627: ToTransportAddress creates queues incompatible with the queue name convention
Similar to https://github.com/Particular/NServiceBus.AzureStorageQueues/issues/261

## Who's affected

Customer's using Sender Side Distribution `DistributionContext.ToTransportAddress` or `TransportInfrastructure.ToTransportAddress` in conjunction with either the default sanitization or a custom sanitization strategy

Both classes are only used in very advanced use cases.

## Symptoms

When queue names are used with the default sanitization or a custom sanitization strategy and either `DistributionContext.ToTransportAddress` or `TransportInfrastructure.ToTransportAddress` is called the queue name returned from that method will not comply with the queue name conventions of the transport. 

## Analysis

Both the `ForwardingTopologySectionManager `and the `EndpointOrientedTopologySectionManager `use the [AddressingLogic ](https://github.com/Particular/NServiceBus.AzureServiceBus/blob/release-8.0/src/Transport/Addressing/AddressingLogic.cs) to apply both the composition and the sanitization strategy to the entity path. 

ToTransportAddress currently does not take the addressing logic into account. This can lead to problems for example with the Service Fabric partition affine routing example since it uses the distribution context to translate to transport addresses.

danielmarbach:
Closes #627 

When queue names are used with the default sanitization or a custom sanitization strategy and either `DistributionContext.ToTransportAddress` or `TransportInfrastructure.ToTransportAddress` is called the queue name returned from that method will not comply with the queue name conventions of the transport.

### DIFF
--- a/src/Transport/Config/AzureServiceBusTransportInfrastructure.cs
+++ b/src/Transport/Config/AzureServiceBusTransportInfrastructure.cs
@@ -20,6 +20,14 @@
 
             var individualizationStrategyType = (Type)Settings.Get(WellKnownConfigurationKeys.Topology.Addressing.Individualization.Strategy);
             individualization = individualizationStrategyType.CreateInstance<IIndividualizationStrategy>(Settings);
+
+            var compositionStrategyType = (Type)Settings.Get(WellKnownConfigurationKeys.Topology.Addressing.Composition.Strategy);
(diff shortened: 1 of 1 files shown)


danielmarbach:
@SeanFeldman could you give this a once over?

PR #647, opened 2017-10-06
Only start MultiProducerConcurrentCompletion when really needed

danielmarbach:


src/Transport/Receiving/MessageReceiverNotifier.cs
```diff
@@ -105,7 +108,10 @@ await Task.WhenAll(tasks)
                 logger.Error("The receiver failed to stop with in the time allowed (30s)");
             }
 
-            await completion.Complete().ConfigureAwait(false);
+            if (completion != null)
+            {
+                await completion.Complete().ConfigureAwait(false);
```
SeanFeldman: Can't be expressed as `await completion?.Complete().ConfigureAwait(false);`?

danielmarbach: https://stackoverflow.com/questions/40436213/why-elvis-operator-doesnt-work-with-async-await

PR #649, opened 2017-10-11
Packaging and minor cleanups

danielmarbach:


danielmarbach:
@Particular/azure-maintainers can we review and merge so that we can push the next beta out

src/Transport/NServiceBus.AzureServiceBus.csproj
```diff
@@ -12,7 +12,6 @@
   <PropertyGroup>
     <PackageId>NServiceBus.Azure.Transports.WindowsAzureServiceBus</PackageId>
     <Description>NServiceBus Transport for the Windows Azure Servicebus</Description>
-    <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
```
SeanFeldman: I'd comment this out rather than remove.
For quick local verification, it's handy to be able to produce a package locally.

src/Transport/Config/ExtensionPoints/AzureServiceBusQueueSettings.cs
```diff
@@ -34,7 +34,6 @@ public AzureServiceBusQueueSettings ForwardDeadLetteredMessagesTo(string forward
         /// <summary>
         /// <remarks> Default is set not to forward.</remarks>
         /// </summary>
-        // TODO: needs to be deprecated
```
SeanFeldman: My bad, sorry 😞

PR #655, opened 2017-10-20
Guard.AgainstUnsetSerializerSetting() should not be invoked during defaults setup

### LINKED ISSUE #654: Guard.AgainstUnsetSerializerSetting() is performed too early
`DefaultConfigurationValues. Apply(SettingsHolder settings)` is responsible to set the defaults for the transport.  

Recently, when the [default serializer was  removed](https://github.com/Particular/NServiceBus.AzureServiceBus/pull/615), a guard for configured serializer was added.

https://github.com/Particular/NServiceBus.AzureServiceBus/blob/ae916363d89057349d9264490f7cfc2cacfcfd1b/src/Transport/Config/DefaultConfigurationValues.cs#L13

I've ran into an issue with the timing of this code when running NSB.Bridge, that uses NSB.Raw. The transport is throwing an exception as it doesn't find the default serializer registered by Core.

Should this check be executed where defaults are set by the transport or should it be moved closer to the point where the transport needs to be started?

/cc @SzymonPobiega

SeanFeldman:
Connect to #654 
Fixes #654 

Don't check serializer when setting default values, defer at a later stage of the transport Start

### DIFF
--- a/src/Tests/Configuration/When_using_default_configuration.cs
+++ b/src/Tests/Configuration/When_using_default_configuration.cs
@@ -1,10 +1,8 @@
 ﻿namespace NServiceBus.Azure.WindowsAzureServiceBus.Tests.Configuration
 {
-    using System;
     using AzureServiceBus;
     using Transport.AzureServiceBus;
     using NUnit.Framework;
-    using Settings;
 
     [TestFixture]
     [Category("AzureServiceBus")]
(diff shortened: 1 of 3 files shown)


SeanFeldman:
Removed unit test as it makes no longer sense.
Testing the guard on its own, not sure if worth it.

danielmarbach:
I'm Ok with not testing it. The code is trivial to review

PR #658, opened 2017-10-24
considering operationcancelled exception as transient …

yvesgoeleven:
… so that it does not trigger the circuitbreaker on shutdown


yvesgoeleven:
please review @Particular/azure-maintainers

Issue #659, opened 2017-10-25
Exception on shutdown in NServiceBus V6


SeanFeldman:
Related to #658
Related to #648

SeanFeldman:
Close via https://github.com/Particular/NServiceBus.AzureServiceBus/commit/171cda89df0554d7a65afb838623444e84e438e9

PR #660, opened 2017-10-26
fix for operation cancelled exception triggering concurrent dispose

yvesgoeleven:
Connects to #659

Backport of https://github.com/Particular/NServiceBus.AzureServiceBus/pull/658

Connects to https://github.com/Particular/NServiceBus.Azure/issues/319

## Who's affected

Anyone using NServiceBus.Azure.Transports.WindowsAzureServiceBus transport version 7 and experiencing exceptions during endpoints shutdown.

## Symptoms 

The circuit breaker might be implemented to shutdown the appdomain, which will cause a garbage collection. Message in flight handles by the transport call into the infrastructure at that point may cause the underlying SDK to attempt cleanup using objects which have already been disposed. As a result of that, endpoint logs exceptions and doesn't end cleanly.

## Original issue description

When an endpoint is stopping and exception is thrown, we check if exception is transient or not. If exception is not transient, the error callback is invoked. That will lead to a circuit breaker invocation.

The circuit breaker might be implemented to shutdown the appdomain, which will cause a garbage collection. If other code calls into the infrastructure at that point, e.g. call Close on the pump, it may cause the underlying SDK to attempt cleanup using objects which have already been disposed.

If exception is a `OperationCancelledException`, it should be treated as transient as well to avoid false circuit breaker invocation.

### DIFF
--- a/src/AcceptanceTests/Cleaner.cs
+++ b/src/AcceptanceTests/Cleaner.cs
@@ -1,0 +1,93 @@
+﻿namespace NServiceBus.Azure.Transports.WindowsAzureServiceBus.AcceptanceTests
+{
+    using System;
+    using System.Linq;
+    using System.Threading.Tasks;
+    using Microsoft.ServiceBus;
+    using Microsoft.ServiceBus.Messaging;
+    using NServiceBus.AcceptanceTests;
+    using Logging;
+    using NUnit.Framework;
+
+    public class Cleaner : NServiceBusAcceptanceTest
+    {
+        static ILog logger = LogManager.GetLogger<Cleaner>();
+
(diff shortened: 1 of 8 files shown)


SeanFeldman:
Tests pass, but for EndpointOriented topology there's an [error -100](https://builds.particular.net/viewLog.html?buildId=249961&tab=buildResultsDiv&buildTypeId=NServiceBusAzure_NServiceBusAzureServiceBus_311TestNetFrameworkOnWindowsEndpoint). Might need to look into that with @Particular/deployment-maintainers

yvesgoeleven:
The test When_failure_happens_after_isolated_send.Should_emit_messages only fails on build server, runs fine locally

Issue #648, opened 2017-10-11
Exception on shutdown in NServiceBus V5


yvesgoeleven:
From @ramonsmits 

I'm frequently getting the following with V5:
 ```2017-10-11 09:39:14.1903|FATAL|13|NServiceBus.Azure.Transports.WindowsAzureServiceBus.AzureServiceBusQueueNotifier|Operation cancelled exception occured on receive for queue PerformanceTest, faulting this notifier
System.OperationCanceledException: The operation was canceled.
   at System.Collections.Concurrent.BlockingCollection`1.TryAddWithNoTimeValidation(T item, Int32 millisecondsTimeout, CancellationToken cancellationToken)
   at NServiceBus.Azure.Transports.WindowsAzureServiceBus.AzureServiceBusDequeueStrategy.EnqueueMessage(BrokeredMessage brokeredMessage)
   at NServiceBus.Azure.Transports.WindowsAzureServiceBus.AzureServiceBusQueueNotifier.OnMessage(IAsyncResult ar)
```

This is during shutdown of an endpoint
Due to that..my process exists with an error code
while my main method returned as it should
Seems like a background thread that isn't stopped correctly
would be nice if this could be prevented.

**Comment by ramonsmits** [comment]
@SeanFeldman commented that `OperationCanceledException` would not be V5. Yes, NServiceBus wasn't async but it is using async code.

https://github.com/Particular/NServiceBus.AzureServiceBus/blob/6.4.1/src/Transport/Receiving/AzureServiceBusDequeueStrategy.cs#L300

**Comment by SeanFeldman** [comment]
@ramonsmits  thread based code yes, not async. ASB transport for V5 didn't do anything that was async AFAIK.

As @yvesgoeleven pointed out, if the ASB internally throwing that in one of it's threads, I don't know if this would be caught by our `catch`.

**Comment by SeanFeldman** [comment]
@ramonsmits where does it show up, in the logs or does it bubble up?

**Comment by SeanFeldman** [comment]
Discussed fix: check if token is cancelled before [invoking circuit breaker](https://github.com/Particular/NServiceBus.AzureServiceBus/blob/2599f99b0a4b882b6bdc1679365ea34fc2aa89ee/src/Transport/Receiving/AzureServiceBusDequeueStrategy.cs#L90)

**Comment by ramonsmits** [comment]
@SeanFeldman It only happened at shutdown it seems:

```
[...]

PR #661, opened 2017-10-26
transient exceptions should not invoke the circuit breaker

yvesgoeleven:
Backport of https://github.com/Particular/NServiceBus.AzureServiceBus/pull/658

Connects to https://github.com/Particular/NServiceBus.Azure/issues/319

Probably fixes https://github.com/Particular/NServiceBus.AzureServiceBus/issues/648 as well

## Who's affected

Anyone using NServiceBus.Azure.Transports.WindowsAzureServiceBus transport version 7 and experiencing exceptions during endpoints shutdown.

## Symptoms 

The circuit breaker might be implemented to shutdown the appdomain, which will cause a garbage collection. Message in flight handles by the transport call into the infrastructure at that point may cause the underlying SDK to attempt cleanup using objects which have already been disposed. As a result of that, endpoint logs exceptions and doesn't end cleanly.

## Original issue description

When an endpoint is stopping and exception is thrown, we check if exception is transient or not. If exception is not transient, the error callback is invoked. That will lead to a circuit breaker invocation.

The circuit breaker might be implemented to shutdown the appdomain, which will cause a garbage collection. If other code calls into the infrastructure at that point, e.g. call Close on the pump, it may cause the underlying SDK to attempt cleanup using objects which have already been disposed

If exception is a `OperationCancelledException`, it should be treated as transient as well to avoid false circuit breaker invocation.


yvesgoeleven:
Failed test is json related, nothing to do with this change

PR #666, opened 2017-11-01
Local address not available during tests

SeanFeldman:
Core has introduced a change in one of the latest beta builds.
The change was related to how `Setting.LocalAddress()` is working. The transport used to set directly settings value for the `SharedQueue`. The change in Core has introduced a concept of a "Receive Component" that cannot be manipulated by the transport directly, used by `Settings.LocalAddress()` logic, and is only instantiated when the receive infrastructure is created.

Tests that were executing w/o Core starting the receive infrastructure, would fail.
- Tests that exercise `MessagePump` directly
- Transport Tests
- Any other unit test that would run `AzureServiceBusTransportInfrastructure.Start()`

This change introduced the following:
1. MessagePump is no longer accessing `Settings.LocalAddress()` directly, and instead is injected with its value.
1. `AzureServiceBusTransportInfrastructure` introduced a [workaround](https://github.com/Particular/NServiceBus.AzureServiceBus/compare/local-address-not-available-during-tests?expand=1#diff-c6c4fbda098703835e5096381aa5bb8dR66) that would allow tests not to fail when `Settings.LocalAddress()` is not available (i.e. Core did not set up the "receive component").

### DIFF
--- a/src/Transport/Config/AzureServiceBusTransportInfrastructure.cs
+++ b/src/Transport/Config/AzureServiceBusTransportInfrastructure.cs
@@ -55,11 +55,21 @@
 
             Guard.AgainstUnsetSerializerSetting(Settings);
 
+            var sendOnly = Settings.GetOrDefault<bool>("Endpoint.SendOnly");
+            var localAddressForSendOnlyEndpoint = ToTransportAddress(LogicalAddress.CreateLocalAddress(Settings.EndpointName(), new Dictionary<string, string>()));
+            try
+            {
+                localAddress = sendOnly ? localAddressForSendOnlyEndpoint : Settings.LocalAddress();
+            }
+            catch
+            {
+                // For tests other than Acceptance tests, LocalAddress() will throw
+                localAddress = localAddressForSendOnlyEndpoint;
+            }
+
             defaultNamespaceAlias = Settings.Get<string>(WellKnownConfigurationKeys.Topology.Addressing.DefaultNamespaceAlias);
             namespaceConfigurations = Settings.Get<NamespaceConfigurations>(WellKnownConfigurationKeys.Topology.Addressing.Namespaces);
             messageSizePaddingPercentage = Settings.Get<int>(WellKnownConfigurationKeys.Connectivity.MessageSenders.MessageSizePaddingPercentage);
-            var sendOnly = Settings.GetOrDefault<bool>("Endpoint.SendOnly");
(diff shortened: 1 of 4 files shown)


SeanFeldman:
@Particular/azure-maintainers this PR will resolve the issue with the broken builds after the latest Core betas.

src/Transport/Config/AzureServiceBusTransportInfrastructure.cs
```diff
@@ -55,11 +55,21 @@ void InitializeIfNecessary()
 
             Guard.AgainstUnsetSerializerSetting(Settings);
 
+            var sendOnly = Settings.GetOrDefault<bool>("Endpoint.SendOnly");
+            var localAddressForSendOnlyEndpoint = ToTransportAddress(LogicalAddress.CreateLocalAddress(Settings.EndpointName(), new Dictionary<string, string>()));
+            try
+            {
+                localAddress = sendOnly ? localAddressForSendOnlyEndpoint : Settings.LocalAddress();
+            }
+            catch
```
boblangley: Is there a reason we are not catching `InvalidOperationException` here?

SeanFeldman: As per discussion, this will only happen when core is not setting up the receive component. It means we're in a test that is not an acceptance test and we'll use the endpoint name as the local address (value assigned in case of a send only endpoint). Because of that, catching a specific exception is not required.

danielmarbach: Can't follow the reasoning why this catch is needed. Can you elaborate?

SeanFeldman: @danielmarbach wasn't https://github.com/Particular/NServiceBus.AzureServiceBus/pull/666#issue-270455147 sufficient? (If not, I need to learn to explain things better 😞 ).

danielmarbach: Not really because it is confusing with the mentioning of send-only. Why isn't it sufficient to set in those tests the sendOnly flag? Or are we saying in these mentioned tests it is actually not a send only endpoint and thus we require this try catch?

SeanFeldman: > Or are we saying in these mentioned tests it is actually not a send only endpoint and thus we require this try catch?

Precisely. It's a workaround.

PR #668, opened 2017-11-08
A race condition can prevent subscriptions from being created when using forwarding topology

### LINKED ISSUE #667: One subscription with multiple rules failes to create correctly on endpoint startup
As is:
When an endpoint is configured with forwarding topology with one entity in the bundle, you would expect the endresult to be 1 bundle with 1 subscription with multiple rules (one rule per event).

However, what seems to be happening is there are multiple calls (one for each event) to Task<SubscriptionDescription> Create in AzureServiceBusForwardingSubscriptionCreator.cs. One of the calls creates the subscription, and the other ones receive a MessagingEntityAlreadyExistsException with "the subscription already exists". However, these calls don't add the rule anymore, making the end result be 1 bundle with 1 subscription with 1 rule.

When I restart the endpoint for a second time, the subscription already exists, and the missing rules are added.

Am I missing something in configuration, or is 
https://github.com/Particular/NServiceBus.AzureServiceBus/blob/db4a1d9ca44f59e6d95f47368e3716826abf4d14/src/Transport/Creation/AzureServiceBusForwardingSubscriptionCreator.cs#L94-L98
missing the code to add the rule when the subscription already exists? Like here:
https://github.com/Particular/NServiceBus.AzureServiceBus/blob/db4a1d9ca44f59e6d95f47368e3716826abf4d14/src/Transport/Creation/AzureServiceBusForwardingSubscriptionCreator.cs#L77-L91

boblangley:
Related to #667
Fixes #667

## Who's affected

Anyone using NServiceBus.Azure.Transports.WindowsAzureServiceBus transport version 7 or greater with forwarding topology and endpoints that subscribe to more than one event.

## Symptoms 

Events that expected to be received are not received and the endpoint has a log message similar to the following:

`Subscription 'subscriptionName' in namespace 'namespaceHost' already exists, another node probably beat us to it."`

### DIFF
--- a/src/Transport/Creation/AzureServiceBusForwardingSubscriptionCreator.cs
+++ b/src/Transport/Creation/AzureServiceBusForwardingSubscriptionCreator.cs
@@ -51,7 +51,9 @@
 
             try
             {
-                if (!await ExistsAsync(topicPath, subscriptionName, metadata.Description, namespaceManager).ConfigureAwait(false))
+                var exists = await ExistsAsync(topicPath, subscriptionName, metadata.Description, namespaceManager).ConfigureAwait(false);
+
+                if (!exists)
                 {
                     var ruleDescription = new RuleDescription
                     {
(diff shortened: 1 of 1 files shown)


danielmarbach:
Like it, very clear

PR #676, opened 2017-11-22
Count topics in bundle with numeric values only

### LINKED ISSUE #675: Topic generation can go awry
The function DetermineTopicsToCreate

https://github.com/Particular/NServiceBus.AzureServiceBus/blob/develop/src/Transport/Topology/Topologies/ForwardingTopologySectionManager.cs#L50

uses BundleConfigurations.GetNumberOfTopicInBundle, which simply counts the number of topics that start with prefix "bundle-"

And then generates n number of entities "bundle-"+ i where i is a number

if there is a topic where i is not a number, e.g. bundle-x,  this will result in a new topic created with a number attached incrementing the number of topics every time the endpoint start.

Example

bundle-1, bundle-2, bundle-X -> 3 topics to create, resulting in
bundle-1, bundle-2, bundle-3, bundle-X -> 4 topics to create
bundle-1, bundle-2, bundle-3, bundle-4, bundle-X -> 5 topics to create, 
etc...

SeanFeldman:
Connects to #675 

Fixes #675 by counting topics that follow a pattern for `bundle_prefix-NUMBER`.

### DIFF
--- a/src/Transport/Topology/NumberOfTopicsInBundleCheck.cs
+++ b/src/Transport/Topology/NumberOfTopicsInBundleCheck.cs
@@ -1,10 +1,8 @@
 ﻿namespace NServiceBus.Transport.AzureServiceBus
 {
-    using System.Collections.Generic;
     using System.Linq;
     using System.Text.RegularExpressions;
     using System.Threading.Tasks;
-    using Microsoft.ServiceBus.Messaging;
     using NServiceBus.AzureServiceBus.Topology.MetaModel;
 
     static class NumberOfTopicsInBundleCheck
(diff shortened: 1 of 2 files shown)


SeanFeldman:

danielmarbach:
@SeanFeldman I pushed a slight refactoring of the test

src/Transport/Topology/NumberOfTopicsInBundleCheck.cs
```diff
@@ -9,6 +10,7 @@ static class NumberOfTopicsInBundleCheck
         public static async Task<NamespaceBundleConfigurations> Run(IManageNamespaceManagerLifeCycleInternal manageNamespaceManagerLifeCycle, NamespaceConfigurations namespaceConfigurations, string bundlePrefix)
         {
             var namespaceBundleConfigurations = new NamespaceBundleConfigurations();
+            var topicInBundleNameRegex = new Regex($@"^{bundlePrefix}\d+$", RegexOptions.CultureInvariant);
```
danielmarbach: Since we only ever execute this once it doesn't make sense to cache this regex, right?

SeanFeldman: Correct @danielmarbach.
The time it will take to compile might exceed the execution time. I thought of caching it, but as you've outlined, we only do it once.

SeanFeldman:
@Particular/azure-maintainers this is ready.
 Same question as on the original issue. - should this be backed out too much of an edge case for that?

Issue #673, opened 2017-11-20
ASB v7 Round Robin partitioning doesn't work


SeanFeldman:
## Who's affected

Anyone using Azure Service Bus transport with `RoundRobinNamespacePartitioning` strategy or custom strategy that is expected to return a different namepspace at run-time for sends and publishes.

## Symptoms 

Instead of returning a different namespace as expected, `RoundRobinNamespacePartitioning` strategy returns a cached namespace. For custom topologies, this would manifest in a similar behaviour if returned namespaces are calculated at run-time.

## Original issue

- Issue for version 8 ( #672 )

SeanFeldman:
Based on PR for V8 (#674) the changes for v7 could include the following:
1. Introduce a contract to indicate a `INamespacePartitioningStrategy` strategy is cache-able. Something like `INamespacePartitioningStrategyCanBeCached` with a property `bool SendingNamespacesCanBeCached { get; }`. 
1. Have all provided by the transport strategies implement it by default.
2. Have `ITopologySectionManagerInternal` implementers test for implementation of `INamespacePartitioningStrategyCanBeCached` and make a decision accordignly.
4. Mark `INamespacePartitioningStrategyCanBeCached` as obsoleted and merged into `INamespacePartitioningStrategy` from transport v8.
5. The default behavior in absence of `INamespacePartitioningStrategyCanBeCached` should be **caching** in order to preserve the previous behavior for any custom created topologies

SeanFeldman:
Fixed via https://github.com/Particular/NServiceBus.AzureServiceBus/commit/e8c61ff128d70e28aed7bdffddf5e4175324c9df

Issue #672, opened 2017-11-20
ASB v8 Round Robin partitioning doesn't work


yvesgoeleven:
## Who's affected

Anyone using Azure Service Bus transport with `RoundRobinNamespacePartitioning` strategy or custom strategy that is expected to return a different namepspace at run-time for sends and publishes.

## Symptoms 

Instead of returning a different namespace as expected, `RoundRobinNamespacePartitioning` strategy returns a cached namespace. For custom topologies, this would manifest in a similar behaviour if returned namespaces are calculated at run-time.

## Affected transport versions

- Version 7 (issue #673 )
- Version 8 (this issue)

## Original issue

As reported on https://discuss.particular.net/t/roundrobin-asb-strategy-behavior/193/3

Eventhough the RoundRobinNamespacePartitioning is set, all messages still get sent to the first namespace in the set.

I haven't had time yet to reproduce this, but I likely cause seems to be caching in the topology section manager.

https://github.com/Particular/NServiceBus.AzureServiceBus/blob/develop/src/Transport/Topology/Topologies/ForwardingTopologySectionManager.cs#L132

Once a destination is chosen, the cache will serve it over and over again, therefore nullifying the purpose of the partitioning strategy

SeanFeldman:
@yvesgoeleven that's indeed the issue.
For namespace partitioning we've got several strategies.

### Single namespace partitioning strategy
Not affected as address is calculated once and cached.

### Fail over namespace partitioning strategy
Not affected as address is calculated once and cached.

### Round Robin namespace partitioning
Affected. Everytime we have a destination address, we need to take a [next registered namespace for sending](https://github.com/Particular/NServiceBus.AzureServiceBus/blob/43e1479ca4c12614c79d20f86ea24c003086a389/src/Transport/Addressing/Partitioning/Strategies/RoundRobinNamespacePartitioning.cs#L41). Because we [cache](https://github.com/Particular/NServiceBus.AzureServiceBus/blob/43e1479ca4c12614c79d20f86ea24c003086a389/src/Transport/Topology/Topologies/ForwardingTopologySectionManager.cs#L132) the [value for the destination](https://github.com/Particular/NServiceBus.AzureServiceBus/blob/43e1479ca4c12614c79d20f86ea24c003086a389/src/Transport/Topology/Topologies/ForwardingTopologySectionManager.cs#L170), the code will continue sending to the same namespace selected for the first time. So this is a bug.

[...]

PR #687, opened 2017-11-29
Obsoleting ICacheSendingNamespaces in v8

SeanFeldman:
Related to #672
Related to #673

`ICacheSendingNamespaces` introduced in 7.2.10 should be obsoleted in v8.


SeanFeldman:

PR #688, opened 2017-12-05
Removing max delivery count

### LINKED ISSUE #680: Public API for MaxDeliveryCount is misleading and should be removed
In issue #677 it was identified that public API for `MaxDeliveryCount` on queues and subscriptions is misleading from version 7 of the transport.

Historically, prior to version 7 of the transport, retries were disconnected from the `BrokeredMessage.DeliveryCount` and messages were retried by competing consumer more than should have been. That was causing messaged to dead-letter. 

With version 7 and above the transport configured `MaxDeliveryCount` to work in conjunction with Recoverablity.This insures messages do not get dead-lettered as a result of multiple failing processing attempts, but rather moved to the error queue. 

Customers trying to override the value of `MaxDeliveryCount` either do it as a "habbit of the past" or because they are not aware of the transport behavior. To remove the unncessary configuration and umbiguity what it's for, `MaxDeliveryCount` on queue and subscription settings should be deprecated and removed.

### Plan of Attack

- [x] Implement --> PR #688
- [ ] Update documentation
- [ ] Update upgrade guide
- [ ] Smoke test samples

SeanFeldman:
Connects to #680 
Fixes #680 

Details in the description of #680


SeanFeldman:

src/AcceptanceTests/OldTests/When_sending_an_oversized_message_from_a_transaction_scope.cs
```diff
@@ -40,7 +40,8 @@ class MyEndpoint : EndpointConfigurationBuilder
         {
             public MyEndpoint()
             {
-                EndpointSetup<DefaultServer>(c => c.ConfigureAzureServiceBus().Queues().MaxDeliveryCount(1));
+                // set delivery count to 1
```
danielmarbach: The comment is disturbing. Can we remove it? It doesn't help me because I see the number zero, then I see the delivery count mentioned to be one and then I have to mentally map those two things which doesn't help me in the context of an acceptance test.

src/AcceptanceTests/OldTests/When_sending_an_oversized_message_without_a_transaction_scope.cs
```diff
@@ -33,7 +33,8 @@ class MyEndpoint : EndpointConfigurationBuilder
         {
             public MyEndpoint()
             {
-                EndpointSetup<DefaultServer>(c => c.ConfigureAzureServiceBus().Queues().MaxDeliveryCount(1));
+                // set delivery count to 1
```
danielmarbach: See above

src/Tests/Creation/When_creating_queues.cs
```diff
@@ -166,23 +166,19 @@ public class When_creating_queues
...
-            extensions.Queues().MaxDeliveryCount(100);
-
             var creator = new AzureServiceBusQueueCreator(settings.Get<TopologySettings>().QueueSettings, settings);
 
             await creator.Create("myqueue", namespaceManager);
 
             var real = await namespaceManager.GetQueue("myqueue");
 
-            Assert.AreEqual(100, real.MaxDeliveryCount);
+            Assert.AreEqual(1, real.MaxDeliveryCount);
```
danielmarbach: So this test relies on a default value? Since the core controls that setting could we at least add a hardcoded value under the settings key?

src/Transport/NServiceBus.AzureServiceBus.csproj
```diff
@@ -18,8 +18,8 @@
     <PackageReference Include="Fody" Version="2.*" PrivateAssets="All" />
     <PackageReference Include="Janitor.Fody" Version="1.*" PrivateAssets="All" />
     <PackageReference Include="Obsolete.Fody" Version="4.*" PrivateAssets="All" />
-    <PackageReference Include="Particular.CodeRules" Version="*" PrivateAssets="All" />
-    <PackageReference Include="Particular.Packaging" Version="*" PrivateAssets="All" />
+    <PackageReference Include="Particular.CodeRules" Version="0.2.0" PrivateAssets="All" />
```
danielmarbach: Is this change needed?

src/Transport/obsoletes.cs
```diff
@@ -134,6 +134,15 @@ public AzureServiceBusQueueSettings EnableExpress(Func<string, bool> condition,
         {
             throw new NotImplementedException();
         }
+
+        /// <summary>
+        /// <remarks> Default is 10.</remarks>
+        /// </summary>
+        [ObsoleteEx(Message = ObsoleteMessages.DeprecatedAndNoLongerRequired, TreatAsErrorFromVersion = "8.0", RemoveInVersion = "9.0")]
```
danielmarbach: Should we add here a specific deprecation message? Like

> MaxDeliveryCount is no longer required. Delivery Count can be controlled by using `Recoverability()...`

or something similar

src/Transport/NServiceBus.AzureServiceBus.csproj
SeanFeldman: Yes. VS 2017 15.5.0 is shipping a new NuGet client that is failing with wildcards.

src/AcceptanceTests/OldTests/When_sending_an_oversized_message_from_a_transaction_scope.cs
SeanFeldman: "disturbing" is an interesting choice of word.
When I will see recoverability with immediate retries set to zero in a few months, will not immediately recall that it will cause `MaxDeliveryCount` to be set to 1. Yet this is intentional in this test to eliminate re-processing.
How about " // Disable immediate retries to have max delivery count of 1" ?

danielmarbach: Why do we even need a comment? In an acceptancetest I try as much as I can to use lingo of the higher level and not the lower level.

SeanFeldman: More than happy for suggestions.

src/Tests/Creation/When_creating_queues.cs
SeanFeldman: > could we at least add a hardcoded value under the settings key?

Yes. 
Curious what you meant by "at least". What would _more_ look like?

SeanFeldman: Why would it be a settings key rather than a constant? Settings key is the "key" used to locate a specific piece of info under settings. In this case it's a plain constant. Will introduce a constant to remove a hard-coded value.

src/AcceptanceTests/OldTests/When_sending_an_oversized_message_from_a_transaction_scope.cs
danielmarbach: Here is my suggestion;how about we drop the comment?

src/Tests/Creation/When_creating_queues.cs
danielmarbach: Sorry my explanation was not clear enough. The previous test set it to 100 and then asserted the count to be 100. Now we are no longer setting a value. Now when you read the test you might be wondering where the value is coming from. A constant would solve it or setting an explicit value

src/AcceptanceTests/OldTests/When_sending_an_oversized_message_from_a_transaction_scope.cs
SeanFeldman: @danielmarbach when I read `When_sending_an_oversized_message_from_a_transaction_scope Should_throw_message_too_large_exception` and look at immediate retries set to zero, it will be hard to recall the reason immediate retries were disabled. You've provided a good suggestion to name things better to help with the context. Now you're saying forget it all and just drop the comment. I'll drop the comment, but this feels wrong. We haven't provided the context nor explanation _why_ immediate retries are disabled.

SeanFeldman:
Comments and feedback are incorporated.
Outstanding issues are:
1. [Comment](https://github.com/Particular/NServiceBus.AzureServiceBus/pull/688#discussion_r154864791) (@danielmarbach if reply answers the question, could you remove change request?)
1. Transport tests are failing probably because of no retries. Will address that.

src/Transport/NServiceBus.AzureServiceBus.csproj
danielmarbach: Sigh, nuget

src/AcceptanceTests/OldTests/When_sending_an_oversized_message_from_a_transaction_scope.cs
danielmarbach: I'm providing input here, you are the master of this PR, feel free to ignore my suggestions. I'll remove my changes request

src/Transport/NServiceBus.AzureServiceBus.csproj
SeanFeldman: We love nuget!
image

SeanFeldman:
@Particular/azure-maintainers while working on the failing transport tests I've had to specifying number of immediate retries for the entire TransportTests. This is not something done by either of the transport. After a few discussions, it looks odd that we're trying to set `MaxDeliveryCount` big enough by one to have DLQ working properly with Recoverability NServiceBus provides.  We could abolish this coupling between immediate retries and `MaxDeliveryCount` all together, setting `MaxDeliveryCount` to `int.MaxValue` and recoverability will still work perfectly fine.

Should customer have a need to override that, they could still accomplish this via custom entity descriptors which are evaluated the last. I've introduced this change in commit https://github.com/Particular/NServiceBus.AzureServiceBus/pull/688/commits/5ed1be6991dbd7c7de5cd4b933790ce91193ae43. Please take the time to review.

src/Transport/Creation/AzureServiceBusQueueCreator.cs
```diff
@@ -12,13 +12,13 @@
 
     class AzureServiceBusQueueCreator
     {
+        internal const int DefaultMaxDeliveryCountForNoImmediateRetries = int.MaxValue;
```
yvesgoeleven: Why is this maxvalue? won't this disable the dlq?

src/Transport/Creation/AzureServiceBusQueueCreator.cs
```diff
@@ -31,7 +31,7 @@ public AzureServiceBusQueueCreator(TopologyQueueSettings queueSettings, ReadOnly
                 DefaultMessageTimeToLive = queueSettings.DefaultMessageTimeToLive,
                 EnableDeadLetteringOnMessageExpiration = queueSettings.EnableDeadLetteringOnMessageExpiration,
                 DuplicateDetectionHistoryTimeWindow = queueSettings.DuplicateDetectionHistoryTimeWindow,
-                MaxDeliveryCount = IsSystemQueue(queuePath) ? 10 : numberOfImmediateRetries,
+                MaxDeliveryCount = DefaultMaxDeliveryCountForNoImmediateRetries,
```
yvesgoeleven: default value will disable dlq basically, is that what we want?

boblangley: Does it actually disable it, as in we could not use it to send non deserializable messages? or does it just "effectively disabled" meaning retries will effectively go on forever?

If it is the first, we should set it to some large number that doesn't disable the DLQ. If it is the second then that is the desired result.

src/Transport/Creation/AzureServiceBusQueueCreator.cs
SeanFeldman: Yes, disable accidental auto-forwarding to DLQ.

src/Transport/Creation/AzureServiceBusQueueCreator.cs
SeanFeldman: We try to calculate the next to max # of receive value for `MaxDeliveryCount` on the entities to ensure ASB's dead-lettering doesn't kick in and allow NSB's recoverability to work properly. I.e. need to ensure DLQ feature is "deactivated". Except we can't. Setting it to `int.MaxValue` does exactly that and indicates that the value is not a user arbitrary one, but set by NSB. No sane person will use this large value :slightly_smiling_face:

Thinking of it, we solely make use of DLQ when _ deliberately_ send there a message that cannot be de-serialized. We never want messages to be sent to DLQ as we have the site queue concept and it's mandatory.

DLQ is a native error queue, in some way. NServiceBus error queue mechanism should take the precedence.

boblangley: I understand that, but the word "disable" @yvesgoeleven used implies an inability to use DLQ, which as you said we use for poison non-deserializable messages.

If it isn't disabled in the sense that it can still be used but the native retry mechanism is irrelevant, then that is the desired goal and the `int.MaxValue` is a good value. If there is undocumented functionality that when set to `int.MaxValue` disables the queue so it cannot be used for poison messages, then we should pick another arbitrarily large and insane value, but that does not trigger this undocumented functionality.

yvesgoeleven: @boblangley it's not disabled in that sense afaik, will still be able to manually deadletter

SeanFeldman: Correct @yvesgoeleven. It's "muted". Still can dead-letter explicitly. For example: messages that fail de-serialization will be dead-lettered by our code.

Issue #686, opened 2017-11-28
Allow connection string to be without credentials for custom STS


yvesgoeleven:
## Who's affected

Customers trying to use a custom Security Token Service to secure their Azure Service Bus namespace

## Symptoms

The connectionstring validation does not allow the format “Endpoint=sb://[namespace name].servicebus.windows.net/” without credentials.

- [x] Fix v7 [PR](https://github.com/Particular/NServiceBus.AzureServiceBus/pull/689)
- [x] Port to ASB v8
- [x] Announcement [PR](https://github.com/Particular/Announcements/pull/208)

yvesgoeleven:
This probably boils down to making the second part in the regex optional: https://github.com/Particular/NServiceBus.AzureServiceBus/blob/master/src/Transport/Topology/MetaModel/ConnectionString.cs#L11

SeanFeldman:
> This probably boils down to making the second part in the regex optional:

This would be a breaking change as we use namespace name, SAS key and SAS token for [equality](https://github.com/Particular/NServiceBus.AzureServiceBus/blob/e8c61ff128d70e28aed7bdffddf5e4175324c9df/src/Transport/Topology/MetaModel/ConnectionString.cs#L49-L55).

yvesgoeleven:
@SeanFeldman it would just need a null check, but it would not break the equality logic. A connectionstring without key would not be equal to one with key.

SeanFeldman:
Is there any doco update that would be required as a result of this change?

SeanFeldman:
Classification is missing (`Feature` or `Bug`)

yvesgoeleven:
> Is there any doco update that would be required as a result of this change?

No, we have no doco about connectionstring formats

yvesgoeleven:
Announcement drafted https://github.com/Particular/Announcements/pull/208

SeanFeldman:
> No, we have no doco about connectionstring formats

Should we? This is a very advanced scenario, and custom, that's why I'm not sure if we should or not.

PR #689, opened 2017-12-05
Hotfix 7.2.11

yvesgoeleven:
Fixes https://github.com/Particular/NServiceBus.AzureServiceBus/issues/686

### DIFF
--- a/src/Tests/Topology/MetaModel/When_parsing_string_to_connection_string.cs
+++ b/src/Tests/Topology/MetaModel/When_parsing_string_to_connection_string.cs
@@ -233,5 +233,15 @@
 
             Assert.AreEqual("namespacename", connectionString.NamespaceName);
         }
+
+        [Test]
+        [TestCase("Endpoint=sb://namespacename.servicebus.windows.net")]
+        public void Shared_access_key_should_be_optional(string connectionStringString)
+        {
+            ConnectionString connectionString;
+            ConnectionString.TryParse(connectionStringString, out connectionString);
+
+            Assert.AreEqual("namespacename", connectionString.NamespaceName);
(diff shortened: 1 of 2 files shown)


danielmarbach:
@yvesgoeleven so no null check required for the equality? Could we have a test that verifies that corner case as well?

yvesgoeleven:
@danielmarbach I'll add a test, but the check uses ReferenceEquals which won't throw nullrefs

src/Tests/Topology/MetaModel/When_parsing_string_to_connection_string.cs
```diff
@@ -233,5 +233,39 @@ public void Should_be_able_to_parse_various_domain_names(string connectionString
...
+            ConnectionString connectionString1, connectionString2;
+            ConnectionString.TryParse(connectionStringString, out connectionString1);
+            ConnectionString.TryParse(connectionStringString, out connectionString2);
+
+            var areEqual = connectionString1.Equals(connectionString2);
+            Assert.IsTrue(areEqual);
+        }
+
+        [Test]
+        [TestCase("Endpoint=sb://namespacename.servicebus.windows.net", "Endpoint=sb://namespacename.servicebus.windows.net;SharedAccessKeyName=name;SharedAccessKey=key")]
```
danielmarbach: I do think using test cases here is a bit of an odd choice but I'm not against it. So approving

SeanFeldman: Tend to agree. When I see `TestCase`, I think of permutations.
Considering that this will be ported forward, could be simplified.

PR #694, opened 2018-01-09
[WIP] AddressingLogic should not be applied multiple times

### LINKED ISSUE #695: HierarchyComposition PathGenerator applied multiple times
Raised by @SzymonPobiega 

When applying `HierarchyComposition` the path generator is applied two times and since the composition strategy is not idempotent incorrect paths are generated.

Sample

```
transport.Composition().UseStrategy<HierarchyComposition>().PathGenerator(path => "scadapter/");
```

leads to `scadapter/scadapter/EntityPath` queue name instead of `scadapter/EntityPath` queue name

### Analysis

When [we fixed](https://github.com/Particular/NServiceBus.AzureServiceBus/commit/0efd41c0903c90c66af8d43aadd1ffb90d4aa425) the `ToTransportAddress` implementation we talked about the effects on the strategies and that the strategies should be idempotent. Since we don't have a test for the hierarchy composition strategy we didn't realize that the hierarchy composition strategy is not. This the core calls `ToTransportAddress` which uses the address generator and the composition strategy to resolve the entity path. That new path is passed into `DetermineSendRessources` which then applies the address generator again leading to incorrect paths.

Composition strategies and address generation should be idempotent
Question is whether DetermineSend destination really needs to reapply the address generation.

## PoA

- [x] Fix in `release-8.0` (PR #696)
- [x] Update [Composition doco](https://docs.particular.net/transports/azure-service-bus/configuration/full#physical-addressing-logic-composition) (docs PR https://github.com/Particular/docs.particular.net/pull/3563)

danielmarbach:
Connects to #695 
Fixes #695 

Hacky and non-complete test to repro the hierarchical composition strategy bug

### DIFF
--- a/src/AcceptanceTests/Addressing/When_sending_message_with_hierarchy_composition_strategy.cs
+++ b/src/AcceptanceTests/Addressing/When_sending_message_with_hierarchy_composition_strategy.cs
@@ -1,0 +1,115 @@
+namespace NServiceBus.Azure.Transports.WindowsAzureServiceBus.AcceptanceTests.Addressing
+{
+    using System.Collections.Concurrent;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using AcceptanceTesting;
+    using AcceptanceTesting.Customization;
+    using AzureServiceBus;
+    using NServiceBus.AcceptanceTests;
+    using NServiceBus.AcceptanceTests.EndpointTemplates;
+    using NUnit.Framework;
+
+    public class When_sending_message_with_hierarchy_composition_strategy : NServiceBusAcceptanceTest
+    {
+        [Test]
+        public async Task Should_send_to_one_namespace_only()
+        {
+            var context = await Scenario.Define<Context>()
+                .WithEndpoint<SourceEndpoint>(b =>
+                {
+                    b.When(async (bus, c) =>
+                    {
+                        await bus.Send(new MyRequest());
+                    });
+                })
+                .WithEndpoint<TargetEndpoint>()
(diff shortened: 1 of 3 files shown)


SeanFeldman:
Changed the test to verify queues for the endpoints exist with the expected paths.
Entities are removed between test executions, so there's no "memory effect".

SeanFeldman:
@Particular/azure-maintainers how accurate `When_apply_addressing_logic` tests are? Looks like they were wrong to begin with. Based on what I see, the only time we'd receive anything that would have a suffix (`@alias` or `@connectionString`) would be when we'd have a `ReplyTo` address and run that through `AddressingLogic`. But then, we don't need to sanitize it or run through composition since `ReplyTo` address has already been sanitized and composed by the sending endpoint.

SeanFeldman:
@Particular/azure-maintainers please review. I believe this PR/issue also surfaced [another bug](https://github.com/Particular/NServiceBus.AzureServiceBus/issues/695#issuecomment-356504603) that should be discussed/tackled separately. 
Please give a thought about `AddressingLogic` comment [here](https://github.com/Particular/NServiceBus.AzureServiceBus/pull/694#issuecomment-356512245).

Note: leaving as WIP until we're firm this is good enough.

SeanFeldman:
Update: 2 alternative PRs will be opened to eventually replace this one
1. Make `HierarchyComposition` idempotent (accompanied with documentation PR)
2. `AddressingLogic` to skip processing of values that have a suffix

SeanFeldman:
Closing this PR in favour of alternatives (#696 with #697).

PR #697, opened 2018-01-10
AddressingLogic should shortcut for addresses with suffix

SeanFeldman:
Related to #695 (https://github.com/Particular/NServiceBus.AzureServiceBus/pull/694#issuecomment-356512245)
Replaces #694
Depends on #696

**WARNING** PR #696 needs to be merged prior to this change.

Addressing logic should not operate on values that are already processed destinations such as `destination@alias` or `destination@connectionString` and can shortcut by returning the as-is.

PR #696, opened 2018-01-10
Make hierarchy composition idempotent

SeanFeldman:
Connects to #695 
Replaces #694 

Make `HierarchyComposition` idempotent by not applying path generator if entity path (queue or topic) is already prefixed with it.

PR #698, opened 2018-01-12
AddressingLogic regression bug fix

SeanFeldman:
As part of #696 and #697 fix, a regression bug was introduced.
Raised this question in https://github.com/Particular/NServiceBus.AzureServiceBus/pull/694#issuecomment-356512245, but looks like it was overlooked.

The issue is when a direct send is used while using secure connection strings and `HierarchyComposistion`  using a command such as `Send("endpoint@alias`, short-cutting in `AddressingLogic` would cause to return `endpoing@alias` w/o running the address through hierarchy composition. This would cause dispatch to a non-existent destination and result in failure.

### DIFF
--- a/src/AcceptanceTests/Addressing/When_sending_directly_with_hierarchy_composition_and_secure_connection.cs
+++ b/src/AcceptanceTests/Addressing/When_sending_directly_with_hierarchy_composition_and_secure_connection.cs
@@ -1,0 +1,93 @@
+﻿namespace NServiceBus.Azure.Transports.WindowsAzureServiceBus.AcceptanceTests.Addressing
+{
+    using System.Threading;
+    using System.Threading.Tasks;
+    using AcceptanceTesting;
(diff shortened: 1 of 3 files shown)


SeanFeldman:
@Particular/azure-maintainers please review thoroughly. [Last time](https://github.com/Particular/NServiceBus.AzureServiceBus/pull/694#issuecomment-356518350) it was not caught 😉

danielmarbach:
Good stuff, I must admit I was a bit lost yesterday with the suffix logic and did not find time to review it properly. So I wanted to propose to wait but then it was already merged :) I'm getting old and slow ;)

danielmarbach:
I must admit I wanted to dig into https://github.com/Particular/NServiceBus.AzureServiceBus/pull/697 but couldn't understand it. I was planning to ask a few questions today in the sync but then it was already merged. Good that we caught it in the meantime

PR #704, opened 2018-02-12
Fix exception message confusion

### LINKED ISSUE #702: Wording for MessageLockLostException should call out PrefetchCount
Original exception says:
> A message lock lost exception occurred while trying to complete a message, you may consider to increase the lock duration or reduce the batch size

The reference to "batch size" is confusing in version 7 and 8 of the transport as it was in available in earlier version. As of version 7 and above, it is known as `"PrefetchCount"`.

Locations:
- [first](https://github.com/Particular/NServiceBus.AzureServiceBus/blob/8fd857f5582995943e643d7d2a5ee27d032d1457/src/Transport/Utils/BrokeredMessageExtensions.cs#L22)
- [second](https://github.com/Particular/NServiceBus.AzureServiceBus/blob/8fd857f5582995943e643d7d2a5ee27d032d1457/src/Transport/Utils/BrokeredMessageExtensions.cs#L66)

TODO:
- [ ] Open an issue to cherry pick this fix into 8.0.x hotfix when one is around.

SeanFeldman:
Connected to #702 
Fixes #702

The wording of exception message is incorrect. Batch Size was replaced by Prefetch Count.

### DIFF
--- a/src/Transport/Utils/BrokeredMessageExtensions.cs
+++ b/src/Transport/Utils/BrokeredMessageExtensions.cs
@@ -19,7 +19,7 @@ namespace NServiceBus.Transport.AzureServiceBus
             catch (MessageLockLostException ex)
             {
                 // It's too late to compensate the loss of a message lock. We should just ignore it so that it does not break the receive loop.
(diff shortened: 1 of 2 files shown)


SeanFeldman:
@Particular/azure-maintainers please review and merge

Issue #707, opened 2018-02-14
Support application/octed-stream side by side with the current application/octect-stream


johnm123:
At the risk of sounding like a pedant, is this a typo? If the header is used by a non NSB consumer expecting it to be the conventional "application/octet-stream" this could perhaps cause problems.

In the documentation (https://docs.particular.net/transports/azure-service-bus/brokered-message-creation) it is referred to as "application/octect-stream". In the code, (BrokeredMessagesToIncomingMessagesConverter.cs) it is referred to in the method below: 

` static string GetDefaultTransportEncoding(ReadOnlySettings settings)
{
            var configuredDefault = settings.Get<SupportedBrokeredMessageBodyTypes> WellKnownConfigurationKeys.Serialization.BrokeredMessageBodyType);
            return configuredDefault == SupportedBrokeredMessageBodyTypes.ByteArray ? "wcf/byte-array" : "application/octect-stream";
        }`

yvesgoeleven:
This is a typo indeed, thanks for reporting @johnm123

SeanFeldman:
It could be fixed for incoming messages to support native integration scenario that use a proper message type (`"application/octet-stream"), but not for the outgoing messages. Outgoing messages could be dispatched to the older endpoints that are not updated and it would be a breaking change.

@johnm123 @asosricharde is this a blocker for you?

johnm123:
Thanks for the replies. @SeanFeldman Yes, I understand fixing it would break compatibility with earlier versions that are expecting the typo to be there. Anyhow, I don't think it's a blocker for us as the non-NSB consumers can assume the content to be a stream type.

johnm123:
Hi Sean,

Thanks for the message, makes sense. I would be happy to do it if this is any help.

SeanFeldman:
@johnm123 you're not expected, but sure you can 😃

[...]

PR #716, opened 2018-06-21
Accept incoming messages with transport encoding "application/octet-stream"

johnm123:
Fixes issue #707


SeanFeldman:
Thank you @johnm123 🙂

PR #719, opened 2018-07-09
Handle missing intent

### LINKED ISSUE #718: Missing MessageIntent prevents transport from handling message
## Who's affected

Customers injecting messages into AzureServiceBus containing no message intent (for example when doing integration)

## Symptoms

When a message is received without a Message Intent property on it, an exception is thrown as the transport tries to convert to an NServiceBus Transport Message. This prevents the message from making it to NServiceBus.

[The code assumes the intent property is present as long as any properties are present](https://github.com/Particular/NServiceBus.AzureServiceBus/blob/support-6.4/src/Transport/Utils/BrokeredMessageConverter.cs#L28)

mikeminutillo:
Fixes #718


src/Tests/app.config
```diff
@@ -0,0 +1,33 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
danielmarbach: 🔥

src/Tests/When_converting_from_BrokeredMessage.cs
```diff
@@ -0,0 +1,23 @@
...
+    public class When_converting_from_BrokeredMessage
+    {
+        [Test]
+        public void Should_handle_missing_intent()
+        {
+            var brokeredMessage = new BrokeredMessage();
+            
+            // NOTE: At least property must be set for the brokered message converter to try and retrieve the intent
+            brokeredMessage.Properties["SomeProperty"] = "SomeValue";
+            //brokeredMessage.Properties[Headers.MessageIntent] = MessageIntentEnum.Publish;
```
danielmarbach: 🔥

src/Transport/Utils/BrokeredMessageConverter.cs
```diff
@@ -21,11 +21,17 @@ public static TransportMessage ToTransportMessage(this BrokeredMessage message)
                     headers[Headers.ReplyToAddress] = message.ReplyTo;
                 }
 
+                var intent = default(MessageIntentEnum);
+                if (message.Properties.TryGetValue(Headers.MessageIntent, out var intentProperty))
+                {
+                    intent = (MessageIntentEnum) Enum.Parse(typeof(MessageIntentEnum), intentProperty.ToString());
```
danielmarbach: would it be safer to use `TryParse` here?

src/Transport/Utils/BrokeredMessageConverter.cs
```diff
@@ -21,11 +21,17 @@ public static TransportMessage ToTransportMessage(this BrokeredMessage message)
                     headers[Headers.ReplyToAddress] = message.ReplyTo;
                 }
 
+                var intent = default(MessageIntentEnum);
```
danielmarbach: What would be a reasonable default? Should we set it specifically to a reasonable default?

SeanFeldman: Can there be a default without knowing the real intent?...

mikeminutillo: I think it would have to be Send/Publish. I don't think it ultimately matters once we get to the receiver

PR #725, opened 2018-07-12
Remove unncessary receiver lifecycle management

### LINKED ISSUE #729: Receiver lifecycle management can cause creation of unnecessary amount of receivers
## Who's affected

Customers running endpoints with AzureServiceBus on machines with many logical CPU cores.

## Symptoms

`OnMessage` delegate registration throws an `ObjectAlreadyDisposedException` which triggers the critical error action

## Analysis

During investigations for SC build failures we stumbled upon the message receive notifier. By looking at the code we came to the conclusion that the `MessageReceiverLifeCycleManager` is not useful or could in certain scenarios even be dangerous. 

The current flow is the following:

A message receive notifier manages multiple clients per entity. A message receive notifier observes messages on one single entity path. The message receive notifier creates up to the number of clients per entity message receivers and manages them. The creation of the receivers happens in the background (we did that because creating those receivers is an expensive operation that could significantly impact the startup time of an endpoint). Previously for creating receivers, it called with a concurrency up to the number of clients per entity into `InitializeAndStartReceiver` which then called into `MessageReceiverLifeCycleManager`. The `MessageReceiverLifeCycleManager` also needs the knowledge of how many clients per entities that should be created. Now multiple concurrent operations compete on the `GetOrAdd` method of the `MessageReceiverLifeCycleManager`. Those operations are lock free but the implications are that up to the number of clients per entity operations can try to create a circular buffer and then create receivers. Only one wins and that is returned. Then up to the number of clients per entity concurrent operations try to get an item from the circular buffer and number of clients per entity are returned. The created receivers stay around in the concurrent dictionary of the message receiver although they are already referenced by the `MessageReceiveNotifier`. All the drawbacks of this approach are not necessary and can be removed by letting the `MessageReceiveNotifier` create the receivers directly. Less code and less garbage will be created. I think that is a win.

To show the implications of the design that we have. On an 8 core machine in the worst case we create 8 x 8 = 64 receivers and 8 circular buffers only to return 8 receivers and then throw away 56 (means all the connections need to be closed, happen in the finalizer but still).

On my 12 core machine with 24 logical cores this can lead to 24 x 24 = 576 receivers being created and then what we saw in the tests is that the connections get dropped by the broker so it is much more likely that we end up in scenarios where an opened receiver is returned which then gets closed before we try to hook up OnMessage, which then bombs in the background leaving the endpoint in some sort of zombie state.

danielmarbach:
Fixes #729


danielmarbach:
@SeanFeldman yes but that would be a much bigger change that might need to come in a minor

PR #727, opened 2018-07-18
Fixing concurrent subscribe issue

### LINKED ISSUE #728: Concurrent subscribe can lead to NullReferenceException or lost subscriptions
## Who's affected

- All customers running with AutoSubscription (default in NServiceBus) or doing manual concurrent subscribe in FeatureStartupTask

## Symptoms

- Under certain circumstances, the concurrent subscribe might add a `null` action to the pending start operations which then leads to a `NullReferenceException` during endpoint start
- In other scenarios due to the race on `pendingStartOperations.Add` subscriptions might override each other which leads to subscription notifiers not being setup

danielmarbach:
Fixes #728

### DIFF
--- a/src/Transport/Receiving/MessageReceiverNotifier.cs
+++ b/src/Transport/Receiving/MessageReceiverNotifier.cs
@@ -19,12 +19,7 @@ namespace NServiceBus.Transport.AzureServiceBus
             this.clientEntities = clientEntities;
             this.brokeredMessageConverter = brokeredMessageConverter;
             this.settings = settings;
-            RefCount = 1;
         }
-
-        public bool IsRunning => isRunning;
-
-        public int RefCount { get; set; }
 
         public void Initialize(EntityInfoInternal entity, Func<IncomingMessageDetailsInternal, ReceiveContextInternal, Task> callback, Func<Exception, Task> errorCallback, Action<Exception> criticalError, Func<ErrorContext, Task<ErrorHandleResult>> processingFailureCallback, int maximumConcurrency)
         {
@@ -62,6 +57,14 @@ namespace NServiceBus.Transport.AzureServiceBus
 
         public void Start()
         {
+            if (Interlocked.Increment(ref refCount) == 1)
+            {
+                StartInternal();                
+            }
+        }
+
+        void StartInternal()
+        {
             stopping = false;
             pipelineInvocationTasks = new ConcurrentDictionary<Task, Task>();
             completion?.Start(CompletionCallback, internalReceivers);
@@ -76,6 +79,7 @@ namespace NServiceBus.Transport.AzureServiceBus
                 {
                     tasks[i] = InitializeAndStartReceiver(i, exceptions);
                 }
+
                 await Task.WhenAll(tasks)
                     .ConfigureAwait(false);
 
@@ -87,6 +91,14 @@ namespace NServiceBus.Transport.AzureServiceBus
         }
 
         public async Task Stop()
+        {
+            if (Interlocked.Decrement(ref refCount) == 0)
+            {
+                await StopInternal().ConfigureAwait(false);
+            }
+        }
+
+        async Task StopInternal()
         {
             stopping = true;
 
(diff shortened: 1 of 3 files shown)


src/Transport/Receiving/MessageReceiverNotifier.cs
```diff
@@ -87,6 +91,14 @@ await Task.WhenAll(tasks)
         }
 
         public async Task Stop()
+        {
+            if (Interlocked.Decrement(ref refCount) == 0)
+            {
+                await StopInternal().ConfigureAwait(false);
```
mikeminutillo: Is it possible that this could be stopped more times than it is started? Afterwards you would have to call Start() enough times to bring it back to zero before it would start again.

src/Transport/Topology/TopologyOperator.cs
```diff
@@ -33,14 +33,14 @@ public void Start(TopologySectionInternal topologySection, int maximumConcurrenc
             topology = topologySection;
 
             StartNotifiersFor(topology.Entities);
+            
+            running = true;
 
-            foreach (var operation in pendingStartOperations)
+            Action operation;
+            while (pendingStartOperations.TryTake(out operation))
```
mikeminutillo: The concurrent bag loses the ordering semantics of the list. I think that's OK in this scenario. It can only be subscribe requests and they should not be order dependent.

danielmarbach: Correct the order doesn't matter

src/Transport/Receiving/MessageReceiverNotifier.cs
danielmarbach: As of now we don't support restarting the same endpoint plus didn't the previous code have exactly the same assumptions? Do you think this is an issue we should address?

mikeminutillo: I agree with this argument. This works better than the previous because multiple `Stop()` calls won't try to shut the thing down after it's already shut down.

timbussmann: similar to mike's comment: is it guaranteed that `Stop` is called at least as many times as start? Otherwise a dedicated counter might be safer?

src/Transport/Receiving/MessageReceiverNotifier.cs
```diff
@@ -356,8 +365,8 @@ static Task EmptyErrorCallback(Exception exception)
         ConcurrentDictionary<Task, Task> pipelineInvocationTasks;
         string fullPath;
         EntityInfoInternal entity;
+        long refCount;
```
timbussmann: any reason for choosing a long over int?

src/Transport/Receiving/MessageReceiverNotifier.cs
```diff
@@ -61,6 +56,14 @@ public void Initialize(EntityInfoInternal entity, Func<IncomingMessageDetailsInt
         }
 
         public void Start()
+        {
+            if (Interlocked.Increment(ref refCount) == 1)
```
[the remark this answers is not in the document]
danielmarbach: Get your point. The problem is that if two concurrent callers try to start the first will start it and if it takes long the second one will "think" it is already started. I don't think this is a real problem here because the receive notifier is a push-based approach. Thus a client of the notifier cannot do anything meaningful with it other than just wait that at some point in time something will be pushed to the delegates that were passed in. Making any assumptions about when that will happen is a violation of the nature of this component (plus in real life scenarios nothing could be in the queue thus making assumptions that something gets pushed as soon as you called start seems wrong)

src/Transport/Receiving/MessageReceiverNotifier.cs
danielmarbach: Nothing other than I was usually using the Int64 based overload. Happy to change though

danielmarbach:
Discussed with @timbussmann and @mikeminutillo and we came to the conclusion that since endpoint restarting is not supported we don't need to address now the fact that the code assumes `Start` and `Stop` are called in the right sequence.

There might be an issue when the following scenario happens:

Subscribe<FooMessage>();
--> Start(); // counter 1
Subscribe<FooMessage>();
---> Start(); // counter 2
Endpoint.Stop
---> Stop(); // counter 1
will not be stopped and keep on running

@mikeminutillo and I concluded that potentially the core and our code makes many assumptions around distinct message types that we don't feel we have to address this now.

PR #730, opened 2018-07-18
Incoming messages can have transport as "octet-stream" or "octect-stream"

danielmarbach:
Resolves issue #707


danielmarbach:
This has already been reviewed as part of #716  will merge once green

PR #750, opened 2018-09-24
[WIP] changed tx scope to include processing failure callback

yvesgoeleven:


danielmarbach:
This PR should be targeted against develop if I understand the maintenance release process correctly

PR #753, opened 2018-09-24
Update ASB client v5

SeanFeldman:
* Updated ASB client to v5
* Bumped .NET Framework to 4.6 (required by ASB client v5)


SeanFeldman:
@danielmarbach you're commenting in your 😴 ?
Agreed, we're going to stick to 4.6. Question is - do we release this as a new major of the transport or not. Technically, it's a breaking change (.NET Framework update). cc @Particular/azure-maintainers

yvesgoeleven:
If there are breaking changes, then it needs to be a major. If not, and it's just a nuget package update and nothing else, then we can release as a minor. As I see no code changes to make it work in this PR, I suggest it should be a minor

SeanFeldman:
@yvesgoeleven it requires a framework bump. How is that a minor release? 🙂

yvesgoeleven:
@SeanFeldman from our perspective nothing changed right? But some sys admin might need to install the new framework on the host machine for it to work?

SeanFeldman:
Correct. And that might not be an option for them.

PR #755, opened 2018-09-25
Simplify subscription rule (SQL filter expression)

### LINKED ISSUE #717: Subscription rules are over-complicated and have a heafty performance impact on the broker
Current implementaiton of [SQL filter](https://github.com/Particular/NServiceBus.AzureServiceBus/blob/7a7870a8edf94a74680d5f6c50514bd93a65842a/src/Transport/Topology/MetaModel/SqlSubscriptionFilter.cs#L14) for the transport have a needless complicated filter structure that could be simplified. This complex filtering has a heavy performance impact on the broker as _all_ conditions of the rules are evaluated and no short circuiting is taking place.

Current SQL filter:

```
[{0}] LIKE '{1}%' OR [{0}] LIKE '%{1}%' OR [{0}] LIKE '%{1}' OR [{0}] = '{1}'"
```

Suggested simpler version:

```
[{0}] LIKE '%{1}%'"
```

The simpler version is logically equivalent to the original and covers all the cases.

SeanFeldman:
Fixes #717

### DIFF
--- a/src/Tests/Topology/MetaModel/When_SqlSubscriptionFilter_is_serializing_an_event.cs
+++ b/src/Tests/Topology/MetaModel/When_SqlSubscriptionFilter_is_serializing_an_event.cs
@@ -12,10 +12,7 @@
         {
             var filter = new SqlSubscriptionFilter(typeof(SomeEvent));
             var result = filter.Serialize();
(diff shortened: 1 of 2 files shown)


SeanFeldman:
Please review and merge

PR #757, opened 2018-09-25
Align default BatchFlushInterval with Microsoft ASB client default

### LINKED ISSUE #699: Client-side batching might be too aggressive
Azure Service Bus client configures client-side batching to use `BatchFlushInterval` of 20ms ([doco](https://docs.microsoft.com/en-us/dotnet/api/microsoft.servicebus.messaging.messagesender.batchflushinterval?view=azure-dotnet)).

ASB Transport is setting default to 500ms, which slows down immediate dispatches. Should transport's default be aligned with the native default?

SeanFeldman:
Fixes #699


SeanFeldman:
Please review and merge.

PR #760, opened 2018-10-12
Fix build

danielmarbach:


src/Tests/Seam/When_dispatching_messages_in_receive_context.cs
```diff
@@ -226,7 +232,7 @@ async Task SendEmptyMessageToLocalTestQueue()
         async Task WaitForCompletion()
         {
             await completed.WaitAsync(tokenSource.Token).IgnoreCancellation();
-            await Task.Delay(TimeSpan.FromSeconds(3)); // allow message count to update on queue
+            await Task.Delay(TimeSpan.FromSeconds(5)); // allow message count to update on queue
```
danielmarbach: It seems like that in some scenarios even 5 seconds is not enough

SeanFeldman: There's not much we can do about, can we?

SeanFeldman:
Merging, but not sure we can get away from occasional failures.

src/Tests/Seam/When_dispatching_messages_in_receive_context.cs
danielmarbach: I agree my actions are a bit hopeless. Another way would be to poll the queue for a period of time until the counter reaches what we are looking for or abort if it doesn't converge to what we'd like to see in the allotted time but I'm not sure this is better.

PR #762, opened 2018-10-18
WIP Migration topology

danielmarbach:


src/Transport/Creation/AzureServiceBusSubscriptionCreator.cs
```diff
@@ -161,6 +161,7 @@ bool MembersAreNotEqual(SubscriptionDescription existingDescription, Subscriptio
                    || existingDescription.EnableDeadLetteringOnFilterEvaluationExceptions != newDescription.EnableDeadLetteringOnFilterEvaluationExceptions
                    || existingDescription.MaxDeliveryCount != newDescription.MaxDeliveryCount
                    || existingDescription.EnableBatchedOperations != newDescription.EnableBatchedOperations
+                   || existingDescription.ForwardTo != newDescription.ForwardTo
```
danielmarbach: Is it safe to assume that with the way the tests and subscribers are set up we will always get unique subscribers, therefore, we are not risking to run into a problem where a subscribers subscription stays around from a previous run and then the old code runs which doesn't check for the ForwardTo to be equals and therefore will not recreate the subscription (leaving it in forwarding mode) and thus failing the build

danielmarbach: Those assumptions would no longer hold up if we decide to let all the ATTs run with the migration topology to verify we did not break any scenario that still should be supported (and I think we should do that). In that case, we would need to patch previous versions to take the ForwardTo into account or wipe the namespace clean everytime we execute tests (which has other concequences I know!)

SeanFeldman:
Closing in favour of #770

Issue #763, opened 2018-11-05
Modifying Forwarding rule only on an already created subscriptions doesn't work


danielmarbach:
## Who's affected

Anyone modifying Forwarding rule only on already created subscriptions.

## Symptoms

[`AzureServiceBusSubscriptionCreator`](https://github.com/Particular/NServiceBus.AzureServiceBus/blob/master/src/Transport/Creation/AzureServiceBusSubscriptionCreator.cs#L150-L165) and [`AzureServiceBusForwardingSubscriptionCreator` ](https://github.com/Particular/NServiceBus.AzureServiceBus/blob/master/src/Transport/Creation/AzureServiceBusForwardingSubscriptionCreator.cs#L199-L214) do not take the `ForwardTo` property into account when comparing existing entities with the entities to be created, leading to updating forwarding rule to be omitted.

SeanFeldman:
fixed via https://github.com/Particular/NServiceBus.AzureServiceBus/commit/1188123bf49a4696c83f782a9352f04074fb00ce

PR #767, opened 2018-11-05
Hotfix 8.0.4

SeanFeldman:
## Who's affected

Anyone modifying Forwarding rule only on already created subscriptions.

## Symptoms

[`AzureServiceBusSubscriptionCreator`](https://github.com/Particular/NServiceBus.AzureServiceBus/blob/master/src/Transport/Creation/AzureServiceBusSubscriptionCreator.cs#L150-L165) and [`AzureServiceBusForwardingSubscriptionCreator` ](https://github.com/Particular/NServiceBus.AzureServiceBus/blob/master/src/Transport/Creation/AzureServiceBusForwardingSubscriptionCreator.cs#L199-L214) do not take the `ForwardTo` property into account when comparing existing entities with the entities to be created, leading to updating forwarding rule to be omitted.

---
Backport of #763


SeanFeldman:
Branch build is failing due to a Fody issue that is caused by GitVersion bug we're heating.
More important build is PR of the branch against `support-8.0` branch, but TC is not using it to report status. When overriding and merging, please use `Create a merge commit` option, **not** `Squash and merge`.

@Particular/azure-maintainers please review and merge.

PR #771, opened 2018-11-08
Check forwarding rule match taking into account the path structure

### LINKED ISSUE #772: Forwarding rule check does not take path into account
Fixes a regression introduced in #763 

## Who's affected

Anyone modifying forwarding rule only on already created subscriptions or customers who updated to version 9.0.1

## Symptoms

danielmarbach:
Fixes #772

### DIFF
--- a/src/AcceptanceTests/NServiceBus.AzureServiceBus.AcceptanceTests.csproj
+++ b/src/AcceptanceTests/NServiceBus.AzureServiceBus.AcceptanceTests.csproj
@@ -11,7 +11,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="WindowsAzure.ServiceBus" Version="5.*" />
+    <PackageReference Include="WindowsAzure.ServiceBus" Version="5.0" />
     <PackageReference Include="NServiceBus.Newtonsoft.Json" Version="2.0.0" />
     <PackageReference Include="NServiceBus.AcceptanceTests.Sources" Version="7.1.0" />
     <PackageReference Include="NUnit" Version="3.10.1" />
--- a/src/Tests/Creation/When_creating_subscription.cs
+++ b/src/Tests/Creation/When_creating_subscription.cs
@@ -15,6 +15,7 @@
     public class When_creating_subscription
     {
         const string topicPath = "topic";
+        const string hierarchyTopicPath = "hierarchy/tenant1/topic";
         static SubscriptionMetadataInternal metadata = new SubscriptionMetadataInternal { Description = "eventname" };
         const string sqlFilter = "1=1";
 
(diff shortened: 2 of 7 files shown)


src/AcceptanceTests/NServiceBus.AzureServiceBus.AcceptanceTests.csproj
```diff
@@ -11,7 +11,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="WindowsAzure.ServiceBus" Version="5.*" />
+    <PackageReference Include="WindowsAzure.ServiceBus" Version="5.0" />
```
SeanFeldman: Once 5.0.2 is out, assuming it's fixes the issue, we should revert this change.

src/Tests/Creation/When_creating_subscription.cs
```diff
@@ -15,22 +15,31 @@
...
+        const string hierarchyTopicPath = "hierarchy/tenant1/topic";
         static SubscriptionMetadataInternal metadata = new SubscriptionMetadataInternal { Description = "eventname" };
         const string sqlFilter = "1=1";
 
         [OneTimeSetUp]
         public void TopicSetup()
         {
-            var namespaceManager = new NamespaceManagerAdapterInternal(NamespaceManager.CreateFromConnectionString(AzureServiceBusConnectionString.Value));
-            if (!namespaceManager.TopicExists(topicPath).Result)
+            var manager = new NamespaceManagerAdapterInternal(NamespaceManager.CreateFromConnectionString(AzureServiceBusConnectionString.Value));
```
SeanFeldman: Unnecessary change.

src/Transport/Creation/AzureServiceBusForwardingSubscriptionCreator.cs
```diff
@@ -210,10 +210,25 @@ bool MembersAreNotEqual(SubscriptionDescription existingDescription, Subscriptio
                    || existingDescription.EnableDeadLetteringOnFilterEvaluationExceptions != newDescription.EnableDeadLetteringOnFilterEvaluationExceptions
                    || existingDescription.MaxDeliveryCount != newDescription.MaxDeliveryCount
                    || existingDescription.EnableBatchedOperations != newDescription.EnableBatchedOperations
-                   || existingDescription.ForwardTo != newDescription.ForwardTo
+                   || ForwardNotSame(existingDescription, newDescription)
                    || existingDescription.ForwardDeadLetteredMessagesTo != newDescription.ForwardDeadLetteredMessagesTo;
         }
 
+        static bool ForwardNotSame(SubscriptionDescription existingDescription, SubscriptionDescription newDescription)
```
SeanFeldman: `ForwardToNotSame`. To match the property name we're checking.

src/Transport/Creation/AzureServiceBusSubscriptionCreator.cs
```diff
@@ -161,10 +161,25 @@ bool MembersAreNotEqual(SubscriptionDescription existingDescription, Subscriptio
                    || existingDescription.EnableDeadLetteringOnFilterEvaluationExceptions != newDescription.EnableDeadLetteringOnFilterEvaluationExceptions
                    || existingDescription.MaxDeliveryCount != newDescription.MaxDeliveryCount
                    || existingDescription.EnableBatchedOperations != newDescription.EnableBatchedOperations
-                   || existingDescription.ForwardTo != newDescription.ForwardTo
+                   || ForwardNotSame(existingDescription, newDescription)
                    || existingDescription.ForwardDeadLetteredMessagesTo != newDescription.ForwardDeadLetteredMessagesTo;
         }
 
+        static bool ForwardNotSame(SubscriptionDescription existingDescription, SubscriptionDescription newDescription)
```
SeanFeldman: Same as above

src/AcceptanceTests/NServiceBus.AzureServiceBus.AcceptanceTests.csproj
danielmarbach: In general, we try to avoid now wild cards since they are too dangerous (breaking the builds suddenly)

PR #777, opened 2018-11-09
Hotfix 8.0.5

### LINKED ISSUE #773: Forwarding rule check does not take path into account
Fixes a regression introduced in #767

## Who's affected

Anyone modifying forwarding rule only on already created subscriptions or customers who updated to version 8.0.4

## Symptoms

SeanFeldman:
Fixes #773


SeanFeldman:
Note: there might be an issue with Fody because of GitVersion issue... as long as `hotfix-8.0.5` build is green, this can be merged.

PR #778, opened 2018-11-09
Hotfix 7.2.14

### LINKED ISSUE #774: Forwarding rule check does not take path into account
Fixes a regression introduced in #768 

## Who's affected

Anyone modifying forwarding rule only on already created subscriptions or customers who updated to version 7.2.13

## Symptoms

SeanFeldman:
Fixes #774


SeanFeldman:
Old project format is failing to load on my machine. Need to switch machine to see if can resolve that.

PR #793, opened 2018-12-17
Update development tooling

bording:
This updates development tooling on the `support-8.0` branch.

Note: It is okay for this to be squashed.


bording:
Actually, the reason this is failing is the GitVersion bug where it's choosing 9.0.0 as the version number when it shouldn't.

We'll just need to merge this, and we can confirm `support-8.0` builds properly after.

PR #795, opened 2018-12-20
Update development tooling

bording:
This updates development tooling on the `support-7.2` branch.

Note: It is okay for this to be squashed.


bording:
I had to downgrade to NUnit 3.10.1 (which means downgrading to Particular.Approvals 0.1.0 as well) due to a bug @SeanFeldman and I [discovered ](https://github.com/nunit/nunit/issues/3138)that was causing some acceptance test to fail.

PR #805, opened 2019-01-15
Hotfix 9.1.1

boblangley:


src/AcceptanceTests/Receiving/When_messages_contains_typed_properties.cs
```diff
@@ -0,0 +1,115 @@
...
+                }
+
+                protected override void Setup(FeatureConfigurationContext context)
+                {
+                    context.AddSatelliteReceiver("customSatellite", $"{Conventions.EndpointNamingConvention(typeof(Receiver))}-satellite", new PushRuntimeSettings(1), DefaultRecoverabilityPolicy.Invoke, Handle);
+                }
+
+                Task Handle(IBuilder builder, MessageContext context)
+                {
+                    var scenarioContext = builder.Build<Context>();
```
danielmarbach: Should we add a guard that ignores messages with acceptance test id different from scenario context testrunid? Satellites don't use the pipeline and thus the behaviors we have would not apply here

SeanFeldman: Yeah. It's a sattelite. I forgot that a sattelite is a different queue and messages with a different test run id won't be filtered out. Better have that to avoid "flakiness".👍

danielmarbach: Done

Issue #811, opened 2019-02-28
Duplicated subscriptions with endpoint-oriented topology


SeanFeldman:
## Who's affected

Anyone using endpoint-oriented topology with the legacy Azure Service Bus transport version 9 and above.

## Symptoms

When a subscribing endpoint creates its subscriptions, a duplicate subscription is created for each event. While subscriber is listening to only one of those, messages in the duplicate subscription are accumulated and could cause event overflow, impacting the entire namespace performance. Eventually, this can lead to degradation and interruption of the normal service operation.

## Workaround

A possible workaround is to manually modify the default rule for the original subscription to match the default rule of the duplicate subscription. After doing this, the duplicate subscription can be deleted. When the endpoint restarts it will no longer create the duplicate subscription. 
_Credit: @onesadjam_

## Will be there duplicate messages?

No. There will be no duplicated delivered to the subscribers as subscriber will only receive from a single subscription.

## Will be there duplicate messages in the migration mode?

No. There will be no duplicates delivered to the subscribers while the system is in the [migration mode](https://docs.particular.net/transports/azure-service-bus/legacy/migration).

onesadjam:
A workaround that we employed is to manually modify the default rule for the original subscription to match the default rule of the duplicate subscription.  After doing this, the duplicate subscription can be deleted.  When the endpoint restarts it will no longer create the duplicate subscription.

SeanFeldman:
That's correct. As Subscription creator won't find a mismatching rule in
the subscription it will probe. Great Adam! I'll add that as a viable
workaround.

PR #813, opened 2019-03-01
Hotfix 9.0.5

SeanFeldman:
## Who's affected

Anyone using endpoint-oriented topology with the legacy Azure Service Bus transport version 9 and above.

## Symptoms

When a subscribing endpoint creates its subscriptions, a duplicate subscription is created for each event. While subscriber is listening to only one of those, messages in the duplicate subscription are accumulated and could cause event overflow, impacting the entire namespace performance. Eventually, this can lead to degradation and interruption of the normal service operation.

## Workaround

A possible workaround is to manually modify the default rule for the original subscription to match the default rule of the duplicate subscription. After doing this, the duplicate subscription can be deleted. When the endpoint restarts it will no longer create the duplicate subscription. 
_Credit: @onesadjam_

## Will be there duplicate messages?

No. There will be no duplicated delivered to the subscribers as subscriber will only receive from a single subscription.

## Will be there duplicate messages in the migration mode?

No. There will be no duplicates delivered to the subscribers while the system is in the [migration mode](https://docs.particular.net/transports/azure-service-bus/legacy/migration).

### DIFF
--- a/src/Tests/Creation/When_creating_subscription_backward_compatible_with_v6.cs
+++ b/src/Tests/Creation/When_creating_subscription_backward_compatible_with_v6.cs
@@ -1,6 +1,7 @@
 ﻿namespace NServiceBus.Azure.WindowsAzureServiceBus.Tests.Creation
 {
     using System;
+    using System.Linq;
     using System.Threading.Tasks;
     using Microsoft.ServiceBus;
     using Microsoft.ServiceBus.Messaging;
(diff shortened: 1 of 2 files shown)


SeanFeldman:
Backport of fix for #811 

## In scope

- Do not create duplicate subscription.
- Log info message advising what subscriptions can be optimized by updating SQL Rule, providing an optimized filter SQL expression.
- Test to verify subscriptions are not duplicated when a previously created subscription with transport version 8 or earlier already exist.

## Out of scope

- Automated SQL Rule optimization

Issue #818, opened 2019-04-17
Exceptions in recoverability policy do not trigger critical error


boblangley:
## Who's affected
Anyone using the transport.

## Symptoms
Exceptions thrown in recoverability policy are not propagated as critical error and hide the underlying infrastructure problems that user code could be responding to.

## Description
By not properly handling exceptions when the recoverability policy throws, and not raising a critical error, the transport does not allow taking an action when the underlying messaging infrastructure is partially failing.

SeanFeldman:
Fixed via https://github.com/Particular/NServiceBus.AzureServiceBus/commit/5763dbe1050244522ca56263f9b0b30ed3e02a57

PR #819, opened 2019-04-17
Exceptions in recoverability policy do not trigger critical error

boblangley:
Backport of fix from #818

## Who's affected
Anyone using the transport.

## Symptoms
Exceptions thrown in recoverability policy are not propagated as critical error and hide the underlying infrastructure problems that user code could be responding to.

## Description
By not properly handling exceptions when the recoverability policy throws, and not raising a critical error, the transport does not allow taking an action when the underlying messaging infrastructure is partially failing.


andreasohlund:
Same failing test as the v9 PR so something seems to be broken?

boblangley:
`CriticalError` wasn't raising the error since no endpoints were registered in the test, and there is no way for the test to do so, which means the the change to pass `CriticalError` isn't testable. I reverted to passing around an Action, but added the string generic argument so we can set the critical error message.

PR #820, opened 2019-04-17
Exceptions in recoverability policy do not trigger critical error

boblangley:
Backport of fix from #818

## Who's affected
Anyone using the transport.

## Symptoms
Exceptions thrown in recoverability policy are not propagated as critical error and hide the underlying infrastructure problems that user code could be responding to.

## Description
By not properly handling exceptions when the recoverability policy throws, and not raising a critical error, the transport does not allow taking an action when the underlying messaging infrastructure is partially failing.

### DIFF
--- a/src/Transport/Receiving/MessageReceiverNotifier.cs
+++ b/src/Transport/Receiving/MessageReceiverNotifier.cs
@@ -13,7 +13,7 @@ namespace NServiceBus.Transport.AzureServiceBus
     using NServiceBus.AzureServiceBus.Utils;
     using Settings;
 
-    class MessageReceiverNotifier : INotifyIncomingMessages
+    class MessageReceiverNotifier : INotifyIncomingMessagesWithCriticalError
     {
         public MessageReceiverNotifier(ICreateMessageReceivers receiverCreator, IConvertBrokeredMessagesToIncomingMessages brokeredMessageConverter, ReadOnlySettings settings)
         {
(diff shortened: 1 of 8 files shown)


andreasohlund:
Test fails with 

System.ApplicationException : Could not build NServiceBus.CriticalError
  ----> System.Exception : Failed to build type: NServiceBus.CriticalError
  ----> System.MissingMethodException : No parameterless constructor defined for this object.
   at NServiceBus.Transport.AzureServiceBus.TransportPartsContainer.Resolve[T]() in C:\BuildAgent\work\a0340fa70aa3bd40\src\Transport\Utils\TransportPartsContainer.cs:line 104

src/Transport/Receiving/MessageReceiverNotifier.cs
```diff
@@ -23,9 +23,11 @@ public MessageReceiverNotifier(ICreateMessageReceivers receiverCreator, IConvert
         }
 
         public bool IsRunning { get; set; }
-        
+
         public int RefCount { get; set; }
-        
+
+        public CriticalError CriticalError { get; set; }
```
andreasohlund: Critical error isn't available in DI if I remember correctly?

andreasohlund:
Raised https://github.com/Particular/NServiceBus.AzureServiceBus/pull/822 to try to fix the issue with resolving the critical error. Thoughts?

boblangley:
I merged your PR, and our tests are green. @SeanFeldman can you review and merge?

src/Transport/Receiving/MessageReceiverNotifier.cs
SeanFeldman: I think you are correct.

PR #822, opened 2019-04-18
Pass critical error explicitly

andreasohlund:


andreasohlund:
Looks like it fixed the issue

Issue #829, opened 2019-06-17
Support for custom subscription filter


tchelidze:
Subscription filter is generated by `SqlSubscriptionFilter` with the following logic

```
$"[{Headers.EnclosedMessageTypes}] LIKE '%{eventType.FullName}%'"
```

Which is not sufficient for the following scenario.

Having a legacy message structure `LegacyNamespace.MessageA` which is used by an existing endpoint. Trying to introduce a new message, `NewNamespace.MessageA` (in a new endpoint) which is compatible with an existing one, hence ServiceBus message body generated by serializing legacy message successfully deserializes to a new message.

Filter generated by `SqlSubscriptionFilter`

```
[NServiceBus.EnclosedMessageTypes] LIKE '%NewNamespace.MessageA%'
```

is filtering out legacy messages.

With the ability to customize how subscription filter is generated, we would be able to generate the following filter

```
[NServiceBus.EnclosedMessageTypes] LIKE '%NewNamespace.MessageA%'
OR [NServiceBus.EnclosedMessageTypes] LIKE '%LegacyNamespace.MessageA%'
```
Which would pass both legacy and new messages.

SeanFeldman:
@tchelidze it sounds like you have two distinct events that happened to overlap in their properties and name, but those are not the same event. Could you explain why you cannot have two different events and have the endpoint to subscribe to both?
Also, if this is affecting your production, please raise a support case here: https://particular.net/support

tchelidze:
Hi @SeanFeldman 

Nope, it's exactly the same event. We decided to improve the structure of the event messages. Existing endpoints are currently using an old message. The plan is to gradually migrate all the existing endpoints as well as add new ones, which will use a new message structure.

SeanFeldman:
Thank you, @tchelidze. If this is the same event and you're refactoring the namespace, could you have a handler on the subscriber side that that handles both `OldNamespace.MessageA` and `NewNamespace.MessageA`?

Note that this transport is a [legacy ASB transport](https://docs.particular.net/transports/azure-service-bus/legacy/) which is replaced by the [new ASB transport.](https://docs.particular.net/transports/azure-service-bus/).

tchelidze:
[...]

PR #830, opened 2019-06-17
Support for custom subscription filter

tchelidze:
- Replaced direct instantiations of SqlSubscriptionFilter and CatchAllSubscriptionFilter with indirect one via ICreateBrokerSideSubscriptionFilter factory.
- Added DefaultCreateBrokerSideSubscriptionFilter implementing an existing way of generating filter.
- Added unit test.

### DIFF
--- a/src/Tests/Configuration/When_configuring_namespace_partitioning.cs
+++ b/src/Tests/Configuration/When_configuring_namespace_partitioning.cs
@@ -2,6 +2,7 @@ namespace NServiceBus.Azure.WindowsAzureServiceBus.Tests.Configuration
 {
     using System;
     using System.Collections.Generic;
+    using AzureServiceBus.Connectivity;
     using Transport.AzureServiceBus;
     using NServiceBus.Configuration.AdvancedExtensibility;
     using Settings;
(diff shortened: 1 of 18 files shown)


SeanFeldman:
Related to https://github.com/Particular/NServiceBus.AzureServiceBus/issues/829

DavidBoike:
@tchelidze 	Thanks for the PR. We'll review it and get back to you.

Although, this is for the legacy Azure Service Bus transport, as opposed to the [new transport](https://docs.particular.net/transports/azure-service-bus/). I'm not sure if we'll be adding additional features. Have you considered migrating?

tchelidze:
> @tchelidze Thanks for the PR. We'll review it and get back to you.
> 
> Although, this is for the legacy Azure Service Bus transport, as opposed to the [new transport](https://docs.particular.net/transports/azure-service-bus/). I'm not sure if we'll be adding additional features. Have you considered migrating?

@DavidBoike Sounds good, looking forward to hearing the feedback.

Atm, we aren't planning to upgrade to the new Asb transport.

DavidBoike:
Hi @tchelidze,

We've decided not to accept this PR, as we don't want to create any new features in the legacy Azure Service Bus transport, only fix bugs, especially when there are so many possible workarounds as have been discussed in https://github.com/Particular/NServiceBus.AzureServiceBus/issues/829 (and to which I will add another in just a second).

PR #832, opened 2019-08-29
Bump the ASB client to the latest version

SeanFeldman:
- Update ASB legacy client to version 6.0.0
- Target .NET Framework 4.6.2 (breaking change)


andreasohlund:
Do we still need this one?

danielmarbach:
I think releasing this as a minor would be a good think because it allows customers to stay on the latest legacy client until they migrated to the new transport

SeanFeldman:
2nd what @danielmarbach said. It's not dead. And customers are using it.
image

SeanFeldman:
> I think releasing this as a minor would be a good think because it allows customers to stay on the latest legacy client until they migrated to the new transport

@danielmarbach, if the specific .NET framework doesn't exist on the customer side, a minor release will break their solution. We've done .NET framework changes in the transport as a major update in the past ([version 8 to 9](https://docs.particular.net/transports/upgrades/asb-8to9)). Would you still say this is a minor change?

danielmarbach:
If we need to bump the target framework then we should bump the major correct. I don't think though that is a big deal or significantly increases the work load given that no other changes would be done but updating the target framework and the SDK

yvesgoeleven:
Replaced by https://github.com/Particular/NServiceBus.AzureServiceBus/pull/865, builds as 10.0

PR #868, opened 2020-03-10
Get rid off SettingsHolderFactory because it is not needed

danielmarbach:


SeanFeldman:
Failing tests require serializer to be set.

danielmarbach:
Yes saw that. Will introduce an extension method thaz does it and then use it where needed

SeanFeldman:
For soon to be deprecated legacy transport this kind of refactoring doesn't add anything. Not sure it's worthy but will defer to you.

danielmarbach:
To be frank I did change this because SC had a few hacks in the code that where no longer necessary and figured I can make the situation a bit better. Anyway

PR #902, opened 2020-07-29
Clarify which Azure SDK this transport targets

lailabougria:


README.md
```diff
@@ -1,6 +1,7 @@
 # Legacy Azure Service Bus transport for NServiceBus
 
 The Azure ServiceBus transport for NServiceBus enables the use of the Azure Service Bus Brokered Messaging service as the underlying transport used by NServiceBus. 
+This transport targets the legacy [WindowsAzure.ServiceBus nuget package](https://www.nuget.org/packages/WindowsAzure.ServiceBus/).
```
danielmarbach, line 4: I don't think we should claim that this package is legacy. Microsoft doesn't claim that. How about aligning the wording to what we used in the other PRs like

```suggestion
This transport targets the Azure Service Bus .NET Framework client library available through the [WindowsAzure.ServiceBus nuget package](https://www.nuget.org/packages/WindowsAzure.ServiceBus/).
```

PR #1010, opened 2021-07-21
Discard test

timbussmann:
testing if the `TestIndependenceSkipBehavior` causes problems with the `When_retrying_message_from_error_queue` that is shipped with the 7.5 package because the control message won't get the test-run-id header.


timbussmann:
tests confirmed that this should solve the issue. Bringing in the last commit as a dedicated PR via #1011

PR #1011, opened 2021-07-21
Don't discard unmarked messages in AcceptanceTests

timbussmann:
only discard messages that have a `TestRunId` header AND where this id doesn't match with the current test run. This prevents raw messages or control messages from being discarded although those are part of this test.

Preliminary requirement to make the additional test added through #1008 pass, as it sends a control message that doesn't go through the mutator logic in the test infrastructure that adds the necessary header.

### DIFF
--- a/src/AcceptanceTests/Infrastructure/TestIndependenceBehavior.cs
+++ b/src/AcceptanceTests/Infrastructure/TestIndependenceBehavior.cs
@@ -5,6 +5,7 @@
     using AcceptanceTesting;
     using Pipeline;
     using MessageMutator;
+    using NUnit.Framework;
 
     class TestIndependenceMutator : IMutateOutgoingTransportMessages
     {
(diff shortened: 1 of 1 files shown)


src/AcceptanceTests/Infrastructure/TestIndependenceBehavior.cs
```diff
@@ -35,9 +36,9 @@ public TestIndependenceSkipBehavior(ScenarioContext scenarioContext)
 
         public Task Invoke(ITransportReceiveContext context, Func<ITransportReceiveContext, Task> next)
         {
-            if (!context.Message.Headers.TryGetValue("$AcceptanceTesting.TestRunId", out var runId) || runId != testRunId)
+            if (context.Message.Headers.TryGetValue("$AcceptanceTesting.TestRunId", out var runId) && runId != testRunId)
```
SeanFeldman, line 39: What control messages are those that they were not needed in the past and need to be accounted for now?

timbussmann: New control messages introduced with NSB 7.5 that acknowledge a successful retry to SC (see https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.Core/ServicePlatform/Retries/RetryAcknowledgementBehavior.cs)

SeanFeldman (approved):

SzymonPobiega (approved):

